PR #14, opened 2019-10-04
skip parameter if no param definition found

nickmahilani:
### Context

skip parameter if no param definition found
### Checklist


mantis-runtime/src/main/java/io/mantisrx/runtime/parameter/ParameterUtils.java
```diff
@@ -293,7 +293,8 @@ private static void validationCheck(Func1 validator, Object value,
                     logger.info("mantis runtime parameter " + name + " used, looking up definition >>>");
                     definition = systemParams.get(name);
                 } else {
-                    throw new IllegalArgumentException("No parameter definition for parameter with name: " + name);
+                    logger.info("No parameter definition for parameter with name: {}, will skip parameter", name);
```
neerajrj: I think making this logger.warn would be better for visibility to the user that something is not right.

nickmahilani: done

PR #26, opened 2019-10-27
Travis CI: Update environment, OpenJDK and CPU platforms

EwoutH:
### Context

Update the Travis CI configuration:
 - Update build environment to Ubuntu 18.04 Bionic
 - Update OpenJDK version to 13.x (currently 13.0.1)
 - Add a job executed on ARMv8 CPU.

### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,11 @@
 language: java
-dist: bionic
-jdk: openjdk13
-arch:
- - amd64
- - arm64
+
+jobs:
+  include:
+    - jdk: openjdk8
+    - jdk: openjdk8
+      arch: arm64
+    - jdk: openjdk13
 
 install: "./installViaTravis.sh"
 script: "./buildViaTravis.sh"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: java
 dist: bionic
 jdk: openjdk13
+arch:
+ - amd64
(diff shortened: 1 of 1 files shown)


neerajrj:
@EwoutH thanks for the PR. Unfortunately, we cannot merge this PR yet.
- Upgrading to JDK 13 would require a lot of vetting on our end. Our plan is to start the process soon but it may take a quarter or so.
- For Bionic we need to upgrade the Mesos version first (at least to 1.4). We are in the process of doing that internally at Netflix and once that is done we will roll it out.
Thanks

EwoutH:
CI passes on most OpenJDK and OracleJDK versions when using Ubuntu 16.04: https://travis-ci.com/EwoutH/mantis/builds/133715078

neerajrj:
Yes, I think our concerns are more about our existing production jobs that may use some libraries that could break with a higher level of JDK. We won't know until we try it out. For the platform itself you are right we should be ok.

jeffchao:
Hi @EwoutH. There needs to be more to this change. Specifically, as @neerajrj said we need more rigor around testing and releasing this. Additionally, the `build.gradle` files need to change and the test is still failing. That said, there are a number of other low hanging fruit that we'll look to moving into Github issues for others to contribute. I'm going to close this PR for now. Thanks for looking into JDK 13 in the meantime.

PR #37, opened 2020-02-26
Minor: Fail construction of `WorkerPorts` if not enough ports.

jeffchao:
### Context

Need at least 5 ports.

Mantis Master calls `WorkerPorts#new(List)` from
`SchedulingService#launchTasks`. If Mesos doesn't give it enough ports,
it will mark the launch as a failure and unschedule the worker
(`SchedulingService#unscheduleWorker`) right at that moment.

Keeping the `isValid` function in case a caller wants to use it.

### Checklist

### DIFF
--- a/mantis-common/src/main/java/io/mantisrx/common/WorkerPorts.java
+++ b/mantis-common/src/main/java/io/mantisrx/common/WorkerPorts.java
@@ -37,8 +37,8 @@ public class WorkerPorts {
     private List<Integer> ports;
 
     public WorkerPorts(final List<Integer> assignedPorts) {
-        if (assignedPorts.size() < 4) {
-            throw new IllegalArgumentException("assignedPorts should have at least 4 ports");
+        if (assignedPorts.size() < 5) {
+            throw new IllegalArgumentException("assignedPorts should have at least 5 ports");
         }
         this.metricsPort = assignedPorts.get(0);
         this.debugPort = assignedPorts.get(1);
@@ -50,6 +50,10 @@ public class WorkerPorts {
 
             sinkPort = assignedPorts.get(idx);
             break;
+        }
+
+        if (!isValid()) {
+            throw new IllegalStateException("worker validation failed on port allocation");
         }
     }
 
@@ -122,7 +126,7 @@ public class WorkerPorts {
     }
 
     /**
-     * Validates that this object has 5 valid ports and all of them are unique.
+     * Validates that this object has at least 5 valid ports and all of them are unique.
      */
     public boolean isValid() {
         Set<Integer> uniquePorts = new HashSet<>();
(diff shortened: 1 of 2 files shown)


jeffchao:
@nickmahilani @neerajrj @codyrioux very simple change, but explanation is in the commit description. What do you think of this approach?

https://github.com/Netflix/mantis-control-plane/blob/master/server/src/main/java/io/mantisrx/server/master/SchedulingService.java#L336-L348

codyrioux:
I'm in support of blowing up if we don't have the necessary 5 unique ports. Quick question, the IllegalArgumentException we throw, is it caught and logged somewhere or do we have a metric on it? It might be useful for us to identify when this is happening.

jeffchao:
It's actually not being caught anywhere as far as I can tell. I'll have to catch it and also add a metric for that in the master's scheduling service.

mantis-common/src/main/java/io/mantisrx/common/WorkerPorts.java
```diff
@@ -131,6 +131,7 @@ public boolean isValid() {
         uniquePorts.add(debugPort);
         uniquePorts.add(customPort);
         uniquePorts.add(sinkPort);
+
         return metricsPort > 0 && consolePort > 0 && debugPort > 0 && customPort > 0 && sinkPort > 0
                 && uniquePorts.size() == 5;
```
nickmahilani: size >= 5 should still be valid

jeffchao:
To be clear, I'd add it on the mantis master side. 🎉

mantis-common/src/main/java/io/mantisrx/common/WorkerPorts.java
jeffchao: oh yeah, thanks. changing it now.

nickmahilani:
@codyrioux I think you showed me the sink port missing in the scheduling info when the incident happened(downstream jobs not reading from the Source job). That could be caused by this https://github.com/Netflix/mantis-control-plane/blob/master/server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java#L1388-L1389
Maybe, we want to force a resubmit of a worker for which the worker metadata doesn't have the sink port in Cassandra due to data corruption etc (cc @neerajrj @jeffchao)

jeffchao:
@nickmahilani I think we have to catch it in the spot you pointed out. This will cover master failover. We might have to catch it on a `WorkerLaunch` event too https://github.com/Netflix/mantis-control-plane/blob/master/server/src/main/java/io/mantisrx/master/jobcluster/job/worker/JobWorker.java#L176. Although, we might not need to check this if we already prevented the launch from the first place from scheduling service right?

nickmahilani:
Yeah, think this PR should already cover the `WorkerLaunch` event handling since it is reading from `WorkerPorts` which now does better validation

jeffchao:
@nickmahilani @codyrioux decided to tighten the constraint and clarify some comments. Let me know if it looks good and I can merge. I've got a wip branch on mantis-control-plane that I can pull it in, but will get to that when I'm back. thanks

codyrioux:
This LGTM but one last question before we merge. Do we always assign a sink port @nickmahilani @neerajrj? What about the event in which a job doesn't declare an SSE sink? I assume we allocate a port but nothing is bound to it. (Which should be fine for this PR, but just something to consider in case we are not allocating a port.)

nickmahilani:
sink port is always allocated, master aka control-plane has no knowledge of job's business logic to know if it uses the sink port or not

PR #38, opened 2020-03-25
Remove unused gauges from DropOperator

neerajrj:
- remove requested and subscribe gauges from DropOperator
- remove host and port tags due to high cardinality issue

### Context

To reduce metrics cost removing 

### Checklist

- [ x] `./gradlew test` passes all tests

### DIFF
--- a/mantis-server/mantis-server-worker-client/src/main/java/io/mantisrx/server/worker/client/SseWorkerConnection.java
+++ b/mantis-server/mantis-server-worker-client/src/main/java/io/mantisrx/server/worker/client/SseWorkerConnection.java
@@ -137,12 +137,9 @@ public class SseWorkerConnection {
         this.connectionType = connectionType;
         this.hostname = hostname;
         this.port = port;
-        Tag[] tags = new Tag[2];
-        tags[0] = new BasicTag("host", Optional.ofNullable(hostname).orElse("none"));
-        tags[1] = new BasicTag("port", Optional.ofNullable(port.toString()).orElse("none"));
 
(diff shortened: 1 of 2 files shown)


mantis-server/mantis-server-worker-client/src/main/java/io/mantisrx/server/worker/client/SseWorkerConnection.java
```diff
@@ -137,12 +137,9 @@ public SseWorkerConnection(final String connectionType,
         this.connectionType = connectionType;
         this.hostname = hostname;
         this.port = port;
-        Tag[] tags = new Tag[2];
-        tags[0] = new BasicTag("host", Optional.ofNullable(hostname).orElse("none"));
-        tags[1] = new BasicTag("port", Optional.ofNullable(port.toString()).orElse("none"));
```
nickmahilani: @neerajrj this change would make the metricGroupId same for all connections coming into the worker from all upstream workers, so effectively there is one metric for all SSE data coming into a worker as opposed to one metric group per connection.. This has some implications like when one of the sse connection is closed, it would remove the entire metric group. This might need some testing to make sure this doesn't cause NPEs etc as SSE conns are created/closed.
Also, Drop based autoscaling relies on dropped / processed ratio of DropOperator metrics, I think we should be good on that with this change, just calling it out for testing

neerajrj: Yes, good point. Will see if I can create a single instance of MetricGroupId

PR #40, opened 2020-03-26
set jacksonversion to 2.10.+ series

neerajrj:
### Context

switching jacksonVersion from 2.+ to 2.10.+ so we don't pull in higher versions like 2.11.0.rc1 etc which is incompatible with scala 2.13

### Checklist

- [ x] `./gradlew test` passes all tests


nickmahilani:
can we rely on the nebula resolution rules instead to align all jackson versions? If we pin to 2.10.X here and nflx platform moves to a different minor version, this could break

PR #42, opened 2020-07-07
Add subproject for shading dependencies

calvin681:
### Context

### Checklist


mantis-server/mantis-server-worker/build.gradle
```diff
@@ -19,7 +19,7 @@ apply plugin: 'eu.appsatori.fatjar'
 ext {
     mantisControlPlaneVersion = '1.2.+'
     mantisRxControlVersion = '1.3.14'
-    mesosVersion = '1.7.2'
+    mesosVersion = '1.3.2'
```
jeffchao, line 22: Looks like a rebase/merge error. This should be `1.7.2`.

mantis-common/src/main/java/com/mantisrx/common/utils/JsonUtility.java
```diff
@@ -16,15 +16,14 @@
 
 package com.mantisrx.common.utils;
 
+import mantis.com.fasterxml.jackson.core.type.TypeReference;
```
jeffchao, line 19: We probably don't want to commandeer the `mantis` namespace globally. We could relocate packages into `io.mantisrx.shaded`.

mantis-shaded/build.gradle
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright 2019 Netflix, Inc.
```
jeffchao, line 2: 2020. You can add in copyright from the contributing directions https://github.com/Netflix/mantis/blob/master/CONTRIBUTING.md.

mantis-common/src/main/java/com/mantisrx/common/utils/JsonUtility.java
calvin681: Makes sense. I was following the pattern from mantis-rxnetty. So we may need to relocate that if we decide to merge it into this shaded project in the future.

codyrioux: Ah, yeah I agree `io.mantisrx.shaded` is the place to put it.

mantis-common/src/main/java/io/mantisrx/server/core/ServiceRegistry.java
```diff
@@ -43,7 +43,7 @@ public MantisPropertiesService getPropertiesService() {
     private MantisPropertiesLoader loadMantisPropertiesService() {
         MantisPropertiesLoader mpl = new DefaultMantisPropertiesLoader(new Properties());
         try {
-            mpl = (MantisPropertiesLoader) Class.forName("com.netflix.mantis.common.properties.MantisFastPropertiesLoader").getConstructor(Properties.class)
+            mpl = (MantisPropertiesLoader) Class.forName("com.netflix.io.mantisrx.shaded.common.properties.MantisFastPropertiesLoader").getConstructor(Properties.class)
```
jeffchao, line 46: This is a new file. Not clear what's the right path here: `com.netflix.mantis` vs `com.netflix.io.mantisrx` vs `io.mantisrx`. Can you explain?

jeffchao: cc @codyrioux

codyrioux: I've never seen this but I think in general for the OSS stuff we'll want to stick to `io.mantisrx` as the root namespace.

calvin681: Reverting this file. This creeped in from find/replace. This class only exists in our internal repo and is loaded dynamically. We will not shade that class.

Issue #47, opened 2020-07-24
SSE Streaming Delimiters


codyrioux:
### Context

There is currently a bug in the SSE implementation when compression is enabled. Mantis compiles a batch of data delimiting each event with a fixed delimiter `$$$`. In the case where an event contains the string `$$$` the event will be erroneously split in two and fail to deserialize downstream.

### Steps to reproduce

Create a job producing events which contain `$$$` and subscribe to it with compression enabled.

### Expected behavior

The events should not be incorrectly bifurcated downstream.

### Actual Behavior

The events are bifurcated when being decompressed downstream, when the list is split the legitimate contents of the string are misidentified as a delimiter and the event is erroneously bifurcated.

PR #48, opened 2020-07-24
Allowing the sink subscriber to specify a delimiter via queryParameters.

codyrioux:
Previously compressed output, which is implicitly batched, used a fixed
delimiter of '$$$' to delimit the events in the batch. This causes
problems when the delimiter string naturally occurs in the events. The
proposed solution is to allow users to allow subscribers to specify a
delimiter.

### Context

This is a proposed solution to #47.

### Checklist


codyrioux:
NOTE: Not ready to merge, does not yet include the decompression / tokenizing aspect.

codyrioux:
This should now be ready for merge providing it passes code review. This will allow delimiters to be specified via query parameter manually. Once this is merged we can update the `Source<>` implementations internally to generate a GUID delimiter when compression is enabled.

PR #79, opened 2020-12-01
Configurable concurrency in Group-Scalar stage

neerajrj:
### Context

- Allow concurrency to be configured for group to scalar stages by setting the job parameter mantis.stageConcurrency
- add log statements in groupBy sample

### Checklist

- [x ] `./gradlew test` passes all tests

### DIFF
--- a/mantis-runtime/src/main/java/io/mantisrx/runtime/executor/StageExecutors.java
+++ b/mantis-runtime/src/main/java/io/mantisrx/runtime/executor/StageExecutors.java
@@ -199,7 +199,7 @@ public class StageExecutors {
      */
     @SuppressWarnings("unchecked")
     private static <K, T, R> Observable<Observable<R>> executeMantisGroupsInParallel(Observable<Observable<MantisGroup<K, T>>> go, Computation computation,
-                                                                                     final Context context, final boolean applyTimeoutToInners, final long timeout) {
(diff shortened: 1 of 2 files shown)


jeffchao (approved):

mantis-runtime/src/main/java/io/mantisrx/runtime/executor/StageExecutors.java
```diff
@@ -199,22 +199,44 @@ public void stop() {}
...
 
-        return
-                go
-                        .lift(new MonitorOperator<>("worker_stage_outer"))
-                        .map(observable -> c
-                                .call(context, observable
-                                        .subscribeOn(Schedulers.computation())
-                                        .lift(new MonitorOperator<MantisGroup<K, T>>("worker_stage_inner_input")))
-                                .lift(new MonitorOperator<R>("worker_stage_inner_output")));
+        if(concurrency == StageConfig.DEFAULT_STAGE_CONCURRENCY) {
```
nickmahilani, line 210: same logic for the GroupedObservable code path as well?

neerajrj: That won't suffice as we would have to avoid the merge that happens early on in the network layer as well

nickmahilani: gotcha

nickmahilani (approved):

PR #82, opened 2020-12-08
Subscribe to source job worker metrics in job master

calvin681:
### Context

This PR adds support for subscribing to drop metrics from source job workers. The feature is gated by a job property. The source job metrics are funneled to the stage 1 operator since that's the stage you need to scale to mitigate the drops.

The PR lays the groundwork to start receiving source job metrics. There will be followup PRs to actually use these metrics in auto scaling and outlier detection.

### Checklist


mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/jobmaster/WorkerMetricHandler.java
```diff
@@ -280,6 +307,28 @@ public void call() {
...
+                                for (Map.Entry<String, GaugeData> group : metricGroups.entrySet()) {
+                                    String metricKey = worker.getKey() + ":" + group.getKey();
+                                    Map<String, Double> gauges = group.getValue().getGauges();
+                                    if (sourceMetricsRecent.containsKey(metricKey) && gauges.containsKey(DROPPED_COUNTER_METRIC_NAME)) {
+                                        sourceJobDrops += gauges.get(DROPPED_COUNTER_METRIC_NAME);
+                                        hasSourceJobDropsMetric = true;
+                                    }
+                                }
+                            }
+                            if (hasSourceJobDropsMetric) {
```
neerajrj, line 325: Is there a way to weed out outliers? As in say source job is dropping data while writing to to only 1 worker, then we don't necessarily want to scale up this job right

calvin681: Yes, that's planned in a followup PR. The outlier detector should automatically resubmit the worker when it detects drops are only to 1 worker.

mantis-runtime/src/main/java/io/mantisrx/runtime/parameter/SourceJobParameters.java
```diff
@@ -0,0 +1,228 @@
...
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonProperty;
+import io.mantisrx.shaded.com.fasterxml.jackson.core.type.TypeReference;
+import io.mantisrx.shaded.com.fasterxml.jackson.databind.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
+
+public class SourceJobParameters {
+    public static final String MANTIS_SOURCEJOB_TARGET_KEY = "target";
```
nickmahilani, line 31: This assumes a multi source job connector. For single source job connectors, we would look at the sourceJobName job param.
For MRE v2 app stream connector, there is no source job name in the job params.

We should probably make this more generic for 
1) Source job name config: either identify the source job names that need to be queried (using another job param value or app stream etc) or provide the source job names in the autoscaling config 
2) the metric group patterns to match against

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/jobmaster/AutoScaleMetricsConfig.java
```diff
@@ -98,6 +120,16 @@ public AggregationAlgo getAggregationAlgo(final String metricGroupName, final St
         return getAllMetrics().keySet();
     }
 
+    public Set<String> generateSourceJobMetricGroups(Set<String> clientIds) {
+        Set<String> results = new HashSet<>();
+        for (String clientId : clientIds) {
+            for (String metricPattern : sourceJobMetrics.keySet()) {
+                results.add(metricPattern.replaceAll(CLIENT_ID_TOKEN, clientId));
```
nickmahilani, line 127: We could probably just look for all ServerSentEventRequestHandler dropped metrics with current-job-Id(should be in the clientId by default?) in the metric group name.. There will also be sockAddr in the metricGroupName, so there will be multiple metric groups per clientId anyways.

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/jobmaster/WorkerMetricHandler.java
```diff
@@ -280,6 +307,28 @@ public void call() {
...
+                                logger.info("Job stage " + stage + ", recent source job metrics count: " +
+                                        sourceMetricsRecent.size());
+                                jobAutoScaleObserver.onNext(
+                                        new JobAutoScaler.Event(StageScalingPolicy.ScalingReason.SourceJobDrop, stage,
+                                                sourceJobDrops / 6.0 / numWorkers, numWorkers, ""));
```
nickmahilani, line 330: why 6.0?

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/jobmaster/WorkerMetricHandler.java
```diff
@@ -280,6 +307,28 @@ public void call() {
...
+
+                            double sourceJobDrops = 0;
+                            boolean hasSourceJobDropsMetric = false;
+                            Map<String, String> sourceMetricsRecent = sourceJobMetricsRecent.asMap();
+                            for (Map.Entry<String, WorkerMetrics> worker : sourceJobWorkersMap.entrySet()) {
+                                Map<String, GaugeData> metricGroups = metricAggregator.getAggregates(worker.getValue().getGaugesByMetricGrp());
+                                for (Map.Entry<String, GaugeData> group : metricGroups.entrySet()) {
+                                    String metricKey = worker.getKey() + ":" + group.getKey();
+                                    Map<String, Double> gauges = group.getValue().getGauges();
+                                    if (sourceMetricsRecent.containsKey(metricKey) && gauges.containsKey(DROPPED_COUNTER_METRIC_NAME)) {
```
nickmahilani, line 319: gauges list won't have the Dropped counter, so this condition will be false always. There is a separate counters list in the metric event. Though, the step based autoscaler has only looked at gauges so far, so the aggregations of the metric data for a single worker and across workers might have some assumptions based on that.

mantis-runtime/src/main/java/io/mantisrx/runtime/parameter/SourceJobParameters.java
calvin681: (1) makes sense. The various connectors may use different params for both `sourceJobName` and `clientId`. There is no clean way to support all of them. I will add a system param called `mantis.jobmaster.autoscale.sourcejob.target` that can be used as override.

(2) what's the use case? Are there other job-to-job transport outside of `ServerSentEventRequestHandler` that can emit drops?

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/jobmaster/AutoScaleMetricsConfig.java
calvin681: The job can have multiple clientIds. The patterns generated would be `ServerSentEventRequestHandler:clientId=clientId1:*`, etc. `AutoScaleMetricConfigTest` should make this clear. Is this what you mean?

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/jobmaster/WorkerMetricHandler.java
calvin681: Similar to what's already done on the RPS metric: https://github.com/Netflix/mantis/pull/82/files/e489031772bf021ee83fe0c145e1e202dbc0638a#diff-8068879b4809e7e6458c38f3ab1ce02fb794d943a9b0b4a1953d21d1a07948a4R307

This is to get a per second number and relates to how atlas cumulates metrics for 6 seconds.

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/jobmaster/WorkerMetricHandler.java
calvin681: The metric gets converted to gauges here: https://github.com/Netflix/mantis/pull/82/files/e489031772bf021ee83fe0c145e1e202dbc0638a#diff-fce5244027e72de847d3bffc7dfdd658a28fd5ab90b802b0dd662268a1112f2bR100

This is to simplify implementation and similar to how the stage ON_NEXT_GAUGE metric is generated. The other option is to make `MetricData`, `WorkerMetric` classes counter aware. That would be a bigger change.

mantis-runtime/src/main/java/io/mantisrx/runtime/parameter/SourceJobParameters.java
nickmahilani: 2) no current use case, this is more forward looking if we add a websocket/grpc sink in future.

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/jobmaster/WorkerMetricHandler.java
nickmahilani: oh ok, I missed the counter to gauge conversion, so was confused reading the code.

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/jobmaster/WorkerMetricHandler.java
nickmahilani: hmm didn't know that the atlas counters are reset every 6 sec.. should probably add a comment

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/jobmaster/AutoScaleMetricsConfig.java
nickmahilani: nevermind, I was thinking the `clientId` would always have the current jobID in it, so we could just look for `ServerSentEventRequestHandler:.*clientId=<currentJobId>.*` but it looks like `clientId` can be overridden in the TargetInfo job param.

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/jobmaster/AutoScaleMetricsConfig.java
```diff
@@ -17,18 +17,28 @@
...
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 
 public class AutoScaleMetricsConfig {
 
+    public static final String CLIENT_ID_TOKEN = "_CLIENT_ID_";
+    public static final String OUTBOUND_METRIC_GROUP_PATTERN = "PushServerSse:clientId=" + CLIENT_ID_TOKEN + ":*";
```
nickmahilani, line 32: suggest using a string constant for `PushServerSse` / `ServerSentEventRequestHandler` and `clientId` that is used for both metric generation and consumption here. So we don't unintentionally break this if/when the metric name is changed or a new tag is added to the metric etc.

nickmahilani (approved):
LGTM

PR #87, opened 2021-01-12
ignore all exceptions during master init for autoscale rules setup

nickmahilani:
### Context

Failure to load autoscale rules due to other exceptions could fail SchedulingService init which in turn prevents master from starting during bootstrap
### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/SchedulingService.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/SchedulingService.java
@@ -238,8 +238,8 @@ public class SchedulingService extends BaseService implements MantisScheduler {
                 } else
                     logger.warn("No auto scale rules setup");
             }
-        } catch (IllegalStateException e) {
-            logger.warn("Ignoring: " + e.getMessage());
+        } catch (Exception e) {
+            logger.error("Ignoring exception during autoscale rules setup", e);
         }
         schedulerBuilder = schedulerBuilder.withMaxOffersToReject(Math.max(1, minMinIdle));
         final TaskScheduler scheduler = schedulerBuilder.build();


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/SchedulingService.java
```diff
@@ -238,8 +238,8 @@ private TaskScheduler setupTaskSchedulerAndAutoScaler(Observable<String> vmLease
                 } else
                     logger.warn("No auto scale rules setup");
             }
-        } catch (IllegalStateException e) {
-            logger.warn("Ignoring: " + e.getMessage());
+        } catch (Exception e) {
+            logger.error("Ignoring exception during autoscale rules setup", e);
```
neerajrj, line 242: @nickmahilani Maybe add a check that if this happens for all ASGs then we terminate the JVM ?

nickmahilani: visibility on ASG level failures is not available here, we could probably add that check in nfmantis where we get the autoscale rules to decide between retryable and non-retryable failures.

neerajrj: sounds good

neerajrj (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/SchedulingService.java
nickmahilani: I will close this PR, the getRules impl in nfmantis can throw an IllegalStateException for failures that can be ignored

PR #93, opened 2021-01-27
Fix SingleThreadScheduler createWorker to not reuse NewThreadWorker created in constructor

nickmahilani:
### Context

- NewThreadWorker shuts down the internal executor on an unsubscribe (https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/schedulers/NewThreadWorker.java#L289), this can exhaust the SingleThreadScheduler threads causing processing to stall if the Source observable onCompletes. 
Fix the SingleThreadScheduler to leverage Rx Scheduler from an ExecutorService with 1 thread and create a worker using this scheduler.
- rename SingleThreadScheduler to MantisRxScheduler
### Checklist

### DIFF
--- a/mantis-runtime/src/main/java/io/mantisrx/runtime/scheduler/MantisRxScheduler.java
+++ b/mantis-runtime/src/main/java/io/mantisrx/runtime/scheduler/MantisRxScheduler.java
@@ -16,25 +16,28 @@
 
 package io.mantisrx.runtime.scheduler;
 
+import static java.util.concurrent.Executors.newFixedThreadPool;
+
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ThreadFactory;
 
 import rx.Scheduler;
-import rx.internal.schedulers.NewThreadWorker;
-
+import rx.schedulers.Schedulers;
 
 /**
- * Schedules work on the same thread that is constructed once as part of the creating this Scheduler.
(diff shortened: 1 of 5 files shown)


mantis-runtime/src/main/java/io/mantisrx/runtime/scheduler/MantisRxScheduler.java
```diff
@@ -16,25 +16,28 @@
...
 
 /**
- * Schedules work on the same thread that is constructed once as part of the creating this Scheduler.
+ * Schedules work on the same fixed thread pool executor with 1 thread that is constructed once as part of creating
+ * this Scheduler.
  */
-public final class SingleThreadScheduler extends Scheduler {
-
-    private final NewThreadWorker worker;
+public final class MantisRxScheduler extends Scheduler {
```
neerajrj, line 31: Maybe a nitpick, should we name it to signify this is a single threaded scheduler?

nickmahilani: I went back and forth on that, can revert the naming if that is more clear.

nickmahilani: updated to MantisRxSingleThreadScheduler

calvin681 (approved):

PR #96, opened 2021-03-18
Replace jcenter with maven central

rpalcolea:
This replaces jcenter with maven central

Also noticed that artifacts from `net.andreinc.*` are in use for some mantis samples and those are in jcenter only. I went ahead and create a backup in our own repos so when jcenter goes away, we will be fine

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,7 @@
 
 buildscript {
     repositories {
-        jcenter()
+        mavenCentral()
         maven {
             url "https://plugins.gradle.org/m2/"
         }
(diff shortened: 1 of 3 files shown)


jeffchao (approved):
LGTM. Thanks for the contribution. I'll merge after CI.

calvin681 (approved):

rpalcolea:
Sounds good @jeffchao , also opened:

https://github.com/Netflix/mantis-rxcontrol/pull/10
https://github.com/Netflix/mantis-rxnetty/pull/5
https://github.com/Netflix/mantis-gradle-plugin/pull/4
https://github.com/Netflix/mantis-mql/pull/4

PR #97, opened 2021-03-26
Increment mantisEventsDropped metric when there is an exception during processing, create new Map in Event by default

calvin681:
### Context

Log mantisEventsDropped metric when there is an exception during processing.

### Checklist


mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/EventDrainer.java
```diff
@@ -108,7 +110,12 @@ public void run() {
                         streamEventList.clear();
                     }
                 } catch (Exception e) {
-                    LOG.warn("Exception processing events for stream {}", stream, e);
+                    LOG.error("Exception processing events for stream {}", stream, e);
```
nickmahilani, line 113: might be ok to log this at WARN/INFO since there is a metric now to indicate the same.

nickmahilani (approved):

PR #103, opened 2021-05-04
Feeds backpressure back to drop operator on SSE connection

calvin681:
### Context

According to RxJava docs, in `flatmap`, the outer Observable is consumed in unbounded mode:

https://javadoc.io/doc/io.reactivex/rxjava/latest/rx/Observable.html#flatMap-rx.functions.Func1-

Thus, prior to this PR, we would consume every item from the RxNetty response at the `flatmap` operator even if the downstream cannot request all of them. Since the `DropOperator` is above this, it will not see any backpressure. And the `incoming_*_dropped` metric is 0.

To fix this, we are setting `maxConcurrent` to 1 on these 2 `flatmap` operations. In each of these instances, each item from the outer Observable produces a finite inner Observable. In the first case, an item leads to either `Observable.error` or Observable.just`. In the second case, it decompress the content into finite events. Setting `maxConcurrent` to 1 forces the outer Observable to be processed sequentially, and propagates backpressure appropriately. This means backpressure is now correctly surfaced in the `DropOperator` on line 287.

### Checklist


nickmahilani:
good find ! Can we add a test for this?

jeffchao (approved):

nickmahilani (approved):

PR #105, opened 2021-05-25
Fix DataDrop metrics base on Counter semantics

calvin681:
### Context

Underlying Counter implementation differs from what the old code expects. Counter value is not monotonously increasing. Therefore, it is not necessary to keep prior values to compute the difference. This simplifies the logic in DataDrop aggregation.

### Checklist


nickmahilani (approved):
LGTM

PR #110, opened 2021-08-30
Fix missing subscription bug when max sub property change

calvin681:
### Context

Clean up logic in AbstractSubscriptionTracker to always use StreamManager as source of truth of what subscription is stored. Before, AbstractSubscriptionTracker tracks the previous subscription internally. But this internal copy can be different from what the StreamManager actually stored, for example due to max subscription being reached. Also added more tests to test AbstractSubscriptionTracker and StreamManager.

### Checklist


nickmahilani (approved):

mantis-publish/mantis-publish-core/src/test/java/io/mantisrx/publish/AbstractSubscriptionTrackerTest.java
```diff
@@ -0,0 +1,345 @@
...
+        Set<String> expected = ImmutableSet.of("id1", "id2");
+        assertEquals(expected, subIds);
+
+        config.setProperty(String.format(MAX_SUBS_PER_STREAM_FORMAT, StreamType.DEFAULT_EVENT_STREAM), 4);
+
+        nextSubs = ImmutableList.of(
+                new MantisServerSubscription("id1", "select * from defaultStream where id = 1", null),
+                new MantisServerSubscription("id2", "select * from defaultStream where id = 2", null),
+                new MantisServerSubscription("id3", "select * from defaultStream where id = 3", null),
+                new MantisServerSubscription("id4", "select * from defaultStream where id = 4", null)
```
Andyz26, line 116: nit: add "id5" sub to validate max cap is still valid.

calvin681: Good call.

mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/AbstractSubscriptionTracker.java
```diff
@@ -66,52 +72,51 @@ public AbstractSubscriptionTracker(MrePublishConfiguration mrePublishConfigurati
...
+
+        for (MantisServerSubscription newSub : curr) {
+            // Add new subscription not present previously
+            if (!previousIds.contains(newSub.getSubscriptionId())) {
+                try {
+                    Optional<Subscription> subscription = SubscriptionFactory.getSubscription(newSub.getSubscriptionId(), newSub.getQuery());
+                    if (subscription.isPresent()) {
+                        streamManager.addStreamSubscription(subscription.get());
+                    } else {
+                        LOG.info("will not add invalid subscription {}", newSub);
```
sundargates, line 86: I'm assuming this should never happen? Do you want to may be increase the log level on this?

mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/AbstractSubscriptionTracker.java
```diff
@@ -66,52 +72,51 @@ public AbstractSubscriptionTracker(MrePublishConfiguration mrePublishConfigurati
...
-                if (subscription.isPresent()) {
-                    streamManager.addStreamSubscription(subscription.get());
-                } else {
-                    LOG.info("will not add invalid subscription {}", subToAdd);
-                }
+                streamManager.removeStreamSubscription(prevId);
             } catch (Throwable t) {
                 if (LOG.isDebugEnabled()) {
-                    LOG.debug("failed to add subscription {}", subToAdd, t);
+                    LOG.debug("failed to remove subscription {}", prevId);
```
sundargates, line 104: Same as above. do we not want to increase the log level on this?

mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/AbstractSubscriptionTracker.java
```diff
@@ -66,52 +72,51 @@ public AbstractSubscriptionTracker(MrePublishConfiguration mrePublishConfigurati
...
     }
 
 
     private void cleanupStaleSubscriptions(String streamName) {
-        StreamSubscriptions streamSubscriptions = previousSubscriptions.get(streamName);
-        if (streamSubscriptions != null) {
-            boolean hasStaleSubscriptionsData = (System.currentTimeMillis() - streamSubscriptions.getCreateTimeMs()) > mrePublishConfiguration.subscriptionExpiryIntervalSec() * 1000;
+        Long lastFetched = streamLastFetchedTs.get(streamName);
+        if (lastFetched != null) {
+            boolean hasStaleSubscriptionsData = (System.currentTimeMillis() - lastFetched) > mrePublishConfiguration.subscriptionExpiryIntervalSec() * 1000;
```
sundargates, line 114: might be a good idea to get rid of the dependency on the System clock here at a later point.

mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/AbstractSubscriptionTracker.java
```diff
@@ -146,12 +151,9 @@ public void refreshSubscriptions() {
...
                         if (subsEnvelopeO.isPresent()) {
                             MantisServerSubscriptionEnvelope subsEnvelope = subsEnvelopeO.get();
-                            propagateSubscriptionChanges(previousSubscriptions
-                                            .getOrDefault(streamName, new StreamSubscriptions(streamName, DEFAULT_EMPTY_SUB_ENVELOPE))
-                                            .getSubsEnvelope().getSubscriptions(),
-                                    subsEnvelope.getSubscriptions());
+                            propagateSubscriptionChanges(streamName, subsEnvelope.getSubscriptions());
                             LOG.debug("{} subscriptions updated to {}", streamName, subsEnvelope);
-                            previousSubscriptions.put(streamName, new StreamSubscriptions(streamName, subsEnvelope));
+                            streamLastFetchedTs.put(streamName, System.currentTimeMillis());
```
sundargates, line 156: nit: same as above. Get rid of the System clock dependency.

mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/AbstractSubscriptionTracker.java
```diff
@@ -174,34 +176,7 @@ public void refreshSubscriptions() {
...
-
-        public MantisServerSubscriptionEnvelope getSubsEnvelope() {
-            return subsEnvelope;
-        }
-
-        public long getCreateTimeMs() {
-            return createTimeMs;
-        }
+    protected Set<String> getCurrentSubIds(String streamName) {
+        return streamManager.getStreamSubscriptions(streamName).stream().map(Subscription::getSubscriptionId).collect(Collectors.toSet());
```
sundargates, line 180: nit: looks > 80 chars. do you want to maybe divide the chain along the operator boundary?

sundargates (approved):

Andyz26 (approved):

mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/AbstractSubscriptionTracker.java
calvin681: What's the issue with using system clock for expiration? What would you replace it with?

sundargates: Because it makes it harder to test. Generally, I recommend something like 

https://github.com/Netflix/spectator/blob/606f3241057a4cd6fc0632625bd4ea11c5c65bc1/spectator-api/src/main/java/com/netflix/spectator/api/Clock.java 
Usage: https://github.com/Netflix/spectator/blob/606f3241057a4cd6fc0632625bd4ea11c5c65bc1/spectator-api/src/main/java/com/netflix/spectator/api/AbstractRegistry.java#L44

or just depending on java.time.Clock. 

Again, this is not required for this change as you didn't introduce it. just might be a good idea to replace it at some point IMO.

PR #112, opened 2021-09-10
Refactor TimedChunker to fix data drop due to race condition

calvin681:
### Context

Simplified the logic and removed Chunker altogether. This fixes the issue where a Chunker instance may be canceled before the executor service has a chance to even start it. When that happened, an event would be lost.

### Checklist


mantis-network/src/test/java/io/reactivex/mantis/network/push/TimedChunkerTest.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        @Override
+        public void process(ConnectionManager<T> connectionManager, List<T> chunks) {
+            if (processingTimeMs > 0) {
+                ScheduledFuture<Boolean> f = scheduledService.schedule(() -> true, processingTimeMs,
+                        TimeUnit.MILLISECONDS);
+                    while (!f.isDone()) {
+                    }
+            }
```
Andyz26, line 180: Redundent?

Andyz26 (approved):

mantis-network/src/test/java/io/reactivex/mantis/network/push/TimedChunkerTest.java
calvin681: This is needed to emulate long processing time. And we cannot just use Thread.sleep either, we need the thread to keep running.

PR #113, opened 2021-09-15
Change meaning of resubmission limit to take time duration into account

calvin681:
### Context

Change meaning of resubmission limit to take time duration into account. For long running jobs that may run for months or years, worker resubmissions will eventually reach the limit. We don't want to penalize such jobs. We should only terminate jobs where workers are resubmitted constantly. That may be due to bad code or external dependency failures.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
@@ -43,6 +43,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import akka.actor.AbstractActorWithTimers;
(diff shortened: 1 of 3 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
```diff
@@ -2232,7 +2239,7 @@ private void resubmitWorker(JobWorker oldWorker) throws Exception {
             Map<Integer, Integer> workerToStageMap = mantisJobMetaData.getWorkerNumberToStageMap();
 
             IMantisWorkerMetadata oldWorkerMetadata = oldWorker.getMetadata();
-            if (oldWorkerMetadata.getTotalResubmitCount()
+            if (recentlyResubmittedWorkersCache.size()
```
Andyz26, line 2242: From the [doc](https://guava.dev/releases/15.0/api/docs/com/google/common/cache/CacheBuilder.html#expireAfterWrite(long,%20java.util.concurrent.TimeUnit)) size() could be inaccurate, what about trigger cleanup first?

Andyz26 (approved):

PR #114, opened 2021-09-17
Andyz/submit with scaling

Andyz26:
### Context

Currently, when a running needs to be re-submitted via UI/API, there is no direct way to inherit the currently running job's worker instance at each stage. This creates the problem for jobs with auto/manual scaling enabled since the accurate worker scale is not inherited from scaling operations and the newly submitted job will fall back to cluster default settings.

The PR here adds a new flag into the StageSchedulingInfo contract object called "inheritInstanceCount" and a class builder for "StageSchedulingInfo" as well. Setting this flag to true in the builder will trigger the cluster actor to load the most recent job (same as quick submit mode) and inherit the previous instance count if applicable.

### Checklist


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -1895,6 +1908,55 @@ private long getExpireAcceptedDelayMs() {
...
+            if (lastSubmittedJobDefn.get().getSchedulingInfo() != null &&
+                    lastSubmittedJobDefn.get().getSchedulingInfo().getStages().size() == givenJobDefn.getSchedulingInfo().getStages().size()) {
+                // Use instance count from last submitted job at each stage.
+                Map<Integer, StageSchedulingInfo> givenStages = givenJobDefn.getSchedulingInfo().getStages();
+                Map<Integer, StageSchedulingInfo> lastStages = lastSubmittedJobDefn.get().getSchedulingInfo().getStages();
+                SchedulingInfo.Builder siBuilder = new SchedulingInfo.Builder().numberOfStages(givenStages.size());
+                givenStages.forEach((stageId, givenStageInfo) -> {
+                    if (lastStages.containsKey(stageId) && givenStageInfo.getInheritInstanceCount()) {
+                        StageSchedulingInfo mergedStageInfo = new StageSchedulingInfo.Builder()
+                                .setNumberOfInstances(lastStages.get(stageId).getNumberOfInstances())
```
calvin681, line 1937: Unfortunately, this doesn't give the current worker count for a stage. This only gives the worker count in job definition when the job was last submitted. The actual current worker count is dynamically changing from auto scaling. I think you would need the jobActor to get the up to date worker count.

Andyz26: @calvin681 Updated the actor logic and test. Could you take a look when you have a chance?

mantis-common/src/main/java/io/mantisrx/runtime/descriptor/StageSchedulingInfo.java
```diff
@@ -29,12 +29,13 @@
...
-    private MachineDefinition machineDefinition;
-    private List<JobConstraints> hardConstraints;
-    private List<JobConstraints> softConstraints;
+    private final int numberOfInstances;
+    private final MachineDefinition machineDefinition;
+    private final List<JobConstraints> hardConstraints;
+    private final List<JobConstraints> softConstraints;
     private StageScalingPolicy scalingPolicy;
-    private boolean scalable;
+    private final boolean inheritInstanceCount;
```
sundargates, line 37: I am wondering if it makes sense semantically for inheritInstanceCount to be an attribute of the StageSchedulingInfo? Shouldn't this be more of an attribute of how the deployments are rolled out? Something like a DeploymentStrategy class and this being a member of that?

Disclaimer: I do not know a lot about this codebase, so many of these questions are based on semantics rather than specifics.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -1332,20 +1332,31 @@ public void onJobSubmit(final SubmitJobRequest request) {
...
             Optional<JobDefinition> jobDefnOp = createNewJobDefinitionFromLastSubmittedInheritSchedInfoAndParameters(existingJobsList, jobManager.getCompletedJobsList(), empty(),jobStore);
             if(jobDefnOp.isPresent()) {
                 logger.info("Inherited scheduling Info and parameters from previous job");
                 resolvedJobDefn = jobDefnOp.get();
             } else {
                 throw new Exception("Job Definition could not retrieved from a previous submission (There may not be a previous submission)");
             }
-
+        }
+        else if (givenJobDefnOp.get().getSchedulingInfo() != null && givenJobDefnOp.get().getSchedulingInfo().requireInheritInstanceCheck()) {
```
sundargates, line 1346: nit: Did you considering making requireInheritanceCheck a property of JobDefinition instead of SchedulingInfo?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -1895,6 +1906,38 @@ private long getExpireAcceptedDelayMs() {
         return empty();
     }
 
+
+    private Optional<JobDefinition> createJobDefinitionInheritingInstanceCount(
+            final JobDefinition givenJobDefn,
+            final List<JobInfo> existingJobsList) {
+        if(logger.isTraceEnabled()) { logger.trace("Enter createJobDefinitionInheritingInstanceCount"); }
```
sundargates, line 1913: do we need this or was this meant for debugging?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -1895,6 +1906,38 @@ private long getExpireAcceptedDelayMs() {
...
+
+        // use last submitted job (if non-terminal) to merge with the givenJobDefn so the result jobDefn will inherit instance count.
+        JobDefinition finalJobDefn = givenJobDefn;
+        Optional<JobId> lastJobId = JobListHelper.getLastSubmittedJobId(existingJobsList, Lists.newArrayList());
+        if (lastJobId.isPresent())
+        {
```
sundargates, line 1919: nit: move the brace to the previous line

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -2162,6 +2223,71 @@ private void updateAndSaveJobCluster(IJobClusterMetadata jobCluster) throws Exce
...
+                            }
+                            else {
+                                if (k == 0) { // handle JobMaster stage
+                                    siBuilder.addJobMasterStage(givenStages.get(k));
+                                }
+                                else { siBuilder.addStage(givenStages.get(k)); }
+                            }
+                        });
+
+                        jobDefnBuilder.withSchedulingInfo(siBuilder.build());
```
sundargates, line 2278: I'm guessing this block deals with merging the provided job definition and the job definition of the currently running job. I would move this logic to JobDefinition::merge or JobDefinition::concat. This would also allow you to easily unit test this logic.

mantis-common/src/main/java/io/mantisrx/runtime/descriptor/StageSchedulingInfo.java
```diff
@@ -29,12 +29,13 @@
...
 
-    private int numberOfInstances;
-    private MachineDefinition machineDefinition;
-    private List<JobConstraints> hardConstraints;
-    private List<JobConstraints> softConstraints;
+    private final int numberOfInstances;
+    private final MachineDefinition machineDefinition;
+    private final List<JobConstraints> hardConstraints;
+    private final List<JobConstraints> softConstraints;
     private StageScalingPolicy scalingPolicy;
```
sundargates, line 36: why is this not final?

mantis-common/src/main/java/io/mantisrx/runtime/descriptor/StageSchedulingInfo.java
```diff
@@ -150,4 +159,69 @@ public boolean equals(Object obj) {
...
+            this.hardConstraints = hardConstraints;
+            return this;
+        }
+
+        public Builder setOptionalSoftConstraints(List<JobConstraints> softConstraints) {
+            this.softConstraints = softConstraints;
+            return this;
+        }
+
+        public Builder setOptionalScalingPolicy(StageScalingPolicy scalingPolicy) {
```
sundargates, line 192: This feels unrelated to this diff. I see a lot of tests have changed as a result of this change. In general, I think it's a good idea to keep refactoring changes isolated from business logic changes. I would put this up as a separate diff. We can also explore using Lombok annotations for simplifying a lot of this code in that diff.

mantis-common/src/main/java/io/mantisrx/runtime/descriptor/SchedulingInfo.java
```diff
@@ -106,26 +110,47 @@ public String toString() {
...
         public Builder numberOfStages(int numberOfStages) {
             this.numberOfStages = numberOfStages;
             return this;
         }
 
-        public Builder singleWorkerStageWithConstraints(MachineDefinition machineDefinition,
-                                                        List<JobConstraints> hardConstraints, List<JobConstraints> softConstraints) {
-            builderStages.put(currentStage, new StageSchedulingInfo(1, machineDefinition, hardConstraints,
-                    softConstraints, null, false));
+        public Builder singleWorkerStageWithConstraints(
```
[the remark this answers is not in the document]
Andyz26: Yeah, the naming can definitely be improved around some of these classes. I added a new issue tracking this one at 
 #118 as the refactor will affect many more unrelated files.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
Andyz26: I see this is the widely used pattern in mantis codebase to provide more detailed debugging traces. I think by default it's turned off.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
Andyz26: I feel that IMantisJobMetadata logically should not be exposed to the public JobDefinition class. I will convert it to a plain map then.

mantis-common/src/main/java/io/mantisrx/runtime/descriptor/StageSchedulingInfo.java
Andyz26: There is an existing setter method but I think it's not used. Will remove.

mantis-common/src/main/java/io/mantisrx/runtime/descriptor/StageSchedulingInfo.java
Andyz26: I used the auto builder generator from IntelliJ thus it auto-converted all the usages.
Definitely +1 on lambok.

mantis-common/src/main/java/io/mantisrx/runtime/descriptor/StageSchedulingInfo.java
Andyz26: Screen Shot 2021-09-22 at 12 12 03 PM

Andyz26: Note: this new DeploymentStrategy class is to provide better logical OO abstraction/isolation at the class level and it is not intended to extend the mantis master for large/complex deployment workflow.

build.gradle
```diff
@@ -24,6 +24,7 @@ buildscript {
     dependencies {
         classpath 'com.netflix.nebula:gradle-netflixoss-project-plugin:9.4.2'
         classpath 'io.mantisrx:mantis-gradle-plugin:1.2.+'
+        classpath "io.freefair.gradle:lombok-plugin:5.3.3.3"
```
calvin681, line 27: Instead of using a plugin, can we just use a compileOnly dependency. See mantis-example-core/build.gradle for example.

Andyz26: The plugin seems to be the recommended method. Can you share your concerns here?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -2162,6 +2223,71 @@ private void updateAndSaveJobCluster(IJobClusterMetadata jobCluster) throws Exce
...
+                        try {
+                            return of(jobDefnBuilder.build());
+                        }
+                        catch (InvalidJobException ije) {
+                            logger.error("Failed to build job definition with inheritance:", ije);
+                            return empty();
+                        }
+                    }
+                });
```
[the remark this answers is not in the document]
Andyz26: Technically this is all in-memory access to another object in the same process so the timeout should be redundant. I will take a deeper look at this interaction.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -1895,6 +1900,37 @@ private long getExpireAcceptedDelayMs() {
...
+                } catch (InterruptedException | ExecutionException | TimeoutException e) {
+                    logger.error("Failed to merge with running job's instance", e);
+                }
+            }
+        }
+        else {
+            logger.info("No running job instance available to inherit stage worker instance {}", givenJobDefn.getName());
+        }
+
+        return createNewJobDefinitionInheritSchedInfoAndParameters(finalJobDefn);
```
calvin681, line 1931: I think this should just return givenJobDefn. This should behave the same as if inherit count is not set.

mantis-common/src/main/java/io/mantisrx/runtime/descriptor/SchedulingInfo.java
```diff
@@ -135,27 +180,73 @@ public Builder multiWorkerScalableStageWithConstraints(int numberOfWorkers, Mach
...
+                    StageSchedulingInfo.builder()
+                            .numberOfInstances(numberOfWorkers)
+                            .machineDefinition(machineDefinition)
+                            .scalable(scalable)
+                            .build());
             currentStage++;
             return this;
         }
 
+        public Builder createWithInstanceInheritance(
```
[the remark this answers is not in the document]
Andyz26: The scheduling builder is a bit counter-intuitive because it has to set the number of stages separately from "add*" calls. When using this method writer needs to call the set stage number again to adjust the stage count. I will add java doc on this one.

build.gradle
calvin681: I don't want to maintain 2 different convention of pulling in a library within the same code base. I think the consistency helps reduce confusion in the future.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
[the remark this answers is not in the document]
nickmahilani: `ask`/Future.get in Akka is generally an anti pattern as it is essentially blocking the actor from making progress on pending messages in its mailbox till the ask completes. If you could rework this to use 'tell' instead that would keep this actor processing its mailbox without blocking for response from JobActor.

If in the worst case, this operation does take 500ms (yes it is in-memory but if the Job Actor mailbox is backlogged for some reason, this request could potentially take a long time), the JobClusterActor is essentially blocked for that entire duration and won't be processing any other messages during that time.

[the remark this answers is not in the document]
calvin681: Going another direction, you can explore if we can keep up to date worker count in jobManager. JobManager already does some caching of job information on active jobs, completed jobs, etc. Should it also track worker count?

Though I'm not sure if this approach would end up being more or less complicated than using `tell`, callbacks.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
Andyz26: Renamed and updated the logic.

build.gradle
Andyz26: I tried removing the old ref in the example folder and it seems to be working fine in the build.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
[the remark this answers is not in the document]
Andyz26: I feel that syncing/maintaining the count info state of each job at two different actors seem a bit costly for this purpose and the potential inconsistency to have a duplicate state might be a headache to deal w/ in other cases. 
I am thinking of maybe adding behavior to the job actor to act on the unfinished job request as well as resume behavior on the cluster actor to resume the operation to make it more async. I will try and see whether that works.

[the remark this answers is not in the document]
Andyz26: @calvin681 @nickmahilani I updated the PR to use tell between the interactions. Could you take another look when you have time?

build.gradle
Andyz26: @calvin681 @nickmahilani I did the runtime dependency comparison between this build and the master and there is no runtime dependency introduced by lombok. (I tried common and runtime folder.)

```
runtimeClasspath - Runtime classpath of source set 'main'.
+--- project :mantis-remote-observable
|    +--- project :mantis-common
|    |    +--- project :mantis-shaded
|    |    +--- com.netflix:mantis-rxnetty:0.4.19.1
|    |    |    +--- io.reactivex:rxjava:1.0.17 -> 1.3.8
|    |    |    +--- io.netty:netty-handler:4.1.5.Final -> 4.1.17.Final
|    |    |    |    +--- io.netty:netty-buffer:4.1.17.Final
|    |    |    |    |    \--- io.netty:netty-common:4.1.17.Final
|    |    |    |    +--- io.netty:netty-transport:4.1.17.Final
|    |    |    |    |    +--- io.netty:netty-buffer:4.1.17.Final (*)
|    |    |    |    |    \--- io.netty:netty-resolver:4.1.17.Final
|    |    |    |    |         \--- io.netty:netty-common:4.1.17.Final
|    |    |    |    \--- io.netty:netty-codec:4.1.17.Final
|    |    |    |         \--- io.netty:netty-transport:4.1.17.Final (*)
|    |    |    +--- io.netty:netty-codec-http:4.1.5.Final -> 4.1.17.Final
|    |    |    |    \--- io.netty:netty-codec:4.1.17.Final (*)
|    |    |    +--- io.netty:netty-transport-native-epoll:4.1.5.Final -> 4.1.17.Final
|    |    |    |    +--- io.netty:netty-common:4.1.17.Final
|    |    |    |    +--- io.netty:netty-buffer:4.1.17.Final (*)
|    |    |    |    +--- io.netty:netty-transport-native-unix-common:4.1.17.Final
|    |    |    |    |    +--- io.netty:netty-common:4.1.17.Final
|    |    |    |    |    \--- io.netty:netty-transport:4.1.17.Final (*)
|    |    |    |    \--- io.netty:netty-transport:4.1.17.Final (*)
|    |    |    \--- org.slf4j:slf4j-api:1.7.6 -> 1.7.32
|    |    +--- io.reactivex:rxjava:1.3.8
|    |    +--- io.netty:netty-codec-http:4.1.17.Final (*)
|    |    +--- io.netty:netty-buffer:4.1.17.Final (*)
|    |    +--- io.netty:netty-transport-native-epoll:4.1.17.Final (*)
|    |    +--- com.netflix.numerus:numerus:1.+ -> 1.1
|    |    +--- org.xerial.snappy:snappy-java:1.1.+ -> 1.1.8.4
|    |    +--- org.jctools:jctools-core:1.+ -> 1.2.1
|    |    +--- com.netflix.spectator:spectator-api:0.92.+ -> 0.92.2
|    |    |    \--- org.slf4j:slf4j-api:1.7.26 -> 1.7.32
|    |    +--- org.slf4j:slf4j-api:1.7.+ -> 1.7.32
|    |    \--- org.slf4j:slf4j-log4j12:1.7.+ -> 1.7.32
|    |         +--- org.slf4j:slf4j-api:1.7.32
|    |         \--- log4j:log4j:1.2.17
|    +--- io.netty:netty-handler:4.1.17.Final (*)
|    +--- com.netflix.rxjava:rxjava-math:0.20.6
|    +--- org.slf4j:slf4j-api:1.7.+ -> 1.7.32
|    \--- org.slf4j:slf4j-log4j12:1.7.+ -> 1.7.32 (*)
+--- project :mantis-network
|    +--- io.netty:netty-handler:4.1.17.Final (*)
|    +--- io.mantisrx:mql-jvm:3.2.+ -> 3.2.9
|    |    \--- io.reactivex:rxjava:1.3.8
|    \--- project :mantis-common (*)
+--- project :mantis-common (*)
\--- org.slf4j:slf4j-api:1.7.+ -> 1.7.32
```

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -1325,31 +1420,32 @@ public void onJobSubmit(final SubmitJobRequest request) {
      * 2. If JobDefinition is not provided, find the last submitted job and use its config (quick submit)
      * @param user submitter
      * @param givenJobDefnOp job defn provided by user in job submit
-     * @return jobdefinition to be used by the actual submit
+     * @return job definition to be used by the actual submit
      * @throws Exception If jobDefinition could not be resolved
      */
-
-
     private JobDefinition getResolvedJobDefinition(final String user, final Optional<JobDefinition> givenJobDefnOp) throws Exception {
```
calvin681, line 1426: I think you can completely revert this method now.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -1889,12 +1985,46 @@ private long getExpireAcceptedDelayMs() {
...
             if(logger.isTraceEnabled()) { logger.trace("Exit createNewJobDefinitionFromLastSubmittedInheritSchedInfoAndParameters"); }
-            return createNewJobDefinitionInheritSchedInfoAndParameters(lastSubmittedJobDefn.get());
+            return cloneToNewJobDefinitionWithoutArtifactNameAndVersion(lastSubmittedJobDefn.get());
         }
         if(logger.isTraceEnabled()) { logger.trace("Exit createNewJobDefinitionFromLastSubmittedInheritSchedInfoAndParameters empty"); }
         return empty();
     }
 
+
+    private Optional<JobDefinition> createJobDefinitionInheritingInstanceCount(
```
calvin681, line 1995: should not be needed anymore

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -2162,6 +2286,50 @@ private void updateAndSaveJobCluster(IJobClusterMetadata jobCluster) throws Exce
...
 
 
+    /*
+     * Use running job instance to merge with given job definition for each stage's worker instance count.
+     * @param runningJobInfo the running job instance with ref to its job actor.
+     * @param givenJobDefn given job definition.
+     * @param forceInheritance whether to force instance inheritance to all stages.
+     * @return CompletionStage<Optional<JobDefinition>> CompletionStage containing the merged job definition.
+     */
+    private CompletionStage<Optional<JobDefinition>> getNonTerminalJobDefnWithWorkerCountInheritance(
```
calvin681, line 2296: Can remove this as well

mantis-common/src/main/java/io/mantisrx/runtime/descriptor/StageDeploymentStrategy.java
```diff
@@ -0,0 +1,40 @@
...
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.ToString;
+
+@Builder
+@ToString()
+@Getter
+public class StageDeploymentStrategy {
```
calvin681, line 29: You would need `equals` and `hashcode` in this class as well right? BTW, lombok has an annotation that can generate that for you.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -2122,37 +2224,31 @@ private void updateAndSaveJobCluster(IJobClusterMetadata jobCluster) throws Exce
...
-                        Optional<IMantisJobMetadata> archivedJob = store.getArchivedJob(completedJob.get().getJobId());
-                        if(archivedJob.isPresent()) {
-                            if(logger.isTraceEnabled()) { logger.trace("Exit getLastSubmittedJobDefinition returning job {} with defn {}", archivedJob.get().getJobId(), archivedJob.get().getJobDefinition()); }
-                            return of(archivedJob.get().getJobDefinition());
-                        } else {
-                            logger.warn("Could not find load archived Job {} for cluster {}", completedJob.get().getJobId(), name);
-                        }
-                    } catch (Exception e) {
-                        logger.warn("Archived Job {} could not be loaded from the store due to {} ", completedJob.get().getJobId(), e.getMessage());
+            Optional<CompletedJob> completedJob = jobManager.getCompletedJob(lastJobId.get());
```
calvin681, line 2237: why is this changed to completedJob?

Andyz26: This is the quick submit path and for jobs with active job actors, it's passed to job actors instead of reaching here.
I will update the java doc on this method.

calvin681 (approved):

PR #133, opened 2022-02-04
Rename defaultStream to __default__.

codyrioux:
We experienced a bug in production in which the `mreAppJobClusterMap`
uses `__default__` as the routing key when a stream does not have a
routing entry. These routing keys are internally used as stream names,
however internally mantis-published used `defaultStream`. This commit
reconciles these two by renaming defaultStream to `__default__`.
`
More work to come in the future to avoid conflation of routing keys and
stream names.

### Context

### Checklist

### DIFF
--- a/mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/api/StreamType.java
+++ b/mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/api/StreamType.java
@@ -21,7 +21,6 @@ public class StreamType {
     /**
      * Default Stream name for emitting events.
      */
-    //public static final String DEFAULT_EVENT_STREAM = "defaultStream";
 	public static final String DEFAULT_EVENT_STREAM = "__default__";
 
     /**
--- a/mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/api/StreamType.java
+++ b/mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/api/StreamType.java
@@ -21,7 +21,8 @@ public class StreamType {
     /**
      * Default Stream name for emitting events.
      */
-    public static final String DEFAULT_EVENT_STREAM = "defaultStream";
+    //public static final String DEFAULT_EVENT_STREAM = "defaultStream";
+	public static final String DEFAULT_EVENT_STREAM = "__default__";
 
     /**
      * Stream name for emitting request events.
--- a/mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/internal/mql/MQLSubscription.java
+++ b/mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/internal/mql/MQLSubscription.java
@@ -123,7 +123,7 @@ public class MQLSubscription implements Subscription, Comparable {
                 .map((String s) -> {
                     if (s.toLowerCase().equals("observable") ||
                             s.toLowerCase().equals("stream") ||
-							s.toLowerCase().equals("defaultstream")) {
+                            s.toLowerCase().equals("defaultstream")) {
                         // Translate the legacy default stream names to map to the default stream.
                         return StreamType.DEFAULT_EVENT_STREAM;
                     } else {
--- a/mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/internal/mql/MQLSubscription.java
+++ b/mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/internal/mql/MQLSubscription.java
@@ -18,6 +18,7 @@ package io.mantisrx.publish.internal.mql;
 
 import io.mantisrx.mql.jvm.core.Query;
 import io.mantisrx.publish.api.Event;
+import io.mantisrx.publish.api.StreamType;
 import io.mantisrx.publish.core.Subscription;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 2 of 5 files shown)


mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/api/StreamType.java
```diff
@@ -21,7 +21,8 @@
     /**
      * Default Stream name for emitting events.
      */
-    public static final String DEFAULT_EVENT_STREAM = "defaultStream";
+    //public static final String DEFAULT_EVENT_STREAM = "defaultStream";
+	public static final String DEFAULT_EVENT_STREAM = "__default__";
```
calvin681, line 25: spacing is off

mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/internal/mql/MQLSubscription.java
```diff
@@ -117,8 +118,26 @@ public boolean matches(Event event) {
         }
     }
 
+	private List<String> sanitizeStreamSubjects(final List<String> subjects) {
+        return subjects.stream()
+                .map((String s) -> {
+                    if (s.toLowerCase().equals("observable") ||
+                            s.toLowerCase().equals("stream") ||
+							s.toLowerCase().equals("defaultstream")) {
```
calvin681, line 126: spacing if off

mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/api/StreamType.java
```diff
@@ -21,7 +21,8 @@
     /**
      * Default Stream name for emitting events.
      */
-    public static final String DEFAULT_EVENT_STREAM = "defaultStream";
+    //public static final String DEFAULT_EVENT_STREAM = "defaultStream";
```
calvin681, line 24: just remove the line

mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/internal/mql/MQLSubscription.java
codyrioux: Interesting. Looked the same in my editor but line 125 has spaces and 126 had tabs. I'm going to need to configure my editor to adhere to the formatting rules clearly.

codyrioux:
Ultimately the configuration map is a map of stream names or routing keys to source jobs used by `mantis-publish` to route events to a source job, and used by our more modern sources to subscribe to streams. This was true in all cases except the default where the routing key was `__default__` and the stream name was `defaultStream`. What this patch does is renames the default stream to `__default__` so as to ensure the routing keys are always stream names.

The `__default__` constant shows up in three different locations (see below). We'll want to unify this to a single constant to prevent them from getting out of sync.

```bash
~/src/github/mantis % rg "__default__"                                                                                                                                                                                                                                                                                                                                                                                                                             (git)-[fix-routing-keys-stream-names]
docs/docs/reference/api.md
807:            "__default__": {
809:              "__default__": "SharedMantisPublishEventSource"

mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/api/StreamType.java
24:	public static final String DEFAULT_EVENT_STREAM = "__default__";

mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/internal/discovery/MantisJobDiscoveryStaticImpl.java
52:            + "{\"__default__\": {\"requestEventStream\": \"SharedPushRequestEventSource\","
53:                                + "\"__default__\": \"" + DEFAULT_JOB_CLUSTER + "\"}}}";

mantis-discovery-proto/src/test/java/com/netflix/mantis/discovery/proto/AppJobClustersMapTest.java
80:        final String mapping = "{\"version\":\"1\",\"timestamp\":2,\"mappings\":{\"testApp\":{\"__default__\":\"SharedMrePublishEventSource\"}}}";

mantis-discovery-proto/src/main/java/com/netflix/mantis/discovery/proto/StreamJobClusterMap.java
29:    public static final String DEFAULT_STREAM_KEY = "__default__";

mantis-discovery-proto/src/main/java/com/netflix/mantis/discovery/proto/AppJobClustersMap.java
33:    public static final String DEFAULT_APP_KEY = "__default__";
```

mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/internal/mql/MQLSubscription.java
sundargates: interesting that the formatter that we added recently isn't catching this issue?

mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/internal/mql/MQLSubscription.java
```diff
@@ -117,8 +118,26 @@ public boolean matches(Event event) {
         }
     }
 
+	private List<String> sanitizeStreamSubjects(final List<String> subjects) {
+        return subjects.stream()
+                .map((String s) -> {
+                    if (s.toLowerCase().equals("observable") ||
+                            s.toLowerCase().equals("stream") ||
+                            s.toLowerCase().equals("defaultstream")) {
```
sundargates, line 126: nit: Could we introduce a new Stream DS (or a Subject DS, not sure what to call it) that has properties around whether it's a default stream or not for instance? We can also expose that through the contract for the Subscription class instead of just returning a list of strings.

calvin681 (approved):

nickmahilani:
The`defaultStream` is the stream name to which events are published by default by the mantis publisher. The `__default__` in the mreAppJobClusterMap is the default routing key and indicates the default source job to use for any stream name that does not have an explicit override in the mreAppJobClusterMap. Rolling out this change safely is non trivial as connectors, source jobs and publshing apps should have the same understanding of default stream else the data flow could be affected (Had to tell from the diff but MQL subscription handling for example is treating `defaultStream` the same as `stream` to handle legacy `SELECT * FROM stream` queries and might start breaking if we change this to `__default__`)

codyrioux:
To pull in some context: We're in a situation right now where cloud gateway
cannot deploy until we deploy a fix. They can't pin back because of a GRPC
compatibility issue and they can't use our current release because of the
performance. The current release recompiles every MQL query on every query
refresh. We're causing a ~10% CPU increase on Zuul.

It does however seem as though the nfmantis-publish* tests are unhappy
about the change. I'm going to revert this one, as well as the commit that
caused the performance regression. We'll figure out next steps next week.

PR #134, opened 2022-02-04
Disabling Checkstyle Errors

sundargates:
### Context

A lot of the mantis modules today fail check style. In addition to that, the number of such violations in some of these modules is so high that it would take us some time to fix them in the first place. In the meantime, these checks eat a lot of the build time. Thus, removing them till we allot some time to revisit the overall code quality. 

### Checklist


codyrioux (approved):
Seems straightforward enough. I assume we would re-enable if and only if we were to commit to bringing the entire repository into compliance.

PR #137, opened 2022-02-07
Code minor cleanup in mantis-client module

liuml07:
### Context

Do some code cleanup during my code walk. Goal is not to change any functional logic but refine it in a clearer/cleaner way.

### Checklist

- ~[ ] Added new tests where applicable~
- ~[ ] Extended README or added javadocs where applicable~
- ~[ ] Added copyright headers for new files from `CONTRIBUTING.md`~

### DIFF
--- a/mantis-client/src/main/java/io/mantisrx/client/MantisClient.java
+++ b/mantis-client/src/main/java/io/mantisrx/client/MantisClient.java
@@ -52,7 +52,6 @@ public class MantisClient {
         public Observable<EndpointChange> locatePartitionedSinkForJob(final String jobId, final int forPartition, final int totalPartitions) {
             return clientWrapper.getMasterClientApi()
                     .flatMap((MantisMasterClientApi mantisMasterClientApi) -> {
-                        Integer sinkStage = null;
                         return mantisMasterClientApi.getSinkStageNum(jobId)
                                 .take(1) // only need to figure out sink stage number once
                                 .flatMap((Integer integer) -> {
@@ -96,7 +95,7 @@ public class MantisClient {
      */
     public MantisClient(Properties properties) {
         clientWrapper = new MasterClientWrapper(properties);
-        this.disablePingFiltering = Boolean.valueOf(properties.getProperty(ENABLE_PINGS_KEY));
+        this.disablePingFiltering = Boolean.parseBoolean(properties.getProperty(ENABLE_PINGS_KEY));
     }
 
     public JobSinkLocator getSinkLocator() {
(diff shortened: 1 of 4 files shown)


mantis-client/src/main/java/io/mantisrx/client/MantisClient.java
```diff
@@ -207,7 +206,7 @@ public String getError() {
         return new SinkClientImpl<T>(jobId, sinkConnectionFunc, getSinkLocator(),
                 numSinkWrkrsSubject
                         .filter((jobSinkNumWorkers) -> jobId.equals(jobSinkNumWorkers.getJobId()))
-                        .map((jobSinkNumWorkers) -> jobSinkNumWorkers.getNumSinkWorkers()),
+                        .map(MasterClientWrapper.JobSinkNumWorkers::getNumSinkWorkers),
```
sundargates, line 209: I usually try to avoid this style because as soon as the class name changes, it results in a change in this file too unnecessarily. If it's a popular class (in this case, MantisClientWrapper is and will be replaced with another type in my rewrite), then the number of such references is high, resulting in many spurious changes.

mantis-client/src/main/java/io/mantisrx/client/SinkConnectionsStatus.java
```diff
@@ -16,27 +16,11 @@
...
-    }
-
-    public long getNumConnected() {
-        return numConnected;
-    }
-
-    public long getTotal() {
-        return total;
-    }
+@Value
```
sundargates, line 21: nit: this is a different change than the other changes in this diff. Let's add lombok annotations separately. The reason I'm slightly worried about lombok changes is that we do not know if these DS are being serialized by jackson somewhere. If we are planning on using jackson with these DS, then the lombok annotations will have to change or the lombok config has to change. It might be better to do these changes separately.

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/JobSubmitResponse.java
```diff
@@ -16,27 +16,12 @@
...
 
-    public boolean isFailed() {
-        return failed;
-    }
-
-    public String getErrorMessage() {
-        return errorMessage;
-    }
+@Value
+public class JobSubmitResponse {
```
sundargates, line 23: same as above.

mantis-client/src/main/java/io/mantisrx/client/MantisClient.java
```diff
@@ -96,7 +95,7 @@
      */
     public MantisClient(Properties properties) {
         clientWrapper = new MasterClientWrapper(properties);
-        this.disablePingFiltering = Boolean.valueOf(properties.getProperty(ENABLE_PINGS_KEY));
+        this.disablePingFiltering = Boolean.parseBoolean(properties.getProperty(ENABLE_PINGS_KEY));
```
sundargates, line 98: never knew the difference between the methods till I saw this. good change!

sundargates (commented):
I'm okay with all the changes except for the lombok ones. I would prefer to do that separately.

mantis-client/src/main/java/io/mantisrx/client/MantisClient.java
liuml07: sounds good

liuml07:
Latest code reverts the lombok change for the reason @sundargates explained. Also address another comments about method reference vs. lambda used as map function.

mantis-client/src/main/java/io/mantisrx/client/MantisClient.java
codyrioux: Huh. I also never knew the difference between these two until just now.

codyrioux:
I'm re-running the tests on this branch. Locally they ran fine so I believe the failure is just a flaky test.

sundargates (approved):
LGTM

mantis-client/src/main/java/io/mantisrx/client/MantisSSEJob.java
```diff
@@ -91,13 +90,7 @@ public String getJobId() {
...
     @Deprecated
     public synchronized Observable<MantisServerSentEvent> connectAndGetObservable() throws IllegalStateException {
-        return connectAndGet()
-                .flatMap(new Func1<Observable<MantisServerSentEvent>, Observable<MantisServerSentEvent>>() {
-                    @Override
-                    public Observable<MantisServerSentEvent> call(Observable<MantisServerSentEvent> o) {
-                        return o;
-                    }
-                });
+        return connectAndGet().flatMap(o -> o);
```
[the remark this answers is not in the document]
liuml07: In general it makes perfect sense. The Java built-in `Function::identity` is very stable across JDK versions since it's introduced. It is well defined and we can use it when it applies.

However, this `connectAndGet()` method returns an Observable object, which has its own `Func1` interface as the `flatMap(Func1)` parameter type. Sadly it's not interoperable with Java built-in `Function` interface. So we can not use it here :(

I did not find reactive Function(s)::identity method; and it is not necessarily better than `x ->x` even they provide one.

liuml07:
I will merge this by EoD if no more comments. Thanks!

PR #138, opened 2022-02-07
Enabling gradle cache for the whole repository in order to speed up the builds

sundargates:
### Context

Gradle cache is a fairly mature feature right now and has been battle-tested at Netflix and other places. Thus, enabling this feature in the mantis repo in the hope that it would help reduce build times. 

### Checklist


sundargates:
THis has reduced build time from 15+ minutes to a couple of minutes assuming we are able to leverage the build cache. 
image

codyrioux (approved):

PR #139, opened 2022-02-07
Fix missing subscription bug max subs change

codyrioux:
### Context
Unreverted the change to fix the max subscription bug. The issue with this code is that it does not recognize the stream named `__default__` coming from the job subscription map. Previously this was hidden by duplicate record keeping. The simplest possible change is to just treat `__default__` as `defaultStream` when looking up existing subscriptions. However this calls for a larger discussion around:

1. Sundaram's idea of introducing a data structure to represent Stream keys.
2. Calvin's proposal to remove `streamName` from `propagateSubscriptionChanges` altogether and simply have it manage all streams at once.

The relevant code is in 6f0dd7ce8aa0f82c3d865ce0cca17473167e8475.

### Checklist

### DIFF
--- a/mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/AbstractSubscriptionTracker.java
+++ b/mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/AbstractSubscriptionTracker.java
@@ -26,19 +26,25 @@ import io.mantisrx.publish.internal.discovery.MantisJobDiscovery;
 import io.mantisrx.publish.internal.metrics.SpectatorUtils;
 import io.mantisrx.publish.proto.MantisServerSubscription;
 import io.mantisrx.publish.proto.MantisServerSubscriptionEnvelope;
-import java.util.Collections;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 
+/**
+ * This class handles the logic for stream -> job cluster discovery. A class that extends this abstract class is
+ * expected to perform the actual fetch given a stream and cluster names. The storage of the subscriptions is off loaded
+ * to the stream manager (See {@link StreamManager}). Thus, the stream manager is the source of truth of current active
+ * subscriptions. For the same reason, the stream manager also stores which streams are registered. This class will only
+ * fetch subscriptions for streams that are registered.
+ */
 public abstract class AbstractSubscriptionTracker implements SubscriptionTracker {
     private static final Logger LOG = LoggerFactory.getLogger(AbstractSubscriptionTracker.class);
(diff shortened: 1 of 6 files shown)


jeffchao:
Nice find @codyrioux 😄

codyrioux:
Thanks @jeffchao but credit goes to Calvin for the original bug find, this one just fixes a performance issue with the fix.

I think I have the smallest possible change that will work here but am considering some larger changes this afternoon.

mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/AbstractSubscriptionTracker.java
```diff
@@ -59,148 +67,167 @@ public AbstractSubscriptionTracker(MrePublishConfiguration mrePublishConfigurati
...
+	protected Set<Subscription> getCurrentSubscriptions() {
+
+		return streamManager
+			.getRegisteredStreams()
+			.stream()
+			.flatMap(streamName -> streamManager.getStreamSubscriptions(streamName).stream())
+			.collect(Collectors.toSet());
+	}
+
+	// TODO: This is only present for unit tests, should be removed.
```
calvin681, line 223: We should just remove this now, and fix any tests that are using it. Those are tests are likely not testing the right assumptions anymore.

codyrioux:
Some flaky tests that don't seem to be failing under a local `gw clean build`.

codyrioux:
Does anyone know if there is a way to get at the `/home/runner/work/mantis/mantis/mantis-control-plane/mantis-control-plane-server/build/reports/tests/test/index.html` file so I can actually see what is happening on the CI server here?

calvin681:
Try download log archive to see if it contains the html file:

image

If you not, then only option is to try to reproduce locally.

codyrioux:
The failing unit tests were caused by the gradle cache change on the `main` branch. Reverting it here appears to have fixed it, all subsequent commits are hit/miss if we re-run the validation checks.

calvin681:
> The failing unit tests were caused by the gradle cache change on the `main` branch. Reverting it here appears to have fixed it, all subsequent commits are hit/miss if we re-run the validation checks.

@sundargates did you see this problem when you made the gradle cache change?

codyrioux:
It seems intermittent. I re-ran the [cache commit's validation](https://github.com/Netflix/mantis/runs/5192542050?check_suite_focus=true) three times with Attempt #2 and Attempt #3 failing but Attempt #4 succeeding. This branch passes with the revert.

codyrioux:
Just played around with caching / parallel disabled but could only get success with both disabled.

mantis-publish/mantis-publish-core/src/main/java/io/mantisrx/publish/AbstractSubscriptionTracker.java
```diff
@@ -59,148 +63,178 @@ public AbstractSubscriptionTracker(MrePublishConfiguration mrePublishConfigurati
...
+						}
+
+						allSubscriptions.addAll(subsEnvelope.getSubscriptions());
+						refreshSubscriptionSuccessCount.increment();
+					} else {
+						failedJobClusters.add(jobCluster);
+						refreshSubscriptionFailedCount.increment();
+					}
+				} catch (Exception ex) {
+					LOG.info("refresh subscriptions failed for {}", jobCluster, ex);
```
calvin681, line 208: add to `failedJobClusters` as well on exception?

PR #140, opened 2022-02-09
Cleaning up other github workflows to make use of the latest and greatest github actions versions

sundargates:
### Context

Similar to https://github.com/Netflix/mantis/pull/138, this PR updates other workflows so that they use 
1. the latest and greatest version of github actions plugins
2. use gradle cache appropriately. 

### Checklist

### DIFF
--- a/.github/workflows/nebula-publish.yml
+++ b/.github/workflows/nebula-publish.yml
@@ -12,36 +12,25 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


calvin681 (approved):

jeffchao:
See a test failure but those might be some of the non-deterministically-failing ones?

jeffchao (approved):

codyrioux (commented):
I think it is the cache change that is causing the non-determinism in tests. I rolled something similar back on the main branch a while ago. Have we figured out why this occurs?

PR #142, opened 2022-03-05
Only count resubmission from worker failures

calvin681:
### Context

If workers are resubmitted too frequently within a time period, the job is automatically terminated. However, there are scenarios where workers are supposed to be resubmitted, such as moving mesos clusters. We should only count resubmissions that are due to errors (worker is an error state and terminated unexpectedly).

### Checklist


codyrioux (approved):
LGTM

PR #145, opened 2022-03-20
[Part 1]: Support for hot task executors in Mantis

sundargates:
### Context

As we are approaching adding support for Mantis on Titus, one of the first goals is to be able to support task executors that are fully loaded (hot) and capable of executing Mantis stages on demand. This diff is the first step in that direction. This change introduces a new API on the server-side that establishes the protocol that the task executor needs to use to communicate with the server.

### Checklist

### DIFF
--- a/mantis-common/src/main/java/io/mantisrx/common/JsonSerializer.java
+++ b/mantis-common/src/main/java/io/mantisrx/common/JsonSerializer.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2022 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.mantisrx.common;
+
+import io.mantisrx.shaded.com.fasterxml.jackson.core.JsonProcessingException;
+import io.mantisrx.shaded.com.fasterxml.jackson.core.type.TypeReference;
+import io.mantisrx.shaded.com.fasterxml.jackson.databind.DeserializationFeature;
+import io.mantisrx.shaded.com.fasterxml.jackson.databind.ObjectMapper;
+import io.mantisrx.shaded.com.fasterxml.jackson.databind.SerializationFeature;
+import io.mantisrx.shaded.com.fasterxml.jackson.databind.ser.FilterProvider;
+import io.mantisrx.shaded.com.fasterxml.jackson.databind.ser.impl.SimpleFilterProvider;
+import io.mantisrx.shaded.com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+import java.io.IOException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class JsonSerializer {
+    private static final Logger logger = LoggerFactory.getLogger(JsonSerializer.class);
+
+    private static final ObjectMapper defaultObjectMapper = new ObjectMapper()
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+            .configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false)
+            .registerModule(new Jdk8Module());
+    public static final SimpleFilterProvider DEFAULT_FILTER_PROVIDER;
+
+    static {
+        DEFAULT_FILTER_PROVIDER = new SimpleFilterProvider();
+        DEFAULT_FILTER_PROVIDER.setFailOnUnknownId(false);
+    }
+
+    private static String toJSON(
+            ObjectMapper mapper,
+            FilterProvider filters,
+            Object object) throws JsonProcessingException {
+        if (filters == null) {
+            filters = DEFAULT_FILTER_PROVIDER;
+        }
+        return mapper.writer(filters).writeValueAsString(object);
+
+    }
+
+    public static <T> T fromJSON(
+            ObjectMapper mapper,
+            String json,
+            TypeReference<T> expectedType) throws IOException {
+        return mapper.readerFor(expectedType).readValue(json);
+    }
+
+    public static <T> T fromJSON(
+            ObjectMapper mapper,
+            String json,
+            Class<T> expectedType) throws IOException {
+        return mapper.readerFor(expectedType).readValue(json);
+    }
+
+    public <T> T fromJSON(String json, Class<T> expectedType) throws IOException {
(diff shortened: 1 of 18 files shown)


mantis-common/src/main/java/io/mantisrx/common/JsonSerializer.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public <T> T fromJSON(String json, Class<T> expectedType) throws IOException {
+        return defaultObjectMapper.readerFor(expectedType).readValue(json);
+    }
+
+    public <T> T fromJSON(String json, TypeReference<T> expectedType) throws IOException {
+        return defaultObjectMapper.readerFor(expectedType).readValue(json);
+    }
+
+    public String toJson(Object object) throws IOException {
```
hmitnflx, line 77: nit: I'd call out the semantic difference between this and `static toJson` in comments. 
Also, could we make this static?

hmitnflx (approved):

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ClusterID.java
```diff
@@ -0,0 +1,33 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.Value;
+
+/**
+ * Represents the cluster that the task executor belongs to.
+ */
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+@Value
+public class ClusterID {
+  String resourceID;
```
calvin681, line 28: this should be private?

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorDisconnection.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package io.mantisrx.server.master.resourcecluster;
+
+import lombok.Value;
+
+/**
+ * Datastructure representing the task executor that needs to be disconnected from the resource cluster.
+ */
+@Value
```
calvin681, line 24: constructor?

mantis-control-plane/mantis-control-plane-core/src/test/java/io/mantisrx/server/master/resourcecluster/TestTaskExecutorRegistration.java
```diff
@@ -0,0 +1,78 @@
...
+ * limitations under the License.
+ */
+package io.mantisrx.server.master.resourcecluster;
+
+import static org.junit.Assert.assertEquals;
+
+import io.mantisrx.common.JsonSerializer;
+import org.junit.Test;
+
+public class TestTaskExecutorRegistration {
```
calvin681, line 23: nit: we should follow convention. call this `TaskExecutorRegistrationTest`. Move the heartbeat test to `TaskExecutorHeartbeatTest`.

mantis-common/src/main/java/io/mantisrx/common/JsonSerializer.java
sundargates: I have removed all the static methods.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ClusterID.java
sundargates: value classes by default in Lombok treat the fields as "private final". 

Here's the documentation: https://projectlombok.org/features/Value

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorDisconnection.java
sundargates: value annotation creates an "allargsconstructor".

mantis-control-plane/mantis-control-plane-core/src/test/java/io/mantisrx/server/master/resourcecluster/TestTaskExecutorRegistration.java
sundargates: Renamed the class.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceClusterGateway.java
```diff
@@ -0,0 +1,57 @@
...
+import io.mantisrx.common.Ack;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Gateway for performing all actions corresponding to the resource cluster from the task executor.
+ */
+public interface ResourceClusterGateway {
+  /**
+   * triggered at the start of a task executor or when the task executor has lost connection to the resource cluster
+   * previously
```
Andyz26, line 27: nit: extra word?

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ClusterID.java
```diff
@@ -0,0 +1,33 @@
...
+
+  public static ClusterID of(String clusterIdStr) {
```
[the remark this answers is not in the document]
codyrioux: I really love the way static factory methods named `of` read in client code. `ClusterId.of("m5.2xl-1")` reads so cleanly to a user even compared to the builder `ClusterId.builder().name("m5.2xl-1").build()`. However I think this can be achieved with Lombok using `@RequiredArgsConstructor(staticName="of")`.

Andyz26 (approved):

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ClusterID.java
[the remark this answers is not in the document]
Andyz26: Very true. "of" is definitely cleaner on the single arg ctor case.

codyrioux (approved):
LGTM but would consider pushing the limits of what we can use Lombok for to decrease the verbosity. Ie: `@RequiredArgsConstructor(staticName="of")` for the `of` static factories.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ClusterID.java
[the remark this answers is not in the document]
codyrioux: That is actually a good point, it can get confusing on the multi-arg constructor case. Especially if parameters are of the same type, which string represents what?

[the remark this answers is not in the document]
sundargates: Thanks for the suggestion. I have used the Lombok annotation in both the places where it makes sense.

calvin681 (approved):

PR #148, opened 2022-03-22
Depend on remote artifacts for mantis-shaded instead of local subproject

hmitnflx:
### Context

Currently, we depend on the local copy of jar (subproject) for mantis-shaded dependency. However, this leads to problems with IntelliJ where you have to manually add the mantis-shaded subproject as an explicit library dependency at the root OR deal with a bunch of redlines for these shaded components.

We tried a bunch of alternatives like publishing to local maven or including these shaded dependencies at the terminal level (leaf projects) and they open-up more issues. We decided to host mantis-shaded in a separate repo but we think this could be a better alternative 
### Checklist

- [*] `./gradlew build` compiles code correctly
- [*] Added new tests where applicable
- [*] `./gradlew test` passes all tests
- [*] Added copyright headers for new files from `CONTRIBUTING.md`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,16 +34,17 @@ buildscript {
 }
 
 ext {
-    versionRxNetty = '0.4.19'
-    versionSuffixRxNettyShaded = '1'
-    versionIoNetty = '4.1.5.Final'
-    versionSlf4j = '1.7.0'
-    versionRxJava = '1.3.8'
-
     junitVersion = '4.11'
     mockitoVersion = '1.9.+'
     slf4jVersion = '1.7.+'
     vavrVersion = '0.9.2'
+
+    versionIoNetty = '4.1.5.Final'
+    versionMantisShaded = '1.3+'
+    versionRxJava = '1.3.8'
+    versionRxNetty = '0.4.19'
+    versionSlf4j = '1.7.0'
+    versionSuffixRxNettyShaded = '1'
 }
 
 allprojects {
@@ -66,6 +67,10 @@ subprojects {
         repositories {
             mavenLocal()
         }
+    }
+
+    configurations.all {
+        exclude group: 'ch.qos.logback', module: 'logback-classic'
     }
 
     repositories {
--- a/build.gradle
+++ b/build.gradle
@@ -70,6 +70,16 @@ subprojects {
     }
 
     configurations.all {
+        if (!"mantis-shaded".equals(project.name)) {
+            exclude group: 'com.fasterxml'
+            exclude group: 'io.vavr.jackson.datatype'
+            exclude group: 'com.google'
+            exclude group: 'org.apache.curator'
+            exclude group: 'org.apache.zookeeper'
+            exclude group: 'org.apache.jute'
+            exclude group: 'jline'
+            exclude group: 'org.jboss.netty'
+        }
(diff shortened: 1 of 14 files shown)


build.gradle
```diff
@@ -68,6 +69,10 @@ subprojects {
         }
     }
 
+    configurations.all {
+        exclude group: 'ch.qos.logback', module: 'logback-classic'
```
sundargates, line 73: you probably want all shaded dependencies here and explicitly filter mantis-shaded here.

build.gradle
```diff
@@ -34,16 +34,17 @@ buildscript {
...
-    versionSlf4j = '1.7.0'
-    versionRxJava = '1.3.8'
-
     junitVersion = '4.11'
     mockitoVersion = '1.9.+'
     slf4jVersion = '1.7.+'
     vavrVersion = '0.9.2'
+
+    versionIoNetty = '4.1.5.Final'
+    versionMantisShaded = '1.3+'
```
sundargates, line 43: should this be `latest.release`?

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/jobmaster/JobMasterService.java
```diff
@@ -105,8 +104,6 @@ private Measurements handleMetricEvent(final String ev) {
 
         } catch (JsonProcessingException e) {
             logger.error("failed to parse json", e);
-        } catch (IOException e) {
```
sundargates, line 108: Is this a spurious change?

sundargates (commented):
Overall looks great! Thanks for fixing this and coming up with the idea! I cannot wait to get this merged.

codyrioux (approved):

calvin681 (approved):

mantis-discovery-proto/build.gradle
```diff
@@ -15,7 +15,7 @@
  */
 
 dependencies {
-    api project(path: ':mantis-shaded', configuration: 'shadow')
+    api "io.mantisrx:mantis-shaded:${rootProject.versionMantisShaded}"
```
sundargates, line 18: this should be implementation probably?

hmitnflx: I'm doing a like-for-like replacement at the moment, will do this in a separate PR.

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/jobmaster/JobMasterService.java
hmitnflx: it was a warning during gradle build, will remove and move to a follow-up PR.

build.gradle
hmitnflx: umm, I'll keep it as it is.

build.gradle
```diff
@@ -68,6 +69,20 @@ subprojects {
         }
     }
 
+    configurations.all {
+        if (!"mantis-shaded".equals(project.name)) {
+            exclude group: 'com.fasterxml'
+            exclude group: 'io.vavr.jackson.datatype'
+            exclude group: 'com.google'
```
sundargates, line 76: do you want put the module name too?

sundargates (approved):
This is an awesome change!

PR #151, opened 2022-03-23
[Part 4]: Clean up worker ports

sundargates:
This diff has two purposes:
1. Establish the semantics of WorkerPorts and make sure all constructors lead to the same behavior (which was not the case before)
2. Make sure WorkerPorts is serializable by both Jackson and Java as it will be sent via wire using java serialization during RPC calls.

### Context

### Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -49,6 +49,7 @@ ext {
 
 ext.libraries = [
         asyncHttpClient: 'org.asynchttpclient:async-http-client:2.12.3',
+        commonsLang3: 'org.apache.commons:commons-lang3:3.5',
         flinkRpcApi: [
                 "org.apache.flink:flink-rpc-core:1.14.2"
         ],
--- a/mantis-common/src/main/java/io/mantisrx/common/WorkerPorts.java
+++ b/mantis-common/src/main/java/io/mantisrx/common/WorkerPorts.java
@@ -20,6 +20,7 @@ import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonCreator;
 import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonIgnore;
 import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonProperty;
+import io.mantisrx.shaded.com.google.common.collect.ImmutableList;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.HashSet;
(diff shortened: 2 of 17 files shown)


build.gradle
```diff
@@ -49,6 +49,7 @@ ext {
 
 ext.libraries = [
         asyncHttpClient: 'org.asynchttpclient:async-http-client:2.12.3',
+        commonsLang3: 'org.apache.commons:commons-lang3:3.5',
```
Andyz26, line 52: Do you think this one might need to be shaded?

mantis-common/src/main/java/io/mantisrx/common/WorkerPorts.java
```diff
@@ -43,39 +52,29 @@ public WorkerPorts(final List<Integer> assignedPorts) {
...
+        this.ports = new ArrayList<>(1);
+        ports.add(assignedPorts.get(4));
 
         if (!isValid()) {
             throw new IllegalStateException("worker validation failed on port allocation");
         }
     }
 
     public WorkerPorts(int metricsPort, int debugPort, int consolePort, int customPort, int sinkPort) {
-        this.ports = new ArrayList<>(5);
```
hmitnflx, line 60: could you pl refactor this to call the `WorkerPorts(List<Integer>);` constructor?
I'd also make `ports` an immutable because we return the value as-is in `getPorts`.

mantis-common/src/main/java/io/mantisrx/common/WorkerPorts.java
```diff
@@ -90,6 +89,11 @@ public WorkerPorts(@JsonProperty("metricsPort") int metricsPort,
         this.consolePort = consolePort;
         this.customPort = customPort;
         this.ports = ports;
+        this.sinkPort = ports.get(0);
```
hmitnflx, line 92: also, this one.

Andyz26 (approved):

hmitnflx (approved):

mantis-common/src/main/java/io/mantisrx/common/WorkerPorts.java
sundargates: +1

mantis-common/src/main/java/io/mantisrx/common/WorkerPorts.java
```diff
@@ -112,11 +101,12 @@ public int getCustomPort() {
 
     @JsonIgnore
     public List<Integer> getAllPorts() {
-        final List<Integer> allPorts = new ArrayList<>(ports);
+        final List<Integer> allPorts = new ArrayList<>();
```
nickmahilani, line 104: this diff changes the order of the sink port in the list, not sure how this function is used but it might break any code path that relies on sink port being first in list

nickmahilani (approved):

build.gradle
sundargates: I don't think so. The breaking API changes in this library are fairly minimum.

mantis-common/src/test/java/io/mantisrx/common/compression/CompressionUtilsTest.java
```diff
@@ -45,7 +46,7 @@
     }
 
     @Test public void shouldTokenizeWithEventsContainingPartialDelimiterMatchesWithCustomDelimiter() {
-        String delimiter = UUID.randomUUID().toString();
+        String delimiter = "a04f0418-bdff-4f53-af7d-9f5a093b9d65";
```
codyrioux, line 49: I'm curious why the fixed value here?

hmitnflx: This was to get around a bug fixed by #154

PR #154, opened 2022-03-25
Fix bug in CompressionUtils that fails with certain delimiters

calvin681:
### Context

There is a bug in CompressionUtils if the delimiter has a prefix that matches the end of a segment, the data is not properly tokenized.

### Checklist

### DIFF
--- a/mantis-common/src/main/java/io/mantisrx/common/compression/CompressionUtils.java
+++ b/mantis-common/src/main/java/io/mantisrx/common/compression/CompressionUtils.java
@@ -217,10 +217,19 @@ public class CompressionUtils {
             for (int i = 0; i < line.length(); i++) {
                 if (line.charAt(i) != delimiterArray[delimiterCount]) {
                     if (delimiterCount > 0) {
-                        for (int j = 0; j < delimiterCount; ++j) {
-                            sb.append(delimiterArray[j]);
+                        boolean prefixMatch = true;
+                        for (int j = delimiterCount - 1; j >= 0; j--) {
+                            if (line.charAt(i) != delimiterArray[j]) {
+                                prefixMatch = false;
+                                break;
+                            }
                         }
-                        delimiterCount = 0;
+                        if (!prefixMatch) {
+                            for (int j = 0; j < delimiterCount; ++j) {
+                                sb.append(delimiterArray[j]);
+                            }
+                            delimiterCount = 0;
+                        }
                     }
                     if (line.charAt(i) != delimiterArray[delimiterCount]) {
                         sb.append(line.charAt(i));
--- a/mantis-common/src/test/java/io/mantisrx/common/compression/CompressionUtilsTest.java
+++ b/mantis-common/src/test/java/io/mantisrx/common/compression/CompressionUtilsTest.java
@@ -19,14 +19,11 @@ import static org.junit.Assert.assertEquals;
 
 import io.mantisrx.common.MantisServerSentEvent;
 import java.io.BufferedReader;
-import java.io.IOException;
 import java.io.StringReader;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
-
-import org.junit.Assert;
 import org.junit.Test;
 
 public class CompressionUtilsTest {
--- a/mantis-common/src/test/java/io/mantisrx/common/compression/CompressionUtilsTest.java
+++ b/mantis-common/src/test/java/io/mantisrx/common/compression/CompressionUtilsTest.java
@@ -28,7 +28,8 @@ import org.junit.Test;
 
 public class CompressionUtilsTest {
 
-    @Test public void shouldTokenizeWithEventsContainingPartialDelimiterMatches() {
+    @Test
+    public void shouldTokenizeWithEventsContainingPartialDelimiterMatches() {
         String testInput = "ab$cdef$$$ghi$jkl$$$lmno$$pqrst$";
         try (BufferedReader reader = new BufferedReader(new StringReader(testInput))) {
             List<MantisServerSentEvent> result = CompressionUtils.tokenize(reader);
@@ -43,7 +44,8 @@ public class CompressionUtilsTest {
         }
     }
 
-    @Test public void shouldTokenizeWithEventsContainingPartialDelimiterMatchesWithCustomDelimiter() {
+    @Test
+    public void shouldTokenizeWithEventsContainingPartialDelimiterMatchesWithCustomDelimiter() {
         String delimiter = "a04f0418-bdff-4f53-af7d-9f5a093b9d65";
 
         String event1 = "ab" + delimiter.substring(0, 9) + "cdef";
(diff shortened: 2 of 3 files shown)


mantis-common/src/test/java/io/mantisrx/common/compression/CompressionUtilsTest.java
```diff
@@ -67,6 +67,61 @@
...
+            assertEquals(event1, result.get(0).getEventAsString());
+            assertEquals(event2, result.get(1).getEventAsString());
+            assertEquals(event3, result.get(2).getEventAsString());
+
+        } catch (IOException ex) {
+            Assert.fail("Tokenization threw an IO exception that was unexpected");
+        }
+    }
+
+    @Test public void testMultiline() {
```
hmitnflx, line 95: nit: the annotation and method definition are on the same line. is it a new formatting rule?

mantis-common/src/main/java/io/mantisrx/common/compression/CompressionUtils.java
```diff
@@ -217,10 +217,19 @@ private static int getTotalByteSize(List<List<byte[]>> nestedEvents, byte[] deli
             for (int i = 0; i < line.length(); i++) {
```
hmitnflx, line 217: +1 to this comment. perhaps there is a way to achieve this using regex or something.

hmitnflx (approved):

hmitnflx:
Thanks for catching and identifying this, really great!

mantis-common/src/test/java/io/mantisrx/common/compression/CompressionUtilsTest.java
```diff
@@ -67,6 +67,61 @@
...
+                + event3;
+        try (BufferedReader reader = new BufferedReader(new StringReader(testInput))) {
+            List<MantisServerSentEvent> result = CompressionUtils.tokenize(reader, delimiter);
+
+            assertEquals("Delimiter: '" + delimiter + "'", result.size(), 3);
+            assertEquals(event1, result.get(0).getEventAsString());
+            assertEquals(event2, result.get(1).getEventAsString());
+            assertEquals(event3, result.get(2).getEventAsString());
+
+        } catch (IOException ex) {
```
liuml07, line 93: I don't think it's necessary to catch this exception here. Just make the test method throw exception, and the test will simply fail in case of IOE. Same in the other test.

mantis-common/src/test/java/io/mantisrx/common/compression/CompressionUtilsTest.java
```diff
@@ -67,6 +67,61 @@
...
+                + event2
+                + delimiter
+                + event3;
+        try (BufferedReader reader = new BufferedReader(new StringReader(testInput))) {
+            List<MantisServerSentEvent> result = CompressionUtils.tokenize(reader, delimiter);
+
+            assertEquals("Delimiter: '" + delimiter + "'", result.size(), 3);
+            assertEquals(event1, result.get(0).getEventAsString());
+            assertEquals(event2, result.get(1).getEventAsString());
+            assertEquals(event3, result.get(2).getEventAsString());
```
liuml07, line 91:
```suggestion
            List<String> actual = result.stream().map(e -> e.getEventAsString()).collect(Collectors.toList());
            assertEquals("Delimiter: '" + delimiter + "'", Arrays.asList(event1,event2,event3), actual);
```
Less duplicate code. Also if the test fails, we know the `result` value.
```
Delimiter: 'ccd' expected:<[abc, def, ghi]> but was:<[abcccddef, ghi]>
Expected :[abc, def, ghi]
Actual   :[abcccddef, ghi]
```

mantis-common/src/main/java/io/mantisrx/common/compression/CompressionUtils.java
```diff
@@ -217,10 +217,19 @@ private static int getTotalByteSize(List<List<byte[]>> nestedEvents, byte[] deli
...
                 if (line.charAt(i) != delimiterArray[delimiterCount]) {
                     if (delimiterCount > 0) {
-                        for (int j = 0; j < delimiterCount; ++j) {
-                            sb.append(delimiterArray[j]);
+                        boolean prefixMatch = true;
```
liuml07, line 220: This whole method is too fancy to me. I may fail your interview if the question is to fix this bug.

Could anyone tell me why this one liner is not sufficient to replace the whole method?
```
return Arrays.stream(bf.lines().collect(Collectors.joining("")).split(Pattern.quote(delimiter))).map(e -> new MantisServerSentEvent(e)).collect(Collectors.toList());
```

(of course please make that multiple lines 🤣 )

It passes all your tests (new and old).

calvin681: My guess is memory. I think this was meant to check lines as they are read in BufferedReader, so never need to pull the entire content into a String.

codyrioux: I believe Calvin is correct -- performance and memory was the reason for this. Most events are pretty small but some people are sending monstrous (many megabyte +) events through Mantis. In the PlayAPI req/resp body logging use case batches can reach several gigabytes every few milliseconds. The large strings and the G1 collector don't get along very well.

codyrioux (approved):

mantis-common/src/main/java/io/mantisrx/common/compression/CompressionUtils.java
liuml07: I was not aware of the large input, but that's a fair reason. I do not know if we can refactor the code in a way that is much simpler based on something like above, but we can tackle that separately I believe.

If we want to optimize memory/cpu for this, there are also some other ideas.
1. If the output strings could be large, we can let the `StringBuilder` have some initial capacity instead of letting it grow from beginning, e.g. `sb = new StringBuilder(4096);`.
2. If the output String can be too small, we can reuse the StringBuilder and just `setLength(0)` instead of `new StringBuilder()`.

codyrioux:
On the note of the last commit to fix spotless. I've actually started adding a git `pre-push` hook in repos I work in:

```
#!/usr/bin/env bash
gw :spotlessApply
gw clean build
```

Makes sure that I don't push code that doesn't pass / won't build.

liuml07:
I think I finally understand the patch. It's merged so it's great.

FWIW, in the future if we want to revisit this, the reason of not processing the whole input string all at once (e.g. the [one-liner](https://github.com/Netflix/mantis/pull/154#discussion_r835651926)) is the memory overhead.

One optimization on top of the oneliner is buffered read. Sample code (passing all tests but not really tested). As the current approach is already well known in the team and I have limited knowledge about the payload, I'm not filing a PR this time.

```java
    static List<MantisServerSentEvent> tokenize(BufferedReader bf, String delimiter) throws IOException {
        final List<MantisServerSentEvent> ret = new ArrayList<>();
        final StringBuilder sb = new StringBuilder();
        final char[] buf = new char[4096];
        while(bf.ready()) {
            int count = bf.read(buf);
            if (count == -1) break;
            sb.append(buf, 0, count);
            int idx = sb.lastIndexOf(delimiter);
            if (idx == -1) continue;
            ret.addAll(tokenize(sb.substring(0, idx), delimiter));
            sb.delete(0, idx);
        }
        ret.addAll(tokenize(sb.toString(), delimiter));
        return ret;
    }

    // one liner
    static List<MantisServerSentEvent> tokenize(String str, String delimiter) throws IOException {
        return Arrays.stream(str.replaceAll("\\n","").split(Pattern.quote(delimiter))).filter(s -> !s.isEmpty()).map(e -> new MantisServerSentEvent(e)).collect(Collectors.toList());
    }
```

PR #155, opened 2022-03-28
[Part 6]: Blob Store abstractions and a classloader factory that uses the blob store

sundargates:
Two features are added through this change.
1). Firstly, I'm introducing a new blob store abstraction and a few implementations that can help download the blob from various stores such as s3, artifactory, etc.
2). A new class loader implementation has also been added to use the blob store to download the requested artifacts from s3 or artifactory or elsewhere and create a class loader object that can run the user's application code.

### Context

In the new Titus world, one can use the cluster to run any user-submitted mantis job. Thus, we need capabilities to both download artifacts on demand and load them into the JVM to run the user's mantis job. This diff adds some of the abilities required for the above functionality. 

### Checklist

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# http://editorconfig.org
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
--- a/build.gradle
+++ b/build.gradle
@@ -44,6 +44,7 @@ ext.versions = [
 
 ext.libraries = [
         asyncHttpClient: "org.asynchttpclient:async-http-client:2.12.3",
+        commonsIo      : "commons-io:commons-io:2.11.0",
         commonsLang3   : 'org.apache.commons:commons-lang3:3.5',
         flinkRpcApi    : [
                 "org.apache.flink:flink-rpc-core:${versions.flink}",
@@ -56,6 +57,8 @@ ext.libraries = [
                 "junit:junit:${versions.junit4}",
                 "junit:junit-dep:${versions.junit4}",
         ],
+        hadoopCommon   : "org.apache.hadoop:hadoop-common:2.7.7",
+        hadoopS3       : "org.apache.hadoop:hadoop-aws:2.7.7",
         junitJupiter   : [
                 "org.junit.jupiter:junit-jupiter-api:${versions.junit5}",
                 "org.junit.jupiter:junit-jupiter-engine:${versions.junit5}",
@@ -72,6 +75,7 @@ ext.libraries = [
         vavr           : "io.vavr:vavr:${versions.vavr}",
         vavrJackson    : "io.vavr:vavr-jackson:${versions.vavr}",
         vavrTest       : "io.vavr:vavr-test:${versions.vavr}",
+        zip4j          : "net.lingala.zip4j:zip4j:2.9.0",
 ]
 
 allprojects {
--- a/build.gradle
+++ b/build.gradle
@@ -35,6 +35,7 @@ buildscript {
 
 ext.versions = [
         flink  : "1.14.2",
+        hadoop : "2.7.7",
         junit4 : "4.11",
         junit5 : "5.4.+",
         mockito: "2.0.+",
@@ -57,8 +58,8 @@ ext.libraries = [
                 "junit:junit:${versions.junit4}",
                 "junit:junit-dep:${versions.junit4}",
         ],
-        hadoopCommon   : "org.apache.hadoop:hadoop-common:2.7.7",
-        hadoopS3       : "org.apache.hadoop:hadoop-aws:2.7.7",
+        hadoopCommon   : "org.apache.hadoop:hadoop-common:${versions.hadoop}",
+        hadoopS3       : "org.apache.hadoop:hadoop-aws:${versions.hadoop}",
         junitJupiter   : [
                 "org.junit.jupiter:junit-jupiter-api:${versions.junit5}",
                 "org.junit.jupiter:junit-jupiter-engine:${versions.junit5}",
@@ -124,7 +125,7 @@ subprojects {
         mavenCentral()
         /**
          * This repository locates artifacts that don't exist in maven central but we had to backup from jcenter
-         * The exclusiveContent 
+         * The exclusiveContent
          */
         exclusiveContent {
             forRepository {
--- a/mantis-common/src/main/java/com/mantisrx/common/utils/Closeables.java
+++ b/mantis-common/src/main/java/com/mantisrx/common/utils/Closeables.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2022 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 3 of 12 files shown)


build.gradle
```diff
@@ -56,6 +57,8 @@ ext.libraries = [
                 "junit:junit:${versions.junit4}",
                 "junit:junit-dep:${versions.junit4}",
         ],
+        hadoopCommon   : "org.apache.hadoop:hadoop-common:2.7.7",
```
hmitnflx, line 60: nit: refactor version to a const in `versions`.

mantis-common/src/main/java/com/mantisrx/common/utils/Closeables.java
```diff
@@ -0,0 +1,53 @@
...
+     * @return a composite closeable
+     */
+    public static Closeable combine(Collection<? extends Closeable> closeables) {
+        return combine(closeables.toArray(new Closeable[0]));
+    }
+
+    public static Closeable combine(Closeable... closeables) {
+        return () -> {
+            List<Throwable> list = new ArrayList<>();
+            for (Closeable closeable: closeables) {
```
hmitnflx, line 40: do we care about ordering here? perhaps, close in the reverse order of the original list...

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/BlobStore.java
```diff
@@ -0,0 +1,110 @@
...
+        @Override
+        public void close() throws IOException {
+            blobStore.close();
+        }
+
+        private String getDestDir(ZipFile zipFile) {
+            return zipFile.getFile().getPath() + "-unzipped";
+        }
+
+        private ZipFile isZipFile(File file) {
```
hmitnflx, line 99: make this return a boolean perhaps or change name to `getZipFile`

.editorconfig
```diff
@@ -0,0 +1,13 @@
+# EditorConfig helps developers define and maintain consistent
```
codyrioux, line 1: Great call. I just installed the plugin for my editor.

mantis-server/mantis-server-worker/src/test/java/io/mantisrx/server/worker/TestBlobStore.java
```diff
@@ -0,0 +1,49 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import io.mantisrx.server.worker.BlobStore.PrefixedBlobStore;
+import java.io.File;
+import java.net.URI;
+import org.junit.Test;
+import org.mockito.Matchers;
+
+public class TestBlobStore {
```
hmitnflx, line 31: nit: `BlobStoreTest`

codyrioux (approved):

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/BlobStore.java
```diff
@@ -0,0 +1,110 @@
...
+        @Override
+        public File get(URI blobUrl) throws IOException {
+            final File localFile = blobStore.get(blobUrl);
+            final ZipFile zipFile = isZipFile(localFile);
+            if (zipFile == null) {
+                return localFile;
+            } else {
+                String destDirStr = getDestDir(zipFile);
+                File destDir = new File(destDirStr);
+                if (!destDir.exists()) {
```
hmitnflx, line 83: I feel like if the file path exists, we should delete it entirely and unzip `zipFile` there ofc logging a warning along with it.
Also, delete or close the `localFile` in this scenario.

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/BlobStore.java
```diff
@@ -0,0 +1,110 @@
...
+                    zipFile.extractAll(destDirStr);
+                }
+                return destDir;
+            }
+        }
+
+        @Override
+        public void close() throws IOException {
+            blobStore.close();
```
hmitnflx, line 92: Should we close it or let our caller close `blobStore`?

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/BlobStore.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        private String getDestDir(ZipFile zipFile) {
```
hmitnflx, line 95: this should be called `getUnzippedFile` or something. Initially I got confused thinking this points to the destination directory containing the zipfile and we'll need to append something to determine the name of the unzipped file.

The name (with `-unzipped` suffix) would look weird it `zipFile` has extensions in the name but I guess that's okay for the time being.

mantis-common/src/main/java/com/mantisrx/common/utils/Closeables.java
sundargates: We don't care about ordering here because the assumption here is that these are independent closeables.

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/BlobStore.java
sundargates: In this case, the blob store object represents the delegate that we do not want to expose to the caller to. So, we need to perform this action.

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/BlobStore.java
sundargates: done

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/BlobStore.java
sundargates: I'm now throwing an exception in case of the destDir not being empty, which is the expectation for this method.

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/BlobStore.java
sundargates: done

mantis-server/mantis-server-worker/src/test/java/io/mantisrx/server/worker/TestBlobStore.java
sundargates: done

build.gradle
sundargates: done

mantis-server/mantis-server-worker/src/test/java/io/mantisrx/server/worker/TestBlobStore.java
calvin681: +1

sundargates:
Please feel free to add any comments. I'll address them in later diffs. Closing for now.

calvin681 (approved):

mantis-common/src/main/java/com/mantisrx/common/utils/Closeables.java
```diff
@@ -0,0 +1,53 @@
...
+                try {
+                    closeable.close();
+                } catch (IOException e) {
```
Andyz26, line 43: Just wondering if we encouter runtime exception what should be the best behavior (e.g. closeable1 failed with runtime exception, shall closeable2 still get closed)?

PR #159, opened 2022-04-02
[Part 9]: Introducing Sink::close

sundargates:
An explicit close is required because in the new model (Titus), we want the ability to stop tasks gracefully. This means shutting down all the resources that the job is currently consuming. In the old world (Apache Mesos), this was not required because we were forcefully shutting down processes which are both not clean as well as not feasible in the new world.

### Context

We are introducing a breaking API change because we want to ensure developers don't forget to close their sinks correctly.

### Checklist


mantis-runtime/src/main/java/io/mantisrx/runtime/sink/ServerSentEventsSink.java
```diff
@@ -141,57 +145,81 @@ public void call(Context context, PortRequest portRequest, final Observable<T> o
...
+            httpServer.start();
         }
         portObservable.onNext(port);
     }
 
+    @Override
+    public void close() throws IOException {
+        if (pushServerSse != null) {
+            pushServerSse.shutdown();
+        } else if (httpServer != null) {
```
Andyz26, line 193: nit: in this case IMO null check is enough to shutdown pushServerSse and httpServer, which is slightly safer if for any reason both of them got created.

codyrioux:
It should be an error state if both somehow got created.

mantis-connectors/mantis-connector-publish/src/main/java/io/mantisrx/connector/publish/source/http/SourceSink.java
```diff
@@ -46,25 +45,30 @@ public Void call(Map<String, List<String>> t1, Context t2) {
...
         sink.call(context, portRequest, observable);
     }
+
+    @Override
+    public void close() throws IOException {
+        try {
+            sink.close();
+        } finally {
+            subscription.unsubscribe();
+        }
```
sundargates, line 72: @calvin681 should we unsubscribe before doing the close?

mantis-examples/mantis-examples-synthetic-sourcejob/src/main/java/io/mantisrx/sourcejob/synthetic/sink/TaggedDataSourceSink.java
```diff
@@ -77,8 +70,24 @@ public void call(Context context, PortRequest portRequest,
...
         sink.call(context, portRequest, observable);
     }
+
+    @Override
+    public void close() throws IOException {
+        try {
+            sink.close();
+        } finally {
+            subscription.unsubscribe();
+        }
```
sundargates, line 91: same as above.

mantis-runtime/src/main/java/io/mantisrx/runtime/sink/Sinks.java
```diff
@@ -70,6 +86,15 @@ public Metadata metadata() {
...
                 sink.init(t);
             }
+
+            @Override
+            public void close() throws IOException {
+                try {
+                    sink.close();
+                } finally {
+                    subscription.unsubscribe();
+                }
```
sundargates, line 96: should the unsubscribe happen before sink::close?

mantis-runtime/src/main/java/io/mantisrx/runtime/sink/Sinks.java
```diff
@@ -96,18 +121,38 @@ public void init(Context t) {
...
             }
+
+            @Override
+            public void close() throws IOException {
+                IOException last = null;
+                for (Sink<T> sink: many) {
+                    try {
+                        sink.close();
+                    } catch (IOException e) {
+                        if (last == null) {
```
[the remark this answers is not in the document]
sundargates: good point. let me clean this up.

mantis-source-jobs/mantis-source-job-kafka/src/main/java/io/mantisrx/sourcejob/kafka/sink/TaggedDataSourceSink.java
```diff
@@ -72,8 +64,24 @@ public void call(Context context, PortRequest portRequest,
...
         sink.call(context, portRequest, observable);
     }
+
+    @Override
+    public void close() throws IOException {
+        try {
+            sink.close();
+        } finally {
+            subscription.unsubscribe();
+        }
```
sundargates, line 85: Same as before.

mantis-runtime/src/main/java/io/mantisrx/runtime/sink/Sinks.java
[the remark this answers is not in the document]
sundargates: fixed this.

sundargates:
let me merge the change now. Any pending comments will be addressed in subsequent diffs.

mantis-connectors/mantis-connector-publish/src/main/java/io/mantisrx/connector/publish/source/http/SourceSink.java
calvin681: TBH, I'm not even sure why we need to eager subscribe to the observable in `call`. Regardless, unsubscribing first would make more logical sense.

calvin681 (approved):

PR #162, opened 2022-04-05
Introducing Source::close

sundargates:
### Context

Every source needs to close any resources such as HTTP clients or thread pools associated with creating the Observable as part of the Source::call API. This diff adds a new method to the source API, which will eventually be called as part of stage execution. 

### Checklist

### DIFF
--- a/mantis-connectors/mantis-connector-job/src/main/java/io/mantisrx/connector/job/source/JobSource.java
+++ b/mantis-connectors/mantis-connector-job/src/main/java/io/mantisrx/connector/job/source/JobSource.java
@@ -19,6 +19,7 @@ package io.mantisrx.connector.job.source;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
+import com.mantisrx.common.utils.Closeables;
 import com.mantisrx.common.utils.MantisSSEConstants;
 import io.mantisrx.client.MantisSSEJob;
 import io.mantisrx.common.MantisServerSentEvent;
@@ -37,22 +38,24 @@ import io.mantisrx.runtime.source.Source;
 import io.mantisrx.shaded.com.google.common.collect.Lists;
 import io.vavr.Tuple;
 import io.vavr.Tuple2;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import rx.Observable;
 
-
+@Slf4j
 public class JobSource extends AbstractSourceJobSource implements Source<MantisServerSentEvent> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(JobSource.class);
     private static JsonParser parser = new JsonParser();
     protected List<TargetInfo> targets;
-
+    private final List<MantisSSEJob> jobs = new ArrayList<>();
 
     public JobSource(List<TargetInfo> targets) {
         this.targets = targets;
@@ -110,6 +113,7 @@ public class JobSource extends AbstractSourceJobSource implements Source<MantisS
             boolean enableCompressedBinary = targetInfo.enableCompressedBinary;
 
             job = getSourceJob(sourceJobName, criterion, clientId, samplePerSec, enableMetaMessages, enableCompressedBinary, obs, Optional.<SinkParameters>empty());
+            jobs.add(job);
 
             if (sourceObs == null) {
                 sourceObs = job.connectAndGet();
@@ -127,6 +131,15 @@ public class JobSource extends AbstractSourceJobSource implements Source<MantisS
             }
         }
         return sourceObs;
+    }
+
+    @Override
+    public void close() throws IOException {
+        try {
+            Closeables.combine(jobs).close();
+        } finally {
+            jobs.clear();
+        }
     }
 
     /**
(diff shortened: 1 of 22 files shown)


mantis-connectors/mantis-connector-job/src/main/java/io/mantisrx/connector/job/source/JobSource.java
```diff
@@ -129,6 +133,15 @@ public JobSource(String targetInfoStr) {
         return sourceObs;
     }
 
+    @Override
+    public void close() throws IOException {
+        try {
+            Closeables.combine(jobs).close();
```
hmitnflx, line 139: if this is how we intend to use it, I think it might be better to write this as a utility function. 
Alternatively, we just make `jobs` a `Closeables` and keep on adding to it instead of the array.

mantis-runtime/src/main/java/io/mantisrx/runtime/source/http/impl/ClientWithResponse.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package io.mantisrx.runtime.source.http.impl;
+
+import lombok.Value;
+import mantis.io.reactivex.netty.protocol.http.client.HttpClient;
+import mantis.io.reactivex.netty.protocol.http.client.HttpClientResponse;
+
+@Value
+class ClientWithResponse<R, E> {
```
hmitnflx, line 24: This is an interesting class to exist. Could you please add javadoc for why it's needed OR how it's used?

hmitnflx (approved):

mantis-connectors/mantis-connector-job/src/main/java/io/mantisrx/connector/job/source/JobSource.java
sundargates: Isn't Closeables::combine (https://github.com/Netflix/mantis/blob/master/mantis-common/src/main/java/com/mantisrx/common/utils/Closeables.java) a good enough utility function? What more do you want to move into a function?

Regarding the second comment, I prefer to keep it stored in the original DS because if you capture it as a closeable, you lose all other capabilities.

mantis-runtime/src/main/java/io/mantisrx/runtime/source/http/impl/ClientWithResponse.java
sundargates: I have moved this within HttpSourceImpl as an inner class because it's only used from within it. I think the JavaDoc is unnecessary, given that this is strictly an implementation detail.

mantis-runtime/src/main/java/io/mantisrx/runtime/source/http/impl/HttpSourceImpl.java
```diff
@@ -708,7 +693,11 @@ public void serverRemoved(ServerInfo server) {
         }
 
         public void reset() {
-            connectedServers.clear();
+            Set<ServerInfo> connectedServerInfos = connectedServers.keySet();
```
calvin681, line 696: LGTM, only concern is that iterating over keyset may throw exception if map is changed. i'm not familiar enough with concurrent hashmap to know if that can happen or not. you can also just use getConnectedServers()

sundargates: @calvin681 I can look into the behavior of keySet() under changes to the underlying concurrent map but do you think removeConnectedServer where we close the underlying HttpClient makes sense?

mantis-runtime/src/main/java/io/mantisrx/runtime/source/Source.java
```diff
@@ -18,13 +18,14 @@
...
 import io.mantisrx.runtime.parameter.ParameterDefinition;
+import java.io.Closeable;
 import java.util.Collections;
 import java.util.List;
 import rx.Observable;
 import rx.functions.Func2;
 
 
-public interface Source<T> extends Func2<Context, Index, Observable<Observable<T>>> {
+public interface Source<T> extends Func2<Context, Index, Observable<Observable<T>>>, Closeable {
```
nickmahilani, line 28: we could avoid all the empty `close()` calls in connectors and user code by having a default empty `close()` impl in the `Source` interface

sundargates: Hmm, the decision to avoid the default close() implementation in the source API was deliberate. We want to ensure that users who implement these APIs think about what resources they need to close actively.

nickmahilani: For a new interface, I agree. But for an existing interface with existing implementations which has been around for a while, it is a tradeoff between a user's build breaking to add an empty `close()` v/s keeping that as the default behavior with an option to override.

sundargates: I don't think this should break a user's build because we are doing a major version upgrade (2.0.0) with the new Titus set of changes. nfmantis modules that pull in mantis artifacts will also undergo a major upgrade. This will ensure only developers that opt in for the new version will receive it. We will also have a bom release reflecting this.

nickmahilani: this makes sense if it is part of 2.0.0

mantis-runtime/src/main/java/io/mantisrx/runtime/source/http/impl/HttpSourceImpl.java
Andyz26: According to the keyset() doc the iterator is 'weakly consistent' and shall never emit modification exception.
In the meantime, I wonder if there is anything blocking/timeout inside the httpclient shutdown call that in extreme cases can cause this operation to hang/stuck for a long time?

Andyz26: If so maybe this can be done as fire-and-forget with timeout in a different thread?

codyrioux (approved):

PR #170, opened 2022-04-13
Iceberg writer stage should include WRITER_MAXIMUM_POOL_SIZE in its parameters

liuml07:
### Context

Iceberg writer stage missed to add `WRITER_MAXIMUM_POOL_SIZE` into the parameter list. So all overrides in jobs will be ignored and the default value will be used regardless.

### Checklist

- [ ] ~Added copyright headers for new files from `CONTRIBUTING.md`~

### DIFF
--- a/mantis-connectors/mantis-connector-iceberg/src/main/java/io/mantisrx/connector/iceberg/sink/writer/IcebergWriterStage.java
+++ b/mantis-connectors/mantis-connector-iceberg/src/main/java/io/mantisrx/connector/iceberg/sink/writer/IcebergWriterStage.java
@@ -102,7 +102,12 @@ public class IcebergWriterStage implements ScalarComputation<Record, DataFile> {
                         .description(WriterProperties.WRITER_FILE_FORMAT_DESCRIPTION)
                         .validator(Validators.alwaysPass())
                         .defaultValue(WriterProperties.WRITER_FILE_FORMAT_DEFAULT)
-                        .build()
+                        .build(),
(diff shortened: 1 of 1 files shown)


calvin681 (approved):

calvin681:
You may want to back port this to the `release-1.3` branch.

sundargates:
> 

@liuml07 It looks like we should be using Parameters::get without the default value in this case. That would surface the runtime exception and hence the problem in this case. 
```
public Object get(String key) {
        // check if required parameter, make sure
        // has a value
        if (requiredParameters.contains(key) &&
                !state.containsKey(key)) {
            throw new ParameterException("Attempting to reference a required parameter witn no value: " + key + ", check parameter definitions for job.");
        }
        // check if parameter definition exists
        if (!parameterDefinitions.contains(key)) {
            throw new ParameterException("Attempting to reference parameter: " + key + ", with no definition, check parameter definitions in job.");
        }
        return state.get(key);
    }

    public Object get(String key, Object defaultValue) {
        Object value = defaultValue;
        try {
            value = get(key);
            if (value == null) {
                value = defaultValue;
            }
        } catch (ParameterException ex) {
            value = defaultValue;
        }
        return value;
    }
```

codyrioux (approved):
I agree with Sundaram in that we should not be using a default at the call site where we fetch the parameter.

liuml07:
Makes sense. I'll file a follow-up PR to change all callsites in Iceberg sink module.

PR #171, opened 2022-04-13
Iceberg sink uses Parameters::get to ensure parameter definition exists

liuml07:
### Context
https://github.com/Netflix/mantis/pull/170#issuecomment-1098264092

> It looks like we should be using Parameters::get without the default value in this case. That would surface the runtime exception and hence the problem in this case.

### Checklist

- [ ] ~Added copyright headers for new files from `CONTRIBUTING.md`~

### DIFF
--- a/mantis-runtime/build.gradle
+++ b/mantis-runtime/build.gradle
@@ -25,7 +25,6 @@ dependencies {
     api project(':mantis-common')
     api libraries.slf4jApi
     testImplementation libraries.junit4
-    testImplementation libraries.junitJupiter
     testImplementation libraries.mockitoCore
 
 }
--- a/mantis-runtime/build.gradle
+++ b/mantis-runtime/build.gradle
@@ -25,6 +25,7 @@ dependencies {
     api project(':mantis-common')
     api libraries.slf4jApi
     testImplementation libraries.junit4
+    testImplementation libraries.junitJupiter
     testImplementation libraries.mockitoCore
 
 }
--- a/mantis-runtime/src/main/java/io/mantisrx/runtime/parameter/Parameters.java
+++ b/mantis-runtime/src/main/java/io/mantisrx/runtime/parameter/Parameters.java
@@ -21,7 +21,6 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
-
 public class Parameters {
 
     private Set<String> requiredParameters = new HashSet<>();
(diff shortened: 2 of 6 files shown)


mantis-runtime/src/main/java/io/mantisrx/runtime/parameter/Parameters.java
```diff
@@ -38,22 +37,34 @@ public Parameters(Map<String, Object> state,
...
         return state.get(key);
     }
 
+    /**
+     * Get parameter value given key without validation.
+     *
+     * If the key is not defined or missing a provided value, the given default value will be returned.
+     */
     public Object get(String key, Object defaultValue) {
-        Object value = defaultValue;
```
codyrioux, line 56: Do the semantics actually change with this? My initial read of the code says no, but if it does it may unintentionally impact other jobs.

mantis-runtime/src/main/java/io/mantisrx/runtime/parameter/Parameters.java
```diff
@@ -38,22 +37,34 @@ public Parameters(Map<String, Object> state,
...
         }
         return state.get(key);
     }
 
+    /**
+     * Get parameter value given key without validation.
+     *
+     * If the key is not defined or missing a provided value, the given default value will be returned.
+     */
     public Object get(String key, Object defaultValue) {
```
sundargates, line 66: What if you have a default value of null for the parameter and pass in a non-default value here as part of defaultValue? What should be the expected behavior? Can we also write tests for this to ensure the behavior remains consistent as we advance?

mantis-runtime/src/main/java/io/mantisrx/runtime/parameter/Parameters.java
liuml07: No, all branches will assign value so it's redundant to have an initial value.

mantis-runtime/src/main/java/io/mantisrx/runtime/parameter/Parameters.java
liuml07: That's a good question, but I really do not have the answer. The expected behavior depends on how it's now being used.

mantis-runtime/src/main/java/io/mantisrx/runtime/parameter/Parameters.java
liuml07: But it's cosmetic and I can revert that. Or better I can rewrite the whole function to use multiple returns.
```
        try {
            final Object value = get(key);
            return value != null ? value : defaultValue;
        } catch (ParameterException ex) {
            return defaultValue;
        }
```

mantis-runtime/src/main/java/io/mantisrx/runtime/parameter/Parameters.java
liuml07: According to this code snippet: https://github.com/Netflix/mantis/blob/master/mantis-runtime/src/main/java/io/mantisrx/runtime/parameter/ParameterUtils.java#L309-L315

It seems "default value of null for the parameter" means the parameter does not have a default value, i.e. missing in the definition. In that case, this `get()` method is doing the right thing: return the provided `defaultValue`. We can add a test for this.

liuml07:
Actually some existing tests fail due to the test parameter "state" is not complete - not including all parameter definitions. So we will need to update the test code a bit. I'll make this ready for review after I fix that failing test in a non-intrusive way.

liuml07:
failing test seems unrelated.

liuml07:
All green. I'll merge by EoD. CC: @sundargates

mantis-runtime/build.gradle
```diff
@@ -25,6 +25,7 @@ dependencies {
     api project(':mantis-common')
     api libraries.slf4jApi
     testImplementation libraries.junit4
+    testImplementation libraries.junitJupiter
```
sundargates, line 28: isn't junit-jupiter meant for usage with junit5?

sundargates (approved):

mantis-runtime/build.gradle
liuml07: Yes. It works just fine if we only use the utility assertion methods from Jupiter. But I agree, it's better not to mix JUnit 4 and Jupiter. I have written the `assertThrows` in the testing code instead of pulling junitJupiter. I'll file a follow-up PR to upgrade JUnit 4 to Jupiter in`mantis-runtime` module.

liuml07:
`testJobComplete (io.mantisrx.master.jobcluster.JobClusterTest) failed` but it seems unrelated. Let me merge from `master` and see if it's still failing.

PR #175, opened 2022-04-14
[ResourceCluster Part1] ResourceClustersManager actor + contracts

Andyz26:
### Context

This PR adds the new ResourceClustersManagerActor and its dependent classes and contract classes as part of the effort to support non-Mesos cluster type for Mantis.

This new actor will be handling both API requests and requests from other actors related to any resource-cluster-related operations.

To support different cluster types (e.g. Kubernetes/Spinnaker), this new ResourceClustersManager actor and its corresponding contracts will provide abstraction only and not be tied to any framework-specific dependencies.

image

### Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,13 +34,14 @@ buildscript {
 }
 
 ext.versions = [
-        flink  : "1.14.2",
-        hadoop : "2.7.7",
-        junit4 : "4.11",
-        junit5 : "5.4.+",
-        mockito: "2.0.+",
-        slf4j  : "1.7.0",
-        vavr   : "0.9.2",
+        flink   : "1.14.2",
+        hadoop  : "2.7.7",
+        junit4  : "4.11",
+        junit5  : "5.4.+",
+        mockito : "2.0.+",
+        mockito3: "3.+",
+        slf4j   : "1.7.0",
+        vavr    : "0.9.2",
 ]
 
 ext.libraries = [
@@ -68,6 +69,7 @@ ext.libraries = [
         mantisShaded   : "io.mantisrx:mantis-shaded:1.3.+",
         mockitoAll     : "org.mockito:mockito-all:${versions.mockito}",
         mockitoCore    : "org.mockito:mockito-core:${versions.mockito}",
+        mockitoCore3   : "org.mockito:mockito-core:${versions.mockito3}",
         rxJava         : "io.reactivex:rxjava:1.3.8",
         rxNettyShaded  : "com.netflix:mantis-rxnetty:0.4.19.1",
         slf4jApi       : "org.slf4j:slf4j-api:${versions.slf4j}",
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/ListResourceClusterRequest.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/ListResourceClusterRequest.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2022 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.control.plane.resource.cluster.proto;
+
+import lombok.Builder;
+import lombok.Value;
+
+@Builder
+@Value
+public class ListResourceClusterRequest {
+}
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/MantisResourceClusterSpec.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/MantisResourceClusterSpec.java
@@ -1,0 +1,150 @@
+/*
+ * Copyright 2022 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.control.plane.resource.cluster.proto;
+
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonCreator;
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Map;
+import java.util.Set;
+import lombok.Builder;
+import lombok.EqualsAndHashCode;
+import lombok.NonNull;
+import lombok.Singular;
+import lombok.Value;
+
+/**
+ * Contract class to define a Mantis resource cluster.
+ */
+@Value
+@Builder
+public class MantisResourceClusterSpec {
+
+    @NonNull
+    String name;
+
+    /**
+     * ID fields maps to cluster name or spinnaker app name.
+     */
+    @NonNull
+    String id;
+
+    @NonNull
+    String ownerName;
+
+    @NonNull
+    String ownerEmail;
+
+    @NonNull
+    MantisResourceClusterEnvType envType;
+
+    @Singular
+    Set<SkuTypeSpec> skuSpecs;
+
+    @Singular
+    Map<String, String> clusterMetadataFields;
+
+    @JsonCreator
+    public MantisResourceClusterSpec(
+            @JsonProperty("name") final String name,
+            @JsonProperty("id") final String id,
+            @JsonProperty("ownerName") final String ownerName,
+            @JsonProperty("ownerEmail") final String ownerEmail,
+            @JsonProperty("envType") final MantisResourceClusterEnvType envType,
+            @JsonProperty("skuSpecs") final Set<SkuTypeSpec> skuSpecs,
+            @JsonProperty("clusterMetadataFields") final Map<String, String> clusterMetadataFields) {
+        this.name = name;
+        this.id = id;
+        this.ownerName = ownerName;
+        this.ownerEmail = ownerEmail;
+        this.envType = envType;
+        this.skuSpecs = skuSpecs;
+        this.clusterMetadataFields = clusterMetadataFields;
+    }
+
+    @Builder
+    @Value
+    @EqualsAndHashCode(onlyExplicitlyIncluded = true)
+    public static class SkuTypeSpec {
+        @NonNull
+        @EqualsAndHashCode.Include
+        String skuId;
+
+        @NonNull
+        SkuCapacity capacity;
+
+        @NonNull
+        String imageId;
+
+        int cpuCoreCount;
+
+        int memorySizeInBytes;
+
+        int networkMbps;
+
+        int diskSizeInBytes;
+
+        @Singular
+        Map<String, String> skuMetadataFields;
+
+        @JsonCreator
+        public SkuTypeSpec(
+                @JsonProperty("skuId") final String skuId,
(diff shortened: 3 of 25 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/ListResourceClusterRequest.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package io.mantisrx.control.plane.resource.cluster.proto;
+
+import lombok.Builder;
+import lombok.Value;
+
+@Builder
+@Value
+public class ListResourceClusterRequest {
```
sundargates, line 24: move these within the actor class.

Andyz26: All these contracts in a single file are going to make it a huge one though? Also, there are composite referenced classes that are already huge by themselves.

Andyz26:
@calvin681  @codyrioux @hmit @sundargates: I would love to get some review/feedback, especially regarding the interface of IResourceClusterProvider and the provisioning part in the ResourceClustersManagerActor.

Andyz26:
@sundargates I've renamed the classes here to avoid naming conflicts. Let me know if you have any other concerns.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/resourceprovider/SimpleFileResourceClusterStorageProvider.java
```diff
@@ -0,0 +1,170 @@
...
+                        .map(mapper::writeValueAsString)
+                        .map(ByteString::fromString)
+                        .runWith(listFileSink, system)
+                        .exceptionally(e -> { throw new RuntimeException("failed to save cluster list. ", e); })
+                        .thenApply(ioRes -> true); // IOResult result is always true.
+    }
+
+    @Override
+    public CompletionStage<RegisteredResourceClustersWritable> getRegisteredResourceClustersWritable() {
+        final ObjectMapper mapper = new ObjectMapper();
```
hmitnflx, line 124: I don't believe you need a separate instance of `ObjectMapper` here since `readerFor` and `ObjectReader`(s) look thread safe anyway.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/MantisResourceClusterSpec.java
```diff
@@ -0,0 +1,153 @@
...
+        String skuId;
+
+        int minSize;
+
+        int maxSize;
+
+        int desireSize;
+
+        @JsonCreator
+        public SkuCapacity(
```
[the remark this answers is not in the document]
Andyz26: This is a workaround (when I tested the original clean version (@value + builder only) in nfmantis the shaded classes and Jackson doesn't play well with each other and other internal SDKs and it can only work when I have both annotation + upgrade shaded Jackson for all combinations).

build.gradle
```diff
@@ -68,6 +69,7 @@ ext.libraries = [
         mantisShaded   : "io.mantisrx:mantis-shaded:1.3.+",
         mockitoAll     : "org.mockito:mockito-all:${versions.mockito}",
         mockitoCore    : "org.mockito:mockito-core:${versions.mockito}",
+        mockitoCore3   : "org.mockito:mockito-core:${versions.mockito3}",
```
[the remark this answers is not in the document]
Andyz26: I tried it briefly and there are quite some breaks in the current code when switching to mockito3. This could be a backlog work item. Tracked at #184

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/MantisResourceClusterSpec.java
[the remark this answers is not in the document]
Andyz26: Let me add some notes here about this.

build.gradle
```diff
@@ -34,13 +34,14 @@ buildscript {
...
-        slf4j  : "1.7.0",
-        vavr   : "0.9.2",
+        flink   : "1.14.2",
+        hadoop  : "2.7.7",
+        junit4  : "4.11",
+        junit5  : "5.4.+",
+        mockito : "2.0.+",
+        mockito3: "3.+",
+        slf4j   : "1.7.0",
+        vavr    : "0.9.2",
```
sundargates, line 44: nit: maybe we should try to not align the colons given that this is constantly going to change based on the longest module name.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/resourceprovider/IResourceClusterProvider.java
```diff
@@ -0,0 +1,48 @@
...
+import io.mantisrx.control.plane.resource.cluster.proto.ResourceClusterProvisionSubmissionResponse;
+import io.mantisrx.control.plane.resource.cluster.proto.ScaleResourceRequest;
+import io.mantisrx.control.plane.resource.cluster.proto.ScaleResourceResponse;
+import java.util.concurrent.CompletionStage;
+
+/**
+ * This interface provides the API to connect resource cluster management actor to actual
+ * implementations of different resource cluster clients e.g. k8s.
+ */
+public interface IResourceClusterProvider {
```
sundargates, line 29: nit: let's rename this to ResourceClusterProvider. We want a consistent naming style for interfaces, and it's already mixed between I.* and regular interface names.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/MantisResourceClusterSpec.java
```diff
@@ -0,0 +1,156 @@
...
+            @JsonProperty("skuSpecs") final Set<SkuTypeSpec> skuSpecs,
+            @JsonProperty("clusterMetadataFields") final Map<String, String> clusterMetadataFields) {
+        this.name = name;
+        this.id = id;
+        this.ownerName = ownerName;
+        this.ownerEmail = ownerEmail;
+        this.envType = envType;
+        this.skuSpecs = skuSpecs;
+        this.clusterMetadataFields = clusterMetadataFields;
+    }
```
sundargates, line 82: I don't think you should need both @Value and @JsonCreator in the same class.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/MantisResourceClusterSpec.java
```diff
@@ -0,0 +1,156 @@
...
+            this.capacity = capacity;
+            this.imageId = imageId;
+            this.cpuCoreCount = cpuCoreCount;
+            this.memorySizeInBytes = memorySizeInBytes;
+            this.networkMbps = networkMbps;
+            this.diskSizeInBytes = diskSizeInBytes;
+            this.skuMetadataFields = skuMetadataFields;
+        }
+    }
+
```
sundargates, line 129: same here.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/MantisResourceClusterSpec.java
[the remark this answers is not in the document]
sundargates: Can you add some tests to make sure you codify the expected behaviors? In this case, I'm guessing you want the classes to be Jackson serializable (which they already are because of our lombok.config checked in at the root level).

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/MantisResourceClusterSpec.java
```diff
@@ -0,0 +1,156 @@
...
+        String skuId;
+
+        int minSize;
+
+        int maxSize;
+
+        int desireSize;
+
+        @JsonCreator
+        public SkuCapacity(
```
sundargates, line 143: nit: Don't think you need this.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/MantisResourceClusterSpec.java
```diff
@@ -0,0 +1,156 @@
...
+ * Contract class to define a Mantis resource cluster. This contract provides the abstraction to provide a generic
+ * definition from Mantis control perspective, and it's up to the implementations of each
+ * {@link io.mantisrx.control.plane.resource.cluster.resourceprovider.IResourceClusterProvider} to translate this spec
+ * to corresponding framework's cluster/node(s) definition.
+ */
+@Value
+@Builder
+public class MantisResourceClusterSpec {
+
+    @NonNull
```
sundargates, line 39: Let's not use the @NonNull annotation. By default, let's assume all fields are non-null by default. If something is nullable, use optional types in case that class is in the API layer (i.e., external-facing) or @Nullable if the field is only within a class (i.e., internal-facing).

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/MantisResourceClusterSpec.java
```diff
@@ -0,0 +1,156 @@
...
+    String id;
+
+    @NonNull
+    String ownerName;
+
+    @NonNull
+    String ownerEmail;
+
+    @NonNull
+    MantisResourceClusterEnvType envType;
```
sundargates, line 55: why do we need this? can this not be obtained from the env the mantis-control-plane is being run from?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/MantisResourceClusterSpec.java
```diff
@@ -0,0 +1,156 @@
...
+            this.memorySizeInBytes = memorySizeInBytes;
+            this.networkMbps = networkMbps;
+            this.diskSizeInBytes = diskSizeInBytes;
+            this.skuMetadataFields = skuMetadataFields;
+        }
+    }
+
+    @Builder
+    @Value
+    public static class SkuCapacity {
```
sundargates, line 132: nit: document.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/MantisResourceClusterSpec.java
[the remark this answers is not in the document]
Andyz26: The tricky part is that these work just fine without any of the extra annotations in OSS repo but started to run into trouble imported in the internal wrapper mixing with other non-shaded Jackson. I will have more tests covering in the nfmantis side.

build.gradle
Andyz26: I think editor config might help streamline these. I will take a look.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/MantisResourceClusterSpec.java
Andyz26: See the other conversation thread regarding the workaround for Jackson.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/MantisResourceClusterSpec.java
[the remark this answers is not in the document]
Andyz26: For OSS I will have some extra coverage on thee too in the follow up PR.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/control/plane/resource/cluster/proto/MantisResourceClusterSpec.java
Andyz26: I would prefer not to have an assumption about env var at this layer but leave it at the entry-point/wrapper to decide how to obtain it.

PR #182, opened 2022-04-18
Resource Cluster Aware Scheduler Introduction

sundargates:
### Context

This diff introduces a new scheduler based on resource cluster fundamentals. 

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/MantisSchedulerFactory.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/MantisSchedulerFactory.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2022 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.server.master.scheduler;
+
+import io.mantisrx.server.master.domain.JobDefinition;
+
+/**
+ * Factory for the Mantis Scheduler based on the JobDefinition
+ */
+public interface MantisSchedulerFactory {
+    MantisScheduler forJob(JobDefinition jobDefinition);
+}
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java
@@ -1,0 +1,304 @@
+/*
+ * Copyright 2022 Netflix, Inc.
+ *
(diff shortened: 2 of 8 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/MantisSchedulerFactory.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package io.mantisrx.server.master.scheduler;
+
+import io.mantisrx.server.master.domain.JobDefinition;
+
+/**
+ * Factory for the Mantis Scheduler based on the JobDefinition
+ */
+public interface MantisSchedulerFactory {
```
Andyz26, line 24: Shall we keep interface naming with prefix "I"?

sundargates: I'm not too fond of that convention, honestly. I don't mind following it if it's consistently followed. However, from whatever I can see from the current Mantis code, I don't think a lot of components follow that style.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java
```diff
@@ -0,0 +1,318 @@
...
+        this.resourceCluster = resourceCluster;
+        this.executeStageRequestFactory = executeStageRequestFactory;
+        this.jobMessageRouter = jobMessageRouter;
+        this.maxScheduleRetries = maxScheduleRetries;
+        this.intervalBetweenRetries = intervalBetweenRetries;
+        this.maxCancelRetries = maxCancelRetries;
+    }
+
+    @Override
+    public Receive createReceive() {
```
sundargates, line 76: what about when the scheduler is not master? should it stop scheduling requests etc..?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java
```diff
@@ -0,0 +1,318 @@
...
+            request.getScheduleRequest().getWorkerId());
+    }
+
+    private void onAssignedScheduleRequestEvent(AssignedScheduleRequestEvent event) {
+        try {
+            TaskExecutorGateway gateway =
+                resourceCluster.getTaskExecutorGateway(event.getTaskExecutorID()).join();
+
+            TaskExecutorRegistration info =
+                resourceCluster.getTaskExecutorInfo(event.getTaskExecutorID()).join();
```
Andyz26, line 120: Are the two 'get' calls here involve any long IO ops for blocking concern?

Andyz26 (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java
sundargates: if the scheduler is not master, then the default ServiceLifecycle shuts down the master and waits for it to get restarted.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java
sundargates: No, these should be pretty cheap. The only reason they get async responses is because they send fire-and-forget messages to the actors involved.

PR #198, opened 2022-04-22
Making ClassLoaderHandle public

sundargates:
### Context

### Checklist


Andyz26:
@sundargates can you add a brief context in PR?

codyrioux:
@Andyz26 I'm speculating but one benefit of the interface being public is that would-be OSS users can extend, implement and refer to it in their own namespaces if they wanted to implement some kind of wrapped version much like we do.

PR #199, opened 2022-04-23
Making sure there's a distinction between metrics publisher and metrics publisher's publish frequency

sundargates:
### Context

One of these properties is a prefix of the other. This works when defined in properties file but it doesn't when we want the props to be expressed in another DSL such as YAML.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/CoreConfiguration.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/CoreConfiguration.java
@@ -54,15 +54,11 @@ public interface CoreConfiguration {
     @Default("true")
     boolean isLocalMode();
 
-    @Config("mantis.metricsPublisher")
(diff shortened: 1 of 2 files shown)


mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/CoreConfiguration.java
```diff
@@ -54,15 +54,11 @@ public interface CoreConfiguration {
...
     @Default("io.mantisrx.common.metrics.MetricsPublisherNoOp")
     MetricsPublisher getMetricsPublisher();
 
     @Config("mantis.metricsPublisher.publishFrequencyInSeconds")
     @Default("15")
     int getMetricsPublisherFrequencyInSeconds();
-
-    @Config("mantis.metricsPublisher.config.prefix")
-    @Default("")
-    String getMetricsPublisherConfigPrefix();
```
sundargates, line 67: removed this because I couldn't find any usage for this.

PR #211, opened 2022-04-29
Refactoring MantisWorker so that the start behavior is consistent with v1.x version

sundargates:
### Context

We need MantisWorker::start to be blocking as the mesos implementation at Netflix makes certain assumptions about it. This diff gets it back to the old behavior. 

### Checklist


codyrioux (approved):
LGTM

fdc-ntflx (commented):
ship-it

PR #213, opened 2022-04-29
[2] Supporting custom key encoders for key-value stages in mantis

hmitnflx:
### Context

### Checklist


codyrioux (commented):
I love this idea for a change. I've often run into situations where I want to use a composite key and end up doing something silly like defining a separator and building a `String` which I then have to deconstruct in the next stage.

PR #212, opened 2022-04-29
[1] Supporting custom key encoders for key-value stages in mantis

hmitnflx:
- Allow user specified key serializer for key-value stages

### Context

Right now the keys inside MantisGroup are assumed to be String always.
This might not be the case in the future and so let's refactor to support
other encoders.

### Checklist

- [x] `./gradlew build` compiles code correctly
- [x] Added new tests where applicable
- [x] `./gradlew test` passes all tests
- [x] Extended README or added javadocs where applicable
- [x] Added copyright headers for new files from `CONTRIBUTING.md`

PR #214, opened 2022-04-30
Code refactor

hmitnflx:
- Preview for follow-up to supporting custom key encoders
- Code refactor to remove redundant generic datatype specification
- Refactor netty to mantis-common Codec in a separate location that can be code shared

### Context

### Checklist

### DIFF
--- a/mantis-runtime/src/main/java/io/mantisrx/runtime/executor/StageExecutors.java
+++ b/mantis-runtime/src/main/java/io/mantisrx/runtime/executor/StageExecutors.java
@@ -16,6 +16,8 @@
 
 package io.mantisrx.runtime.executor;
 
+import static io.mantisrx.runtime.parameter.ParameterUtils.STAGE_CONCURRENCY;
+
 import com.mantisrx.common.utils.Closeables;
 import io.mantisrx.common.MantisGroup;
 import io.mantisrx.common.metrics.Counter;
@@ -30,22 +32,18 @@ import io.mantisrx.runtime.source.Index;
 import io.mantisrx.server.core.ServiceRegistry;
 import io.reactivex.mantis.remote.observable.RxMetrics;
 import io.reactivx.mantis.operators.GroupedObservableUtils;
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import rx.Observable;
 import rx.functions.Action0;
 import rx.functions.Action1;
-import rx.functions.Func1;
 import rx.functions.Func2;
 import rx.internal.util.RxThreadFactory;
 import rx.observables.GroupedObservable;
 import rx.schedulers.Schedulers;
-
-import java.io.Closeable;
-import java.io.IOException;
-import java.util.concurrent.TimeUnit;
-
-import static io.mantisrx.runtime.parameter.ParameterUtils.STAGE_CONCURRENCY;
 
 
 public class StageExecutors {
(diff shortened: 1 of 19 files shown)


mantis-runtime/src/main/java/io/mantisrx/runtime/executor/StageExecutors.java
```diff
@@ -492,31 +479,24 @@ else if (stage instanceof ScalarToGroup) {
             toSink = setupScalarToGroupStage(scalarStage, source, context);
         } else if (stage instanceof KeyToKey) {
             KeyToKey keyToKey = (KeyToKey) stage;
-            Observable<Observable<GroupedObservable<String, T>>> source =
+            Observable<Observable<GroupedObservable<Object, T>>> source =
```
hmitnflx, line 482: I would like thoughts on this. I would ideally like to do this in a better way than just converting to an `Object`.

calvin681: I think key needs to be a string and cannot be arbitrary objects for group stages that are not the last. This may not be enforced anywhere, but stage may fail if key is not a string.

hmitnflx: I'm actually trying to change that (key limited to a `String` only) with #212 and a follow-up PR to this.
So far my observation is that the limitation is because of the hard-coded key encoders in use. If we move away from that and provide a way for users to specify key encoders, it should be fine.

calvin681: that's fair

calvin681 (approved):

mantis-runtime/src/main/java/io/mantisrx/runtime/executor/StageExecutors.java
codyrioux: Some immediate (and half-baked) thoughts that come to mind. They're more intended as discussion starters than suggestions.

1. What about creating a wrapper type `MantisGroupKey<T>` or something? It may still end up being `MantisGroupType<Object>` at this point but at least we can guarantee some interface.
2. Is it possible for us to store the type in the `StageConfig` class and fetch it out?
3. It is possible for us to use reflection here to determine the type?

codyrioux (approved):

mantis-runtime/src/main/java/io/mantisrx/runtime/executor/StageExecutors.java
Andyz26: +1 to @codyrioux 's comment above. 
If I understand correctly the 'k1' type info is lost during keyToKey <-> StageConfig input and if it is possible to store and retrieve the k1 type inside KeyToKey (which is a safe cast here with type check already) it would be great. If not then I feel like losing compile-time type check here is not a huge problem as this is something fairly easy to capture in dev phase when applying a wrong key codec.

hmitnflx: I took another stab at using the correct generic type to preserve the return type. 
Either way, I don't think it matters much because the mantis code doesn't handle generics very well. Moreover, the compiled code loses this type information anyway...

hmitnflx: I like the idea of moving the key type to `StageConfig` with support for `Void` keytype for scalar stages. #218 does this exact thing by promoting the keyCodec to that stage. I'll take a stab at another refactor in a subsequent PR.

PR #217, opened 2022-05-01
Move TaskExecutor to mantis-server-agent, remove dependency on mantis-runtime

calvin681:
### Context

The motivation for this change is that we want to run a bare bone TaskExecutor that does not depend on mantis-runtime. The Mantis job should bring its own runtime, and the executor can launch jobs with any runtime version. In order to achieve this, we introduce the ITask interface. Executor is only aware of this interface, runtime is expected to provide the concrete implementation.

Implications:
- nfmantis TitusMain should use mantis-server-agent instead of mantis-server-worker
- ITask does not support passing a Job object directly. If you are able to create a Job object, that means you are already runtime aware.
- TaskExecutor related classes are moved mantis-server-agent. Some common classes that require MantisMasterGateway are moved to mantis-control-plane-client. Majority of the PR is just moving classes.

### Checklist


mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
```diff
@@ -416,25 +421,31 @@ public CompletableFuture<Ack> submitTask(ExecuteStageRequest request) {
...
-            Optional.empty());
-
-        setCurrentTask(task);
-
-        scheduleRunAsync(this::startCurrentTask, 0, TimeUnit.MILLISECONDS);
-        return CompletableFuture.completedFuture(Ack.getInstance());
+        try {
+            UserCodeClassLoader userCodeClassLoader = ClassLoaderHandle.createUserCodeClassloader(request, classLoaderHandle);
+            ClassLoader cl = userCodeClassLoader.asClassLoader();
+            final ITask task = InstantiationUtil.instantiate("io.mantisrx.server.worker.Task", ITask.class, cl);
```
sundargates, line 427: What do you think about using ServiceLoader::load (https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html#load-java.lang.Class-java.lang.ClassLoader-) instead of this? This would help us avoid hardcoding the task class name and have that specified as part of the runtime. It's cleaner, IMO, as you can easily migrate to TaskV2 in the future by just changing the runtime. 

One more suggestion: instead of using setters for passing the context to the ITask instance, how about we create an ITask::Factory abstraction, which takes these parameters as context and returns an ITask instance as a return parameter. This would result in the runtime having to define an ITask::Factory implementation (say TaskFactory) as part of META-INF/services.

calvin681: Yes, we can try using ServiceLoader. That would be cleaner.

I'm not sure what introducing an extra layer of Factory buy us. Can you explain more? If the concern is called multiple setXXX methods, we could replace with a single init(...) method in the ITask interface. When we introduce ITaskV2, we would also need ITaskV2.Factory. I want to understand better what's the benefits of the added layer.

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
```diff
@@ -416,25 +421,34 @@ public CompletableFuture<Ack> submitTask(ExecuteStageRequest request) {
...
+            task.setUserCodeClassLoader(userCodeClassLoader);
+            task.setSinkSubscriptionStateHandlerFactory(subscriptionStateHandlerFactory);
+            task.setHostname(Optional.of(getHostname()));
+
+            setCurrentTask(task);
+
+            scheduleRunAsync(this::startCurrentTask, 0, TimeUnit.MILLISECONDS);
+            return CompletableFuture.completedFuture(Ack.getInstance());
+        } catch (Exception ex) {
+            return CompletableFutures.exceptionallyCompletedFuture(
```
Andyz26, line 443: Add log?

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
```diff
@@ -416,25 +421,34 @@ public CompletableFuture<Ack> submitTask(ExecuteStageRequest request) {
...
+                new TaskNotFoundException(null));
```
Andyz26, line 444: nit: wrap ex as cause?

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/MantisWorker.java
```diff
@@ -121,27 +125,33 @@ public void run() {
...
+                                        Clock.systemDefaultZone()));
+                                task.setJob(jobToRun);
+
+                                taskStatusUpdateSubscription =
+                                    task
+                                        .getStatus()
+                                        .subscribe(statusUpdateHandler::onStatusUpdate);
+                                task.startAsync();
+                            } catch (Exception ex) {
+                                throw new RuntimeException("worker failed to start", ex);
```
Andyz26, line 152: nit: add log

Andyz26 (approved):

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/Task.java
```diff
@@ -19,62 +19,85 @@
...
+
+    @Override
+    public void setUserCodeClassLoader(UserCodeClassLoader userCodeClassLoader) {
+        this.userCodeClassLoader = userCodeClassLoader;
+    }
+
+    @Override
+    public void setSinkSubscriptionStateHandlerFactory(SinkSubscriptionStateHandler.Factory sinkSubscriptionStateHandlerFactory) {
         this.sinkSubscriptionStateHandlerFactory = sinkSubscriptionStateHandlerFactory;
+    }
```
sundargates, line 92: nit: can we initialize these fields in one go?

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/Task.java
```diff
@@ -107,8 +130,7 @@ public void doRun() throws Exception {
                 workerMetricsClient,
                 sinkSubscriptionStateHandlerFactory,
                 hostname),
-            classLoaderHandle,
-            getJobProviderClass(),
+            getJobProviderClass(),userCodeClassLoader,
```
sundargates, line 133: nit: move userCodeClassLoader to the next line.

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/MantisWorker.java
```diff
@@ -121,27 +125,33 @@ public void run() {
...
+                                    .Factory
+                                    .forEphemeralJobsThatNeedToBeKilledInAbsenceOfSubscriber(
+                                        gateway,
+                                        Clock.systemDefaultZone()));
+                                task.setJob(jobToRun);
+
+                                taskStatusUpdateSubscription =
+                                    task
+                                        .getStatus()
+                                        .subscribe(statusUpdateHandler::onStatusUpdate);
```
sundargates, line 149: The task also needs to expose `vmTaskStatusSubject` which will have to be fed into the mesos executor callback handler. https://github.com/Netflix/mantis/blob/master/mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/MantisWorker.java#L155-L157

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/MantisWorker.java
```diff
@@ -22,12 +22,16 @@
 import io.mantisrx.runtime.Job;
 import io.mantisrx.server.core.BaseService;
 import io.mantisrx.server.core.Service;
+import io.mantisrx.server.core.WrappedExecuteStageRequest;
+import io.mantisrx.server.master.client.ClassLoaderHandle;
 import io.mantisrx.server.master.client.HighAvailabilityServices;
 import io.mantisrx.server.master.client.HighAvailabilityServicesUtil;
 import io.mantisrx.server.master.client.MantisMasterGateway;
+import io.mantisrx.server.master.client.SinkSubscriptionStateHandler;
```
sundargates, line 30: should MantisWorker move to the agent module as well?

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/ITask.java
```diff
@@ -0,0 +1,45 @@
...
+
+    void setMantisMasterGateway(MantisMasterGateway masterMonitor);
+
+    void setUserCodeClassLoader(UserCodeClassLoader userCodeClassLoader);
+
+    void setSinkSubscriptionStateHandlerFactory(SinkSubscriptionStateHandler.Factory sinkSubscriptionStateHandlerFactory);
+
+    void setHostname(Optional<String> hostname);
+
+    Observable<Status> getStatus();
```
sundargates, line 42: nit: we need to return an Observable<VirtualMachineTaskStatus> too.

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/ITask.java
```diff
@@ -0,0 +1,45 @@
...
+
+    void setWorkerConfiguration(WorkerConfiguration config);
+
+    void setMantisMasterGateway(MantisMasterGateway masterMonitor);
+
+    void setUserCodeClassLoader(UserCodeClassLoader userCodeClassLoader);
+
+    void setSinkSubscriptionStateHandlerFactory(SinkSubscriptionStateHandler.Factory sinkSubscriptionStateHandlerFactory);
+
+    void setHostname(Optional<String> hostname);
```
sundargates, line 40: rename to initialize()

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/MantisWorker.java
calvin681: See my commit description. MantisWorker is Job aware. It supports passing a Job object directly. It must depend on mantis-runtime. So it's not possible to move to agent.

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/ITask.java
calvin681: what do you mean by this?

sundargates: I updated the comment. Every task needs to return an observable that represents the state of the VM Machine. The Mesos executor callback handler is using this.

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/ITask.java
```diff
@@ -0,0 +1,45 @@
...
+
+    Observable<Status> getStatus();
+
+    WorkerId getWorkerId();
```
sundargates, line 39: don't think we need this since we create the task for a given worker request.

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/ITask.java
sundargates: Ignore this comment. It's not valid as discussed offline.

codyrioux (approved):

PR #218, opened 2022-05-02
[2] Supporting custom key encoders for key-value stages in mantis

hmitnflx:
- allow non-string codecs for input in mantis group to scalar stages
- pass the custom keyencoder to be used during deserialization

### Context
Right now the keys inside MantisGroup are assumed to be String always.
This might not be the case in the future and so let's refactor to support
other encoders. Also see companion PRs #212 and #214 

### Checklist


mantis-runtime/src/main/java/io/mantisrx/runtime/GroupToGroup.java
```diff
@@ -49,6 +50,11 @@
 
     GroupToGroup(GroupComputation<K1, T, K2, R> computation,
                  Config<K1, T, K2, R> config, Codec<T> inputCodec) {
+        this(computation, config, (Codec<K1>) Codecs.string(), inputCodec);
+    }
+
+    GroupToGroup(GroupComputation<K1, T, K2, R> computation,
```
hmitnflx, line 56: This actually results in a diamond inheritance. I'm thinking that avoiding `KeyValueInputStageConfig` and keeping the key encoder in `Stages` class is along with StageConfig is a good compromise. 
I'm unsure if I should make that optional or nullable, though. Thoughts?

mantis-runtime/src/main/java/io/mantisrx/runtime/ScalarStages.java
```diff
@@ -27,17 +27,17 @@
 public class ScalarStages<T> extends Stages<T> {
 
     ScalarStages(SourceHolder<?> source, StageConfig<?, ?> stage, Codec<T> inputCodec) {
-        super(source, stage, inputCodec);
+        super(source, stage, null, inputCodec);
     }
 
     ScalarStages(Stages<?> self, StageConfig<?, ?> stage, Codec<T> inputCodec) {
-        super(self.getSource(), self.getStages(), stage, inputCodec);
+        super(self.getSource(), self.getStages(), stage, null, inputCodec);
```
hmitnflx, line 34: thoughts on Optional vs nullable here?

calvin681:
Can you put in the PR description the motivation for non-string keys?

mantis-runtime/src/main/java/io/mantisrx/runtime/KeyValueInputStageConfig.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package io.mantisrx.runtime;
+
+import io.mantisrx.common.codec.Codec;
+import io.mantisrx.common.codec.Codecs;
+import io.mantisrx.runtime.parameter.ParameterDefinition;
+import java.util.List;
+
+public class KeyValueInputStageConfig<K, T, R> extends StageConfig<T, R> {
```
calvin681, line 24: why not add InputKeyCodec and OutputKeyCodec to StageConfig directly? is this class necessary?

hmitnflx: I can definitely do that. That's exactly my question [here](https://github.com/Netflix/mantis/pull/218#discussion_r863155877)

calvin681 (approved):

PR #219, opened 2022-05-04
Adding serialVersionUIDs to all the DS that are serialized during RPC

sundargates:
### Context

This helps avoid the JVM in generating these version IDs and also makes it easy to keep these data structures backward compatible.

### Checklist


mantis-common/src/main/java/io/mantisrx/common/Ack.java
```diff
@@ -24,4 +24,5 @@
  */
 @Value(staticConstructor = "getInstance")
 public class Ack implements Serializable {
+    private static final long serialVersionUID = 1L;
```
Andyz26, line 27: nit: can you add a link to the doc about how this interact with JVM?

Andyz26 (approved):

mantis-common/src/main/java/io/mantisrx/common/Ack.java
sundargates: +1

codyrioux (approved):
I love the notion of using `1L` and incrementing it should we want to break compatibility vs the random numbers that people sometimes use. LGTM.

PR #222, opened 2022-05-09
Handle Reconciliation only for Mesos Workers

sundargates:
### Context

Support for Titus clusters has resulted in the set of workers launched as per Fenzo not matching the group of workers launched according to the WorkerRegistry. This diff aims to fix that.

### Checklist


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/events/WorkerRegistryV2.java
```diff
@@ -64,10 +66,11 @@ public void clearState() {
      * @return
      */
     @Override
-    public int getNumRunningWorkers() {
+    public int getNumRunningWorkers(@Nullable ClusterID resourceCluster) {
```
Andyz26, line 69: Add doc about the null indication.

Andyz26 (approved):

sundargates:
> looks good to me. qq: why null in some places vs optional in other?

I wouldn't say I usually like to use Optional parameters in a method. This link (https://www.baeldung.com/java-optional#misuages) broadly explains the thinking.

Issue #194, opened 2022-04-22
Get rid of all Object Mappers throughout the codebase


sundargates:
### Context

Object Mappers are used throughout the codebase. The problem with this is that they are all configured differently. Therefore, if we need to introduce a new module, we will have to change all the ObjectMappers. 

### Steps to reproduce

Explain the steps to reliably reproduce the issue.

### Expected behavior

Explain what you think should happen.

### Actual Behavior

Explain what actually happens instead.

PR #224, opened 2022-05-10
Consolidation ObjectMapper Usage (Step 1)

ilooner:
A small first step towards consolidating usages of ObjectMapper by deleting unecessary classes that use ObjectMapper.

 - Removing SimpleSchedulerObserver because there is no usages anywhere of it.
 - Removing FileBasedOffsetCheckpointStrategy because it was only intended for use in testing, but it was exposed to production users through the API. If we really need to unit test offset commits in the future, I think we can just use a simple in memory implementation.

### Context

This is a first step for resolving the ObjectMapper removal issue https://github.com/Netflix/mantis/issues/194

### Checklist


hmitnflx (approved):

sundargates (approved):
Thanks for identifying that FileBasedOffsetCheckpoint Strategy was only intended for testing purposes.

PR #228, opened 2022-05-12
Consolidation ObjectMapper Usage (Step 4)

ilooner:
### Context

Switched DefaultMantisApiClient from ObjectMapper to JsonSerializer as part of https://github.com/Netflix/mantis/issues/194 . I also moved the testing code to a test class, and tested the change manually against our production api server and a source job.

### Checklist

### DIFF
--- a/mantis-publish/mantis-publish-core/build.gradle
+++ b/mantis-publish/mantis-publish-core/build.gradle
@@ -22,6 +22,7 @@ ext {
 
 dependencies {
     implementation "io.mantisrx:mql-jvm:$mqlVersion"
+    implementation project(":mantis-common-serde")
     api project(":mantis-discovery-proto")
 
     api "com.netflix.spectator:spectator-api:$spectatorVersion"
@@ -38,6 +39,7 @@ dependencies {
     testImplementation libraries.junitJupiter
     testImplementation libraries.mockitoCore
     testImplementation libraries.slf4jLog4j12
+    testImplementation libraries.commonsIo
 }
 
 test {
--- a/mantis-publish/mantis-publish-core/src/test/java/io/mantisrx/publish/internal/discovery/mantisapi/DefaultMantisApiClientTest.java
+++ b/mantis-publish/mantis-publish-core/src/test/java/io/mantisrx/publish/internal/discovery/mantisapi/DefaultMantisApiClientTest.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2022 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.publish.internal.discovery.mantisapi;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.get;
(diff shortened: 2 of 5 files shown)


mantis-publish/mantis-publish-core/src/test/java/io/mantisrx/publish/internal/discovery/mantisapi/DefaultMantisApiClientTest.java
```diff
@@ -0,0 +1,59 @@
...
+import io.mantisrx.publish.config.SampleArchaiusMrePublishConfiguration;
+import java.util.Optional;
+import java.util.Properties;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+public class DefaultMantisApiClientTest {
+    @Disabled
```
hmitnflx, line 35: why add it as disabled?

mantis-publish/mantis-publish-core/src/test/java/io/mantisrx/publish/internal/discovery/mantisapi/DefaultMantisApiClientTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+        props.put("mantis.publish.discovery.api.hostname", apiServerHostName);
+        config.setProperties(props);
+
+        final PropertyRepository propsRepo = DefaultPropertyFactory.from(config);
+        final DefaultMantisApiClient defaultMantisApiClient = new DefaultMantisApiClient(new SampleArchaiusMrePublishConfiguration(propsRepo), HttpClient.create(registry));
+        final CompletableFuture<JobDiscoveryInfo> jobDiscoveryInfoCompletableFuture = defaultMantisApiClient.jobDiscoveryInfo(jobClusterName);
+        final JobDiscoveryInfo jobDiscoveryInfo = jobDiscoveryInfoCompletableFuture.get(1, TimeUnit.SECONDS);
+
+        System.out.println(jobDiscoveryInfo);
```
hmitnflx, line 52: Could you please test key-value pairs in `assertEquals`, `assertTrue`, etc instead? I don't think it's a good idea to just print to console for testing.

ilooner: @hmitnflx I'm not sure what to check because those values will depend on the job cluster you point the client at. If you have some pointers to setup an api server + source job in a unit test, I can try that and add some asserts then.

mantis-publish/mantis-publish-core/build.gradle
```diff
@@ -22,6 +22,7 @@ ext {
 
 dependencies {
     implementation "io.mantisrx:mql-jvm:$mqlVersion"
+    implementation project(":mantis-common")
```
Andyz26, line 25: [For ur consideration] mantis-common probably will bring in quite some unused deps to the publish side of things which becomes part of the client MRE deps. I feel that it would be nice if the serde part can be separated out in this case.

ilooner: @Andyz26 K I will try making `mantis-serde` which will depend on `mantisShaded`. `mantisShaded` has some extra stuff in it, but that's already being pulled into the client through `mantis-discovery-proto`. I'll open a seperate PR for seperating out the serde and get that other PR merged before this one.

mantis-publish/mantis-publish-core/src/test/java/io/mantisrx/publish/internal/discovery/mantisapi/DefaultMantisApiClientTest.java
ilooner: @hmitnflx I spoke too soon. I saw this morning other tests use WireMock to mock the api server. I have updated this test to use WireMock stubs as well, added asserts, and enabled the test. Let me know what you think.

mantis-publish/mantis-publish-core/build.gradle
ilooner: Updated this pr to use `mantis-common-serde`

sundargates (approved):
LGTM!

hmitnflx (approved):

PR #238, opened 2022-07-15
Ensure there is only 1 subscription in PushTrigger

calvin681:
### Context

There is an implicit assumption that subscription should be 1. Before, the subscriptionActive gauge is always set to 1 whenever there is at least 1 subscription. I change that to increment/decrement on subscribe/unsubscribe to more accurately reflect the number of subscriptions. Before, the unsubscribe was purposely commented out. Clearly, this can lead to multiple subscription, which violates the assumption. With this change, we will unsubscribe the old subscription when a new subscription is created. This should correct keep active subscription at 1.

### Checklist

### DIFF
--- a/mantis-network/src/main/java/io/reactivex/mantis/network/push/ObservableTrigger.java
+++ b/mantis-network/src/main/java/io/reactivex/mantis/network/push/ObservableTrigger.java
@@ -22,11 +22,13 @@ import io.mantisrx.common.metrics.Metrics;
 import io.reactivx.mantis.operators.DisableBackPressureOperator;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import rx.Observable;
+import rx.Scheduler;
 import rx.Subscription;
 import rx.functions.Action0;
 import rx.functions.Action1;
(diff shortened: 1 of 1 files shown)


hmitnflx (approved):

mantis-network/src/main/java/io/reactivex/mantis/network/push/ObservableTrigger.java
```diff
@@ -155,60 +162,68 @@ private static <K, V> PushTrigger<KeyValuePair<K, V>> groupTrigger(final String
...
-                                () -> {
-                                    logger.info("Observable used to push data completed, on server with name: " + name);
-                                    if (doOnComplete != null) {
-                                        doOnComplete.call();
-                                    }
-                                }
-
-                        ));
+                    // decouple from calling thread
+                    .observeOn(Schedulers.computation())
```
codyrioux, line 169: Thoughts on computation scheduler vs IO scheduler for this task?

codyrioux (approved):

mantis-network/src/main/java/io/reactivex/mantis/network/push/ObservableTrigger.java
calvin681: Actually I wonder why we need to specify a schedule at all. It can just run on whatever scheduler it's already running on. @hmitnflx do you know? Maybe I can just remove this line.

mantis-network/src/main/java/io/reactivex/mantis/network/push/ObservableTrigger.java
```diff
@@ -54,33 +54,40 @@ private static <T> PushTrigger<T> trigger(final String name, final Observable<T>
...
+                    .doOnSubscribe(() -> {
+                        logger.info("Subscription is ACTIVE for observable trigger with name: " + name);
+                        subscriptionActive.increment();
+                    })
+                    .doOnUnsubscribe(() -> {
+                        logger.info("Subscription is INACTIVE for observable trigger with name: " + name);
+                        subscriptionActive.decrement();
+                    })
+                    .subscribe((T data) -> queue.write(data)
+                        ,
```
sundargates, line 70: nit: move this to the previous line?

PR #245, opened 2022-08-13
Handle load archive job cache error

Andyz26:
### Context

Currently, when the archive job cache tried to load any out-of-date archive job it will trigger an error thrown from the cache get and becomes an unhandled error (to radar detection as well).
Adding the IOException type to the catch clause to reduce the error noise.

### Checklist


hmitnflx (approved):

sundargates:
Is there a reason why we are loading old archived jobs? should we fix that instead?

Andyz26:
> Is there a reason why we are loading old archived jobs? should we fix that instead?

@sundargates My understanding is that this could be from stale actors on failed jobs that didn't get persisted at all. Treating it as a cache miss and the get() call to a cache should not be considered critical/unhandled exception anyway.

PR #253, opened 2022-08-16
Making sure container attribute field is not getting serialized as part of TaskExecutorRegistration

sundargates:
### Context

I was noticing that taskExecutorContainerDefinitionId was getting serialized as part of the TaskExecutorRegistration state. This diff aims to get rid of that. 
```
       "taskExecutorContainerDefinitionId": {
            "resourceID": "some value"
        }
```

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
@@ -19,6 +19,7 @@ import io.mantisrx.common.WorkerConstants;
 import io.mantisrx.common.WorkerPorts;
 import io.mantisrx.runtime.MachineDefinition;
 import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonCreator;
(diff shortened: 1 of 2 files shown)


mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
```diff
@@ -36,36 +37,36 @@
  * Different fields help identify the task executor in different dimensions.
  */
 @Builder
-@FieldDefaults(makeFinal=true, level= AccessLevel.PRIVATE)
+@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
```
Andyz26, line 40: Can you remind me why this is not on "@value" tag?

Andyz26 (approved):

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
sundargates: because I want to support deserializing two types of TaskExecutorRegistrations. It is not possible to do this with the value tag.

PR #254, opened 2022-08-19
Support resource cluster based job submission API

Andyz26:
### Context

Add support in job submission API to specify the target resource cluster. This change is supposed to be back-compat.
* Also making "labels" in JobDefintion to be unique on each label name.

### Checklist

### DIFF
--- a/mantis-common/src/main/java/io/mantisrx/runtime/descriptor/DeploymentStrategy.java
+++ b/mantis-common/src/main/java/io/mantisrx/runtime/descriptor/DeploymentStrategy.java
@@ -16,10 +16,13 @@
 
 package io.mantisrx.runtime.descriptor;
 
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonInclude;
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonInclude.Include;
 import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.Map;
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
+import lombok.Getter;
 import lombok.Singular;
 import lombok.ToString;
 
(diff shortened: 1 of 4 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobDefinition.java
```diff
@@ -141,6 +151,16 @@ public void validateSchedulingInfo() throws InvalidJobException {
...
     }
 
+    private void postProcess() {
+        if (this.deploymentStrategy != null && !Strings.isNullOrEmpty(this.deploymentStrategy.getResourceClusterId())) {
+            this.labels.put(
+                SystemLabels.MANTIS_RESOURCE_CLUSTER_NAME_LABEL.label,
+                new Label(
+                    SystemLabels.MANTIS_RESOURCE_CLUSTER_NAME_LABEL.label,
+                    this.deploymentStrategy.getResourceClusterId()));
+        }
```
sundargates, line 161: What's the rationale for having this information in both DeploymentStrategy as well as in the form of labels?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobDefinition.java
```diff
@@ -50,7 +60,7 @@ public class JobDefinition {
     private final SchedulingInfo schedulingInfo;
     private final DeploymentStrategy deploymentStrategy;
     private final int withNumberOfStages;
-    private List<Label> labels;
+    private Map<String, Label> labels; // Map label->name to label instance.
```
sundargates, line 63: This representation feels weird, given that a label is essentially a key-value pair. However, if this is done for dedup purposes, then we might as well convert this into a Map<String, String> and (key, value) tuple representing the label.

mantis-common/src/test/java/io/mantisrx/runtime/descriptor/DeploymentStrategyTest.java
```diff
@@ -64,4 +69,33 @@ public void shouldNotRequireInheritInstanceCheck() {
...
+        String expected = "{"
+            + "\"stageDeploymentStrategyMap\":"
+            + "{"
+            + "\"1\":{\"inheritInstanceCount\":false},"
+            + "\"2\":{\"inheritInstanceCount\":true},"
+            + "\"3\":{\"inheritInstanceCount\":true}}"
+            + "}";
+
+        final DeploymentStrategy ds = serializer.fromJSON(expected, DeploymentStrategy.class);
+        assertEquals(expected.replaceAll("[\\n\\s]+", ""), serializer.toJson(ds));
```
sundargates, line 99: Add a check on what the resourceClusterID is for this JSON.

mantis-common/src/main/java/io/mantisrx/runtime/descriptor/DeploymentStrategy.java
```diff
@@ -16,34 +16,50 @@
...
+    @Getter
+    private final Map<Integer, StageDeploymentStrategy> stageDeploymentStrategyMap;
+
+    /**
+     * If this field is not empty, it's used to indicate this is a resource cluster stack deployment and will be hosted
+     * to the given resource cluster Id.
+     */
+    @Getter
+    @JsonInclude(Include.NON_NULL)
+    private final String resourceClusterId;
```
sundargates, line 43: Also, can we conform to ClusterID type here instead of String?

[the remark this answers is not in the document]
Andyz26: Since this is the request payload and we cannot guarantee the sender to serialize the optional field I think make it optional won't help much.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobDefinition.java
Andyz26: IMO the system label is more of a impl detail and having this leaked to public API is putting too much logic to the sender/UI side.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobDefinition.java
Andyz26: It's to ensure the unqiueness of per-label-name to label instance. I feel that converting 'Label' instance is not necessary in this case (String/Key, Object instance) seems to be a fairly normal way to track?

mantis-common/src/test/java/io/mantisrx/runtime/descriptor/DeploymentStrategyTest.java
Andyz26: Good point.

mantis-common/src/main/java/io/mantisrx/runtime/descriptor/DeploymentStrategy.java
Andyz26: Actually, I realized this is the runtime contract that doesn't have dep on the "clusterID" in the mantis control plane. Given the fact that this is to be used to set the label value underlying the strong type doesn't help much (since it's going to be converted to label string again).

mantis-common/src/test/java/io/mantisrx/runtime/descriptor/DeploymentStrategyTest.java
Andyz26: It's covered in test "testSerialization" above.

sundargates (approved):

PR #256, opened 2022-08-26
Support get active jobs per ResourceCluster

Andyz26:
### Context

Add API to fetch active jobs for a given resource cluster ID.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
@@ -289,6 +289,7 @@ class ResourceClusterActor extends AbstractActorWithTimers {
             .filter(Objects::nonNull)
             .map(WorkerId::getJobId)
             .distinct()
+            .sorted((String::compareToIgnoreCase))
             .skip(req.getStartingIndex().orElse(0))
             .limit(req.getPageSize().orElse(3000))
             .collect(Collectors.toList());
@@ -686,6 +687,7 @@ class ResourceClusterActor extends AbstractActorWithTimers {
 
     @Value
     @Builder
+    @AllArgsConstructor // needed for build to work with custom ctor.
     static class GetActiveJobsRequest {
         ClusterID clusterID;
         Optional<Integer> startingIndex;
@@ -695,12 +697,6 @@ class ResourceClusterActor extends AbstractActorWithTimers {
             this.clusterID = clusterID;
             this.pageSize = Optional.empty();
             this.startingIndex = Optional.empty();
-        }
-
-        public GetActiveJobsRequest(ClusterID clusterID, Optional<Integer> startingIndex, Optional<Integer> pageSize) {
-            this.clusterID = clusterID;
-            this.pageSize = pageSize;
-            this.startingIndex = startingIndex;
         }
     }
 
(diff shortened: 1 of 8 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -277,6 +281,28 @@ private TaskExecutorsList getTaskExecutors(Predicate<Entry<TaskExecutorID, TaskE
...
 
+    private void getActiveJobs(GetActiveJobsRequest req) {
+        List<String> pagedList = this.taskExecutorStateMap
+            .values()
+            .stream()
+            .map(TaskExecutorState::getWorkerId)
+            .filter(Objects::nonNull)
+            .map(WorkerId::getJobId)
+            .distinct()
+            .skip(req.getStartingIndex().orElse(0))
```
sundargates, line 292: It might be a good idea to do a sort here since the list will not be deterministic otherwise.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -658,6 +684,26 @@ static class GetRegisteredTaskExecutorsRequest {
...
+            this.clusterID = clusterID;
+            this.pageSize = Optional.empty();
+            this.startingIndex = Optional.empty();
+        }
+
+        public GetActiveJobsRequest(ClusterID clusterID, Optional<Integer> startingIndex, Optional<Integer> pageSize) {
+            this.clusterID = clusterID;
+            this.pageSize = pageSize;
+            this.startingIndex = startingIndex;
+        }
```
sundargates, line 704: nit: you don't need to explicitly write this constructor since you have an `@Value` annotation for the class which brings in the `@AllArgsConstructor`.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterAkkaImpl.java
```diff
@@ -198,6 +201,23 @@ public CompletableFuture<Ack> disableTaskExecutorsFor(Map<String, String> attrib
                 .toCompletableFuture();
     }
 
+    @Override
+    public CompletableFuture<PagedActiveJobOverview> getActiveJobOverview(
+        Optional<Integer> startingIndex,
+        Optional<Integer> maxSize) {
```
sundargates, line 207: nit: I prefer not to use Optionals in method parameters. Here's some context: http://dolszewski.com/java/java-8-optional-use-cases/

sundargates (approved):
LGTM, except for how Optionals are used as method and constructor parameters. It's not a big deal, though.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
Andyz26: It seems if I don't add it explicitly or use '@allArgsConstructor' the extra ctor broke the '@value' and it won't compile.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterAkkaImpl.java
Andyz26: I read the link and this case actually feels like the exact case where the inputs are already optional<T> from the akka HTTP API for optional params and it doesn't create the extra wrapping on the caller side.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
Andyz26: Good catch!

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterAkkaImpl.java
sundargates: That makes sense. let's keep it this way.

PR #257, opened 2022-08-30
Fix dead letter Ack from RC scaler

Andyz26:
### Context

The ack returned to self() after the async pipe operation is redirected to the dead-letter actor since the self() reference is lost. I am removing the extra Ack from the call.

### Checklist


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterScalerActor.java
```diff
@@ -259,9 +259,6 @@ private void onTriggerClusterUsageRequest(TriggerClusterUsageRequest req) {
     private void onTriggerClusterRuleRefreshRequest(TriggerClusterRuleRefreshRequest req) {
         log.info("{}: Requesting cluster rule refresh", this.clusterId);
         this.fetchRuleSet();
-
-        // when calling from API there is no need to wait.
-        getSender().tell(Ack.getInstance(), self());
```
sundargates, line 264: Is this always triggered by the self-actor?

I'm wondering if we should just ignore the ack message sent to this actor (hopefully from the actor itself).

sundargates (approved):

PR #259, opened 2022-09-01
Making sure MasterConfiguration returns ResourceClusterStorageProvider type instead of just the className

sundargates:
### Context

This will ensure orgs can plug in their ResourceClusterStorageProvider implementations.

### Checklist


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/MasterMain.java
```diff
@@ -162,13 +161,10 @@ public MasterMain(ConfigurationFactory configFactory, AuditEventSubscriber audit
...
             Configuration configuration = loadConfiguration();
 
-            ResourceClusterStorageProviderAdapter resourceClusterStorageProvider =
-                new ResourceClusterStorageProviderAdapter(this.config.getResourceClusterStorageProvider(), system);
-
             final ActorRef resourceClustersHostActor = system.actorOf(
                 ResourceClustersHostManagerActor.props(
                     new ResourceClusterProviderAdapter(this.config.getResourceClusterProvider(), system),
-                    resourceClusterStorageProvider),
+                    config.getResourceClusterStorageProvider()),
```
Andyz26, line 167: I think the adapter was added to pass in the system parameter. Without it how do we share actor system with the provider impl? Or are you proposing the provider impl to always uses its own actor system?

Andyz26 (approved):

Andyz26 (approved):

PR #262, opened 2022-09-13
Handle null return values in load/get call for guava cacheloader

hmitnflx:
### Context

This effectively reverts #245 where we introduced a null return value in load/get method for guava cacheloader which isn't liked. I'm using an optional return value instead with an explicit null on any exception.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/MantisJobStore.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/MantisJobStore.java
@@ -382,7 +382,7 @@ public class MantisJobStore {
 
     private class ArchivedJobsMetadataCache {
 
-        private final Cache<String, IMantisJobMetadata> cache;
+        private final Cache<String, Optional<IMantisJobMetadata>> cache;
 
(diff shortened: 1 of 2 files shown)


Andyz26 (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/MantisJobStore.java
```diff
@@ -393,33 +393,25 @@ private class ArchivedJobsMetadataCache {
 
         IMantisJobMetadata getJob(String jobId) {
             try {
-                return cache.get(jobId, () -> loadArchivedJob(jobId));
-            } catch (ExecutionException e) {
+                final Optional<IMantisJobMetadata> jobMetadata = cache.get(jobId, () -> loadArchivedJobImpl(jobId));
+                return jobMetadata.orElse(null);
```
sundargates, line 397: Isn't this problematic? We should not cache "no value" as when there is a value available for the same key in the future, we will return "no value" as it was cached.

Also, feel that this is not addressing the core of the issue which is why are we missing the value in the first place and if the value can sometimes be missing, shouldn't the API take that into account when calling this?

PR #263, opened 2022-09-21
Defaulting to the mesos cluster in case of non-terminal events with unknown job definitions

sundargates:
### Context

When we receive non-terminal events (such as heartbeats) for jobs that are currently not running according to the control plane, we send them to the resource cluster specified in their job definition to decommission them. However, we ignore the event if there's a failure to load the job definition at this point. This diff changes this behavior to assume that all ghost workers that don't have corresponding job definitions belong to mesos. 

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/MantisSchedulerFactory.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/MantisSchedulerFactory.java
@@ -18,13 +18,18 @@ package io.mantisrx.server.master.scheduler;
 
 import io.mantisrx.server.master.domain.JobDefinition;
 import io.mantisrx.server.master.resourcecluster.ClusterID;
+import java.util.Optional;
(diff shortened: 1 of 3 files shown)


hmitnflx (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/MantisSchedulerFactory.java
```diff
@@ -17,10 +17,20 @@
...
 public interface MantisSchedulerFactory {
     MantisScheduler forJob(JobDefinition jobDefinition);
+
+    /**
+     * returns the MantisScheduler based on the ClusterID.
+     *
+     * @param clusterID cluster ID for which the mantisscheduler is requested.
+     * @return MantisScheduler corresponding to the ClusterID.
+     */
+    MantisScheduler forClusterID(@Nullable ClusterID clusterID);
```
Andyz26, line 35: nit: any reason not to use optional here directly?

Andyz26 (approved):

PR #268, opened 2022-10-03
Revive runtime isolation + ITask Factory

Andyz26:
### Context

Revive PR: #217 
Add ITask Factory to allow customization to ITask implementation and classloader override.

### Checklist


calvin681 (approved):

build.gradle
```diff
@@ -35,14 +35,15 @@ buildscript {
...
-        mockito3: "3.+",
-        slf4j   : "1.7.0",
-        vavr    : "0.9.2",
+        flink    : "1.14.2",
+        hadoop   : "2.7.7",
+        junit4   : "4.11",
+        junit5   : "5.4.+",
+        mockito  : "2.0.+",
+        mockito3 : "3.+",
+        spectator: "1.3.+",
```
sundargates, line 44: We should figure out how to replace spectator-api with micrometer eventually https://micrometer.io/docs/concepts#_purpose

Andyz26: insight team's doc is actually suggesting not to use micrometer but stick with spectator. I feel like there is already a good base of mantis metrics abstraction and we just need to fully remove all direct spectator refs there.

mantis-control-plane/mantis-control-plane-client/build.gradle
```diff
@@ -17,10 +17,13 @@
...
     implementation libraries.spotifyFutures
+    implementation libraries.hadoopCommon
     api project(":mantis-control-plane:mantis-control-plane-core")
 
     api project(":mantis-remote-observable")
 
     testImplementation libraries.junit4
     testImplementation libraries.mockitoAll
+    testImplementation libraries.spectatorApi
+    testImplementation(testFixtures(project(":mantis-common")))
```
sundargates, line 28: Instead of doing this for every project, can we just put `testImplementation libraries.spectatorApi` in the root `build.gradle` for all subprojects? This way, we can remove it once rather than in all `build.gradle` files.

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/ClassLoaderHandle.java
```diff
@@ -13,8 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.mantisrx.server.worker;
+package io.mantisrx.server.master.client;
```
sundargates, line 16: Honestly, I'm not sure if this is the right package for this class. mantis-control-plane-client is supposed to hold all the stuff relating to talking to the mantis-control-plane. So, I'm not sure if this code belongs there. 

I would put this under mantis/taskexecutor/api or mantis/server/worker/client.

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/ITask.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.flink.util.UserCodeClassLoader;
+import rx.Observable;
+
+public interface ITask extends Service {
+
+    void initialize(WrappedExecuteStageRequest request,
+                    WorkerConfiguration config,
+                    MantisMasterGateway masterMonitor,
+                    UserCodeClassLoader userCodeClassLoader,
+                    SinkSubscriptionStateHandler.Factory sinkSubscriptionStateHandlerFactory,
```
sundargates, line 34: Should we get rid of these dependencies, as it would be hard to keep them consistent as we make iterations on the code? IMO, we can just keep ExecuteStageRequest, Properties instead of WorkerConfiguration?

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/config/MetricsCollector.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.mantisrx.server.worker.metrics;
+package io.mantisrx.server.master.client.config;
```
sundargates, line 17: Same as above. Doesn't belong here IMO.

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/ITask.java
Andyz26: These are the ones I hope to convert to fully serializable parameters to avoid any ref sharing across ITask.

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/config/WorkerConfigurationUtils.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+package io.mantisrx.server.master.client.config;
+
+import io.mantisrx.server.core.MetricsCoercer;
+import java.util.Properties;
+import org.skife.config.ConfigurationObjectFactory;
+
+public class WorkerConfigurationUtils {
```
sundargates, line 23: I'm guessing this library needs to live in TaskExecutor module.

mantis-common/src/main/java/io/mantisrx/common/metrics/Metrics.java
```diff
@@ -179,6 +180,7 @@ public Map<MetricId, Timer> timers() {
     }
 
     public static class Builder {
+        private static Iterable<Tag> CommonTags = Collections.emptyList();
```
sundargates, line 183: What's the rationale for having this singleton state, given that we already have the SpectatorRegistryFactory, which allows you to set a registry with a given set of common tags?

Andyz26: Inside Springboot job runtime the registry is no longer built/control by mantis worker and there is no API right now to set common tags in the SBN registry injected.

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/config/WorkerConfigurationUtils.java
Andyz26: It's also needed on nfmantis worker titus. Let me see if I can move this one to runtime-loader instead.

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/config/MetricsCollector.java
Andyz26: This one is directly referened by WorkerConfiguration which needs to stay inside control-plane-client.

Andyz26:
@sundargates For "SinkParameter" it's possible to move it to control-plane-core but its actual package name is being used in source jobs too and renaming it will create a large splash of changes. In this case I don't think it's worth much to only move it to core module without renaming it.

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/ClassLoaderHandle.java
```diff
@@ -13,8 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.mantisrx.server.worker;
+package io.mantisrx.runtime.loader;
 
+import io.mantisrx.server.core.ExecuteStageRequest;
```
sundargates, line 18: I'm guessing this is coming from mantis-control-plane-client?

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/ClassLoaderHandle.java
```diff
@@ -64,4 +65,12 @@ public void close() throws IOException {
...
         };
     }
+
+    static UserCodeClassLoader createUserCodeClassloader(ExecuteStageRequest executeStageRequest, ClassLoaderHandle classLoaderHandle) throws Exception {
+        // triggers the download of all missing jar files from the job manager
+        final UserCodeClassLoader userCodeClassLoader =
+            classLoaderHandle.getOrResolveClassLoader(ImmutableList.of(executeStageRequest.getJobJarUrl().toURI()), ImmutableList.of());
+
+        return userCodeClassLoader;
+    }
```
sundargates, line 75: Instead of having this as a static method, you can have it as a default method of the ClassLoaderHandle interface. This way, it's a method of the class rather than a utility which is what a static method is. 

```
default UserCodeClassLoader createUserCodeClassLoader(ExecuteStageRequest executeStageRequest)
```

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/ITask.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.flink.util.UserCodeClassLoader;
+import rx.Observable;
+
+public interface ITask extends Service {
+
+    void initialize(WrappedExecuteStageRequest request,
+                    WorkerConfiguration config,
+                    MantisMasterGateway masterMonitor,
+                    UserCodeClassLoader userCodeClassLoader,
+                    Factory sinkSubscriptionStateHandlerFactory,
```
sundargates, line 36: Does this need to be passed at all? I feel like it can live as part of the task definition.

mantis-runtime-loader/README.md
```diff
@@ -0,0 +1,4 @@
+# Mantis-runtime-loader
+
+This module is used to host loader classes used by mantis worker (e.g. TaskExecutor) to load given job instance's runtime code and user code.
```
sundargates, line 3: nice :-)

mantis-server/mantis-server-agent/build.gradle
```diff
@@ -15,8 +15,21 @@
  */
 
 dependencies {
-    compile project(":mantis-control-plane:mantis-control-plane-core")
+    implementation project(":mantis-runtime-loader")
+    implementation project(":mantis-server:mantis-server-worker-client")
```
sundargates, line 19: Isn't this a transitive dependency? Do we need to define this explicitly?

mantis-runtime-loader/build.gradle
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+dependencies {
+    implementation libraries.hadoopCommon
```
sundargates, line 18: I don't think you need to depend on `hadoopCommon` given that I don't see any symbols from it in this package.

mantis-server/mantis-server-agent/build.gradle
```diff
@@ -15,8 +15,21 @@
...
+    implementation libraries.commonsIo
+    implementation libraries.hadoopCommon
+    implementation libraries.spotifyFutures
 
     implementation libraries.slf4jApi
     implementation libraries.slf4jLog4j12
+
+    testImplementation libraries.junit4
+    testImplementation libraries.mockitoAll
+    testImplementation project(":mantis-server:mantis-server-worker")
```
sundargates, line 33: Is there a reason why we depend on mantis-server-worker here? Shouldn't it be the other way around? In that, I expect fixtures from this module to be used from within the mantis-server worker.

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/cgroups/CpuAcctsSubsystemProcess.java
```diff
@@ -14,9 +14,9 @@
  * limitations under the License.
  */
 
-package io.mantisrx.server.worker.metrics.cgroups;
+package io.mantisrx.server.agent.cgroups;
```
sundargates, line 17: s/agent.cgroups/agent.metrics.cgroups instead?

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/config/WorkerConfigurationUtils.java
sundargates: any updates on this?

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/config/WorkerConfiguration.java
```diff
@@ -14,10 +14,9 @@
  * limitations under the License.
  */
 
-package io.mantisrx.server.worker.config;
+package io.mantisrx.server.master.client.config;
```
sundargates, line 17: can this also be part of runtime-loader?

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/ClassLoaderHandle.java
Andyz26: yeah

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/ITask.java
Andyz26: We can trim down a bit here but the overall impact is still the same. I think atm I prefer we refactor these to something serializable instead.

mantis-server/mantis-server-agent/build.gradle
Andyz26: this is "server-worker-client" and the loader only depends on "control-plane-client".

mantis-server/mantis-server-agent/build.gradle
Andyz26: "TaskExecutor" is in server-agent and there is a integration test for it and the test requires full worker as testImplementation to run.

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/config/WorkerConfiguration.java
Andyz26: Moved.

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/TaskFactory.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Interface to factory building ITask implementation instance. Can be override to use customized ITask impl.
+ */
+public interface TaskFactory {
+    ITask getITaskInstance(ExecuteStageRequest request, ClassLoader cl);
+
+    UserCodeClassLoader getUserCodeClassLoader(
+        ExecuteStageRequest request,
+        ClassLoaderHandle classLoaderHandle);
```
sundargates, line 30: What's the rationale for this to be part of TaskFactory? This has got nothing to do with TaskFactory semantically.

Andyz26: The usercode class loader is needed to get the ITask impl (and also required as param in get ITask method).

Andyz26: In the current structure the TaskExecutor needs to make calls to this interface while the impl of the interface is controlled by worker module.

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
```diff
@@ -97,17 +104,30 @@ public class TaskExecutor extends RpcEndpoint implements TaskExecutorGateway {
...
+    }
+
     public TaskExecutor(
         RpcService rpcService,
         WorkerConfiguration workerConfiguration,
         HighAvailabilityServices highAvailabilityServices,
         ClassLoaderHandle classLoaderHandle,
-        Factory subscriptionStateHandlerFactory) {
+        SinkSubscriptionStateHandler.Factory subscriptionStateHandlerFactory,
+        TaskFactory taskFactory) {
```
sundargates, line 130: Annotate this as @Nullable

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/SingleTaskOnlyFactory.java
```diff
@@ -0,0 +1,50 @@
...
+import io.mantisrx.server.core.ExecuteStageRequest;
+import java.util.ServiceLoader;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.flink.util.UserCodeClassLoader;
+
+/**
+ * This factory is used when there is only 1 ITask implementation.
+ */
+@Slf4j
+public class SingleTaskOnlyFactory implements TaskFactory {
```
sundargates, line 31: Does this need to be public?

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
```diff
@@ -425,19 +446,41 @@ public CompletableFuture<Ack> submitTask(ExecuteStageRequest request) {
...
+                wrappedRequest,
+                workerConfiguration,
+                masterMonitor,
+                userCodeClassLoader,
+                subscriptionStateHandlerFactory,
+                Optional.of(getHostname())
+            );
+
+            scheduleRunAsync(() -> {
+                setCurrentTask(task);
```
sundargates, line 472: setCurrentTask(task) should be run on the main thread right?

Andyz26: yeah and that's exactly what "scheduleRunAsync" does (schedule task in main thread).

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
```diff
@@ -425,19 +446,41 @@ public CompletableFuture<Ack> submitTask(ExecuteStageRequest request) {
...
+                startCurrentTask();
+            }, 0, TimeUnit.MILLISECONDS);
+        } catch (Exception ex) {
+            log.error("Failed to submit task, request: {}", wrappedRequest.getRequest(), ex);
+            listeners.enqueue(getTaskFailedEvent(null, ex));
+        }
+        finally {
+            getIOExecutor().execute(listeners::dispatch);
+        }
```
sundargates, line 481: [Note to self] go through this one more time.

mantis-server/mantis-server-worker/build.gradle
```diff
@@ -25,18 +25,11 @@ ext {
...
     compile project(":mantis-server:mantis-server-worker-client")
 
-    api libraries.flinkRpcApi
-    implementation libraries.commonsIo
-    implementation libraries.hadoopCommon
-    implementation libraries.zip4j
-    implementation libraries.hadoopS3
-    implementation libraries.spotifyFutures
-    implementation libraries.flinkRpcImpl
-
```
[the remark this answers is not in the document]
sundargates: The structure I'm thinking about is mantis-taskexecutor depends on mantis-runtime-loader, and mantis-runtime also depends on mantis-runtime-loader, but mantis-taskexecutor directly loads the task definition from mantis-runtime. Why have a separate module?

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/Task.java
```diff
@@ -157,4 +174,16 @@ public Observable<VirtualMachineTaskStatus> getVMStatus() {
...
+        if (userCodeClassLoader != null) {
+            if (userCodeClassLoader.asClassLoader() instanceof Closeable) {
+                try {
+                    ((Closeable) userCodeClassLoader.asClassLoader()).close();
+                } catch (IOException ex) {
+                    log.error("Failed to close user class loader successfully", ex);
+                }
+            }
+        }
+    }
```
sundargates, line 188: hmm, the task is not opening the class loader - so why is it given the responsibility of closing the class loader? Should this be part of the actor that's actually opening the class loader?

PR #277, opened 2022-10-30
Making sure resource cluster storage provider does not assume the root directory

sundargates:
### Context

Resource Cluster Storage Provider was making assumptions about what the root directory was for the resource cluster storage provider. This diff aims to get rid of these assumptions. 

### Checklist


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/resourceprovider/SimpleFileResourceClusterStorageProvider.java
```diff
@@ -260,15 +265,16 @@ public CompletionStage<Boolean> updateClusterScaleRules(ResourceClusterScaleRule
             .thenApply(ioRes -> true); // IOResult result is always true.
     }
 
-    private Path getClusterSpecFilePath(String dirName, String clusterId) {
-        return Paths.get(dirName, "clusterspec-" + clusterId);
+    @VisibleForTesting
+    File getClusterSpecFilePath(String clusterId) {
```
calvin681, line 269: nit: remove the `Path` suffix in these methods

Andyz26 (approved):

PR #280, opened 2022-10-31
Move getUserCodeClassLoader to interface default

Andyz26:
### Context

Move getUserCodeClassLoader to have a default interface version.

### Checklist

### DIFF
--- a/mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/TaskFactory.java
+++ b/mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/TaskFactory.java
@@ -25,7 +25,13 @@ import org.apache.flink.util.UserCodeClassLoader;
 public interface TaskFactory {
     RuntimeTask getRuntimeTaskInstance(ExecuteStageRequest request, ClassLoader cl);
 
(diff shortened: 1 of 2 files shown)


mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/TaskFactory.java
```diff
@@ -25,7 +25,13 @@
 public interface TaskFactory {
     RuntimeTask getRuntimeTaskInstance(ExecuteStageRequest request, ClassLoader cl);
 
-    UserCodeClassLoader getUserCodeClassLoader(
+    default UserCodeClassLoader getUserCodeClassLoader(
```
sundargates, line 28: I'm unsure if this is the right place for this method, given that it doesn't pertain to creating a Task itself. I would maybe move this to some Utils method or have the logic as part of ClassLoaderHandle::createUserCodeClassLoader.

Andyz26: sounds good. I will keep a note on this too.

PR #284, opened 2022-11-03
Add JobArtifact model and storeNewArtifact function in StorageProvider

fdc-ntflx:
Add JobArtifact model and storeNewArtifact function in StorageProvider interface

### Context

This PR aims to define the model for Job artifacts we plan to store going forward. A new functionality has been added to MantisStorageProvider to be able to store new job artifacts (right now it only saves 1 artifact at the time, maybe in the future we can store more than one at the time).

Note: before merging I plan to change `JobArtifact` model a bit to have a list of dependencies where the runtime is only 1 of the dependencies; can be empty if the artifact is in the form of a uber jar.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/domain/JobArtifact.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/domain/JobArtifact.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2022 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.server.core.domain;
+
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonProperty;
+import java.time.Instant;
+import java.util.Map;
+import lombok.Builder;
+import lombok.Value;
+
+@Value
+@Builder
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class JobArtifact {
+    // Job artifact unique identifier
+    @JsonProperty("artifactID")
+    ArtifactID artifactID;
+    // Job artifact name
+    @JsonProperty("name")
+    String name;
+    // Job artifact version
+    @JsonProperty("version")
+    String version;
+    // Job artifact creation timestamp
+    @JsonProperty("createdAt")
+    Instant createdAt;
+    // Runtime Type Identification. For java potential values: spring-boot, guice
+    @JsonProperty("runtimeType")
+    String runtimeType;
+    // Job artifact external dependencies
+    // Example: {"mantis-runtime": "1.0.0"}
+    @JsonProperty("dependencies")
+    Map<String, String> dependencies;
+    // Job entrypoint clas.
+    @JsonProperty("entrypoint")
+    String entrypoint;
+}
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/MasterApiAkkaService.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/MasterApiAkkaService.java
@@ -30,6 +30,8 @@ import akka.stream.Materializer;
 import akka.stream.javadsl.Flow;
 import com.netflix.spectator.impl.Preconditions;
 import io.mantisrx.master.api.akka.route.MantisMasterRoute;
+import io.mantisrx.master.api.akka.route.handlers.JobArtifactRouteHandler;
+import io.mantisrx.master.api.akka.route.handlers.JobArtifactRouteHandlerImpl;
 import io.mantisrx.master.api.akka.route.handlers.JobClusterRouteHandler;
 import io.mantisrx.master.api.akka.route.handlers.JobClusterRouteHandlerAkkaImpl;
 import io.mantisrx.master.api.akka.route.handlers.JobDiscoveryRouteHandler;
@@ -48,6 +50,7 @@ import io.mantisrx.master.api.akka.route.v0.JobStatusRoute;
 import io.mantisrx.master.api.akka.route.v0.MasterDescriptionRoute;
 import io.mantisrx.master.api.akka.route.v1.AdminMasterRoute;
 import io.mantisrx.master.api.akka.route.v1.AgentClustersRoute;
+import io.mantisrx.master.api.akka.route.v1.JobArtifactsRoute;
 import io.mantisrx.master.api.akka.route.v1.JobClustersRoute;
 import io.mantisrx.master.api.akka.route.v1.JobDiscoveryStreamRoute;
 import io.mantisrx.master.api.akka.route.v1.JobStatusStreamRoute;
@@ -60,9 +63,9 @@ import io.mantisrx.server.core.master.MasterDescription;
 import io.mantisrx.server.core.master.MasterMonitor;
 import io.mantisrx.server.master.ILeadershipManager;
 import io.mantisrx.server.master.LeaderRedirectionFilter;
-import io.mantisrx.server.master.persistence.IMantisStorageProvider;
 import io.mantisrx.server.master.resourcecluster.ResourceClusters;
 import io.mantisrx.server.master.scheduler.MantisScheduler;
+import io.mantisrx.server.master.store.MantisStorageProvider;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
@@ -83,7 +86,7 @@ public class MasterApiAkkaService extends BaseService {
     private final ResourceClusters resourceClusters;
     private final ActorRef statusEventBrokerActor;
     private final int port;
-    private final IMantisStorageProvider storageProvider;
+    private final MantisStorageProvider storageProvider;
     private final MantisScheduler scheduler;
     private final LifecycleEventPublisher lifecycleEventPublisher;
     private final MantisMasterRoute mantisMasterRoute;
@@ -100,7 +103,7 @@ public class MasterApiAkkaService extends BaseService {
                                 final ResourceClusters resourceClusters,
                                 final ActorRef resourceClustersHostManagerActor,
                                 final int serverPort,
-                                final IMantisStorageProvider mantisStorageProvider,
+                                final MantisStorageProvider mantisStorageProvider,
                                 final MantisScheduler scheduler,
                                 final LifecycleEventPublisher lifecycleEventPublisher,
                                 final ILeadershipManager leadershipManager,
@@ -170,6 +173,9 @@ public class MasterApiAkkaService extends BaseService {
         final LastSubmittedJobIdStreamRoute v1LastSubmittedJobIdStreamRoute = new LastSubmittedJobIdStreamRoute(jobDiscoveryRouteHandler);
         final JobStatusStreamRoute v1JobStatusStreamRoute = new JobStatusStreamRoute(jobStatusRouteHandler);
 
+        final JobArtifactRouteHandler jobArtifactRouteHandler = new JobArtifactRouteHandlerImpl(storageProvider);
+        final JobArtifactsRoute v1JobArtifactsRoute = new JobArtifactsRoute(jobArtifactRouteHandler);
+
         final LeaderRedirectionFilter leaderRedirectionFilter = new LeaderRedirectionFilter(masterMonitor, leadershipManager);
         final ResourceClusterRouteHandler resourceClusterRouteHandler = new ResourceClusterRouteHandlerAkkaImpl(
             resourceClustersHostManagerActor);
@@ -184,6 +190,7 @@ public class MasterApiAkkaService extends BaseService {
             v0AgentClusterRoute,
             v1JobClusterRoute,
             v1JobsRoute,
+            v1JobArtifactsRoute,
             v1AdminMasterRoute,
             v1AgentClustersRoute,
             v1JobDiscoveryStreamRoute,
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/Jackson.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/Jackson.java
@@ -131,7 +131,7 @@ public class Jackson {
                            });
     }
 
-    private static String toJSON(
+    public static String toJSON(
             ObjectMapper mapper,
             FilterProvider filters,
             Object object) throws JsonProcessingException {
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/JobArtifactRouteHandlerImpl.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/JobArtifactRouteHandlerImpl.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.master.api.akka.route.handlers;
+
+import static io.mantisrx.master.jobcluster.proto.BaseResponse.ResponseCode.SERVER_ERROR;
+import static io.mantisrx.master.jobcluster.proto.BaseResponse.ResponseCode.SUCCESS;
+
+import io.mantisrx.master.jobcluster.proto.JobArtifactProto;
+import io.mantisrx.server.core.domain.JobArtifact;
+import io.mantisrx.server.master.store.MantisStorageProvider;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class JobArtifactRouteHandlerImpl implements JobArtifactRouteHandler {
+    private final MantisStorageProvider mantisStorageProvider;
+
+    public JobArtifactRouteHandlerImpl(MantisStorageProvider mantisStorageProvider) {
+        this.mantisStorageProvider = mantisStorageProvider;
+    }
+
+    @Override
+    public CompletionStage<JobArtifactProto.SearchJobArtifactsResponse> search(JobArtifactProto.SearchJobArtifactsRequest request) {
+        return CompletableFuture.supplyAsync(() -> {
+            try {
+                final List<JobArtifact> jobArtifactList = mantisStorageProvider.listJobArtifacts(request.getName(), request.getVersion());
+                return new JobArtifactProto.SearchJobArtifactsResponse(request.requestId, SUCCESS, "", jobArtifactList);
+            } catch (IOException e) {
+                log.warn("Error while fetching job artifacts. Traceback: {}", e.getMessage(), e);
+                return new JobArtifactProto.SearchJobArtifactsResponse(request.requestId, SERVER_ERROR, e.getMessage(), Collections.emptyList());
+            }
+        });
+    }
+
+    @Override
+    public CompletionStage<JobArtifactProto.ListJobArtifactsByNameResponse> listArtifactsByName(JobArtifactProto.ListJobArtifactsByNameRequest request) {
+        return CompletableFuture.supplyAsync(() -> {
+            try {
+                final List<String> artifactNames = mantisStorageProvider.listJobArtifactsByName(request.getPrefix());
+                return new JobArtifactProto.ListJobArtifactsByNameResponse(request.requestId, SUCCESS, "", artifactNames);
+            } catch (IOException e) {
+                log.warn("Error while searching job artifact names. Traceback: {}", e.getMessage(), e);
+                return new JobArtifactProto.ListJobArtifactsByNameResponse(request.requestId, SERVER_ERROR, e.getMessage(), Collections.emptyList());
+            }
+        });
+    }
+
+    @Override
+    public CompletionStage<JobArtifactProto.UpsertJobArtifactResponse> upsert(JobArtifactProto.UpsertJobArtifactRequest request) {
+        final JobArtifact jobArtifact = request.getJobArtifact();
+        return CompletableFuture.supplyAsync(() -> {
+            try {
+                mantisStorageProvider.addNewJobArtifact(jobArtifact);
+                return new JobArtifactProto.UpsertJobArtifactResponse(request.requestId, SUCCESS, "", jobArtifact.getArtifactID());
+            } catch (IOException e) {
+                log.warn("Error while storing new job artifact. Traceback: {}", e.getMessage(), e);
+                return new JobArtifactProto.UpsertJobArtifactResponse(request.requestId, SERVER_ERROR, e.getMessage(), jobArtifact.getArtifactID());
+            }
+        });
+    }
+}
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/proto/JobArtifactProto.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/proto/JobArtifactProto.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.master.jobcluster.proto;
+
+import com.netflix.spectator.impl.Preconditions;
+import io.mantisrx.server.core.domain.ArtifactID;
+import io.mantisrx.server.core.domain.JobArtifact;
+import java.util.List;
+import lombok.EqualsAndHashCode;
+import lombok.Value;
+
+public class JobArtifactProto {
+
+    @EqualsAndHashCode(callSuper = true)
+    @Value
+    public static class SearchJobArtifactsRequest extends BaseRequest {
+        String name;
+        String version;
+
+        public SearchJobArtifactsRequest(String name, String version) {
+            super();
+            Preconditions.checkNotNull(name, "JobArtifact name cannot be null");
+            this.name = name;
+            this.version = version;
+        }
+    }
+
+    @EqualsAndHashCode(callSuper = true)
+    @Value
+    public static class SearchJobArtifactsResponse extends BaseResponse {
+        List<JobArtifact> jobArtifacts;
+
+        // TODO(fdichiara): add paginated list.
+        public SearchJobArtifactsResponse(
+            long requestId,
+            ResponseCode responseCode,
+            String message,
+            List<JobArtifact> jobArtifacts) {
+            super(requestId, responseCode, message);
+            this.jobArtifacts = jobArtifacts;
+        }
+    }
+
+    @EqualsAndHashCode(callSuper = true)
+    @Value
+    public static class ListJobArtifactsByNameRequest extends BaseRequest {
+        String prefix;
+
+        public ListJobArtifactsByNameRequest(String prefix) {
+            super();
+            this.prefix = prefix;
+        }
+    }
+
+    @EqualsAndHashCode(callSuper = true)
+    @Value
+    public static class ListJobArtifactsByNameResponse extends BaseResponse {
+        List<String> names;
+
+        // TODO(fdichiara): add paginated list.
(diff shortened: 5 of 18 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/proto/JobArtifactProto.java
```diff
@@ -0,0 +1,47 @@
...
+public class JobArtifactProto {
+
+    public static final class ListJobArtifactsRequest extends BaseRequest {
+        public ListJobArtifactsRequest() {
+            super();
+        }
+    }
+
+    @Getter
+    @ToString
```
sundargates, line 33: Use @Value instead since this is a value class.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
```diff
@@ -0,0 +1,55 @@
...
+        @JsonProperty("entrypoint") String entrypoint
+    ) {
+        this.name = name;
+        this.project = project;
+        this.version = version;
+        this.createdAt = createdAt;
+        this.runtimeType = runtimeType;
+        this.runtimeVersion = runtimeVersion;
+        this.entrypoint = entrypoint;
+    }
```
sundargates, line 54: Use @Value instead and you can get rid of this.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
```diff
@@ -0,0 +1,55 @@
...
+@Data
+@Builder
+public class JobArtifact {
+    private final String name;
+    private final String project;
+    private final String version;
+    private final long createdAt;
+    private final String runtimeType;
+    private final String runtimeVersion;
+    private final String entrypoint;
```
sundargates, line 34: nit: add comments.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
```diff
@@ -0,0 +1,55 @@
...
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+public class JobArtifact {
+    private final String name;
+    private final String project;
+    private final String version;
+    private final long createdAt;
```
sundargates, line 31: nit: use Instant instead.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
```diff
@@ -0,0 +1,55 @@
...
+    private final String runtimeType;
+    private final String runtimeVersion;
```
sundargates, line 33: These are very specific to how Netflix stores artifacts. I would maybe generalize this into Map<String, String> tags.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
```diff
@@ -0,0 +1,55 @@
...
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonCreator;
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+public class JobArtifact {
+    private final String name;
```
sundargates, line 28: Instead of name, should we use id instead? I'm guessing you want this to be unique globally. If so, id would be a better name.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/MantisStorageProvider.java
```diff
@@ -183,4 +183,12 @@ List<MantisWorkerMetadataWritable> getArchivedWorkers(String jobid)
     void deleteExpiredDisableTaskExecutorRequest(DisableTaskExecutorsRequest request) throws IOException;
 
     List<DisableTaskExecutorsRequest> loadAllDisableTaskExecutorsRequests(ClusterID clusterID) throws IOException;
+
+    void storeNewJobArtifact(JobArtifact jobArtifact) throws JobArtifactAlreadyExistsException, IOException;
+
+    void initJobArtifacts();
```
sundargates, line 189: nit: It appears that this method only has side effects instead of returning the current snapshot. However, currently, MantisStorageProvider is not implemented as an actor or as a service. Instead, it's purely used as a data access layer. Given that, wondering if we should return a List<JobArtifact> here.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/NoopStorageProvider.java
```diff
@@ -162,4 +163,20 @@ public void deleteExpiredDisableTaskExecutorRequest(DisableTaskExecutorsRequest
...
+    public void storeNewJobArtifact(JobArtifact jobArtifact) throws JobArtifactAlreadyExistsException, IOException {
+    }
+
+    @Override
+    public void initJobArtifacts(){
+    }
+
+    @Override
+    public boolean jobArtifactExists(String name){
+        return true;
```
sundargates, line 177: shouldn't this default to false.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
sundargates: I would go even one step further and define an ArtifactID symbol and use that here. Then, I would refer to Artifact through that ID everywhere.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifactAlreadyExistsException.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.server.master.store;
+
+public class JobArtifactAlreadyExistsException extends Exception {
+
+    public JobArtifactAlreadyExistsException(String message) {
```
sundargates, line 21: nit: refer to the ArtifactID.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/JobArtifactRouteHandlerImpl.java
```diff
@@ -0,0 +1,69 @@
...
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class JobArtifactRouteHandlerImpl implements JobArtifactRouteHandler {
+    private static final Logger logger = LoggerFactory.getLogger(JobArtifactRouteHandlerImpl.class);
```
Andyz26, line 35: nit: use "@slf4j" annotation instead.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/JobArtifactRouteHandlerImpl.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public JobArtifactRouteHandlerImpl(MantisStorageProvider mantisStorageProvider) {
+        this.mantisStorageProvider = mantisStorageProvider;
+    }
+
+    @Override
+    public CompletionStage<JobArtifactProto.ListJobArtifactsResponse> get(JobArtifactProto.ListJobArtifactsRequest request) {
+        return CompletableFuture.supplyAsync(() -> {
+            try {
+                final List<JobArtifact> jobArtifactList = mantisStorageProvider.listJobArtifacts();
```
Andyz26, line 47: "request" var is not used?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/JobArtifactRouteHandlerImpl.java
```diff
@@ -0,0 +1,69 @@
...
+                return new JobArtifactProto.ListJobArtifactsResponse(request.requestId, SUCCESS, "", jobArtifactList);
+            } catch (IOException e) {
+                logger.warn("Error while fetching job artifacts. Traceback: {}", e.getMessage(), e);
+                return new JobArtifactProto.ListJobArtifactsResponse(request.requestId, CLIENT_ERROR, e.getMessage(), Collections.emptyList());
```
Andyz26, line 51: is this "client_error"?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/JobArtifactRouteHandlerImpl.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    public CompletionStage<JobArtifactProto.CreateJobArtifactResponse> create(JobArtifactProto.CreateJobArtifactRequest request) {
+        final JobArtifact jobArtifact = request.getJobArtifact();
+        return CompletableFuture.supplyAsync(() -> {
+            try {
+                mantisStorageProvider.storeNewJobArtifact(jobArtifact);
+                return new JobArtifactProto.CreateJobArtifactResponse(request.requestId, SUCCESS, "", jobArtifact.getArtifactID());
+            } catch (IOException | JobArtifactIDAlreadyExistsException e) {
+                logger.warn("Error while storing new job artifact. Traceback: {}", e.getMessage(), e);
+                return new JobArtifactProto.CreateJobArtifactResponse(request.requestId, CLIENT_ERROR, e.getMessage(), jobArtifact.getArtifactID());
```
Andyz26, line 65: "io exception" probably doesn't map to client error.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/JobArtifactRouteHandlerImpl.java
```diff
@@ -0,0 +1,69 @@
...
+                return new JobArtifactProto.ListJobArtifactsResponse(request.requestId, SUCCESS, "", jobArtifactList);
+            } catch (IOException e) {
+                logger.warn("Error while fetching job artifacts. Traceback: {}", e.getMessage(), e);
+                return new JobArtifactProto.ListJobArtifactsResponse(request.requestId, CLIENT_ERROR, e.getMessage(), Collections.emptyList());
+            }
+        });
+    }
+
+    @Override
+    public CompletionStage<JobArtifactProto.CreateJobArtifactResponse> create(JobArtifactProto.CreateJobArtifactRequest request) {
```
Andyz26, line 57: is this insert or override?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/ArtifactID.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package io.mantisrx.server.master.store;
+
+import lombok.RequiredArgsConstructor;
+import lombok.Value;
+
+@RequiredArgsConstructor(staticName="of")
+@Value
+public class ArtifactID {
```
Andyz26, line 24: nit: doc with example.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
```diff
@@ -0,0 +1,44 @@
...
+    @JsonProperty("version")
+    String version;
+    @JsonProperty("createdAt")
+    Instant createdAt;
+    @JsonProperty("runtimeType")
+    String runtimeType;
+    @JsonProperty("dependencies")
+    Map<String, String> dependencies;
+    @JsonProperty("entrypoint")
+    String entrypoint;
```
Andyz26, line 43: is this needed?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
```diff
@@ -0,0 +1,44 @@
...
+    @JsonProperty("artifactID")
+    ArtifactID artifactID;
+    @JsonProperty("project")
+    String project;
+    @JsonProperty("version")
+    String version;
+    @JsonProperty("createdAt")
+    Instant createdAt;
+    @JsonProperty("runtimeType")
+    String runtimeType;
```
Andyz26, line 39: SBN vs non-SBN? nit: add doc about candiates.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/MantisStorageProvider.java
```diff
@@ -183,4 +183,12 @@ List<MantisWorkerMetadataWritable> getArchivedWorkers(String jobid)
     void deleteExpiredDisableTaskExecutorRequest(DisableTaskExecutorsRequest request) throws IOException;
 
     List<DisableTaskExecutorsRequest> loadAllDisableTaskExecutorsRequests(ClusterID clusterID) throws IOException;
+
+    void storeNewJobArtifact(JobArtifact jobArtifact) throws JobArtifactIDAlreadyExistsException, IOException;
+
+    List<JobArtifact> initJobArtifacts();
```
Andyz26, line 189: what's this one for? Do we need to load whole list in master memory?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/JobArtifactRouteHandlerImpl.java
fdc-ntflx: at the moment the request doesn't have any params since it fetches all. `request` is only used to create the response using the `responseId` .

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/JobArtifactRouteHandlerImpl.java
fdc-ntflx: good catch. this is server error.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/JobArtifactRouteHandlerImpl.java
fdc-ntflx: I think this is an `upsert`. will rename it.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
fdc-ntflx: you mean the `entrypoint`? good question. I foresee TEs in the future to only use metadata from DB to instantiate a job, so I guess the entrypoint is important?

fdc-ntflx: What do you think?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
```diff
@@ -0,0 +1,44 @@
...
+    @JsonProperty("dependencies")
+    Map<String, String> dependencies;
```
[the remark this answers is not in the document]
fdc-ntflx: I'm thinking depth 1 for now as `Map<dep_name, dep_version>` . What do you think?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
fdc-ntflx: +1

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/MantisStorageProvider.java
fdc-ntflx: in theory yes that was the original idea but now I'm not using it anymore so maybe I should get rid of it.

hmitnflx: @fdc-ntflx As part of PR #285, I'm merging the functionality in these two classes together. It'd save me some post-merge migration if you could resubmit this diff on top of that PR. Probably saves you some testing pain as well hopefully!

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/JobArtifactRouteHandlerImpl.java
Andyz26: Got it. So I assume filtering etc. is will be added there?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
Andyz26: The current structure has some metadata files inside job jar to mark runtime type and 'entry class'.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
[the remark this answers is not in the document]
Andyz26: btw what's the usage for this? This might make the API response unexpectedly large.

Andyz26:
LGTM. Can you track follow-up/TODOs in the PR description or issue also?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
fdc-ntflx: yeah I'm thinking about getting rid of that file in the future. this gets us one step closer!

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/JobArtifactRouteHandlerImpl.java
fdc-ntflx: I still have to figure that out but probably yes.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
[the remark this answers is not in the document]
fdc-ntflx: good point. The first use case for this is to include the mantis runtime lib + version without being too explicit about it (since other companies may build a uber jar and not having to specify the runtime)

mantis-control-plane/mantis-control-plane-server/build.gradle
```diff
@@ -31,6 +31,7 @@ ext {
 dependencies {
     api project(':mantis-control-plane:mantis-control-plane-core')
 
+    api "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.11.2"
```
sundargates, line 34: I think this dep should be shaded in mantis-shaded and used through that. Are there reasons why we don't want to shade this?

fdc-ntflx: I think the import is using mantis-shaded already so I think there is no point in adding this.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/MasterApiAkkaService.java
```diff
@@ -100,7 +103,7 @@ public MasterApiAkkaService(final MasterMonitor masterMonitor,
                                 final ResourceClusters resourceClusters,
                                 final ActorRef resourceClustersHostManagerActor,
                                 final int serverPort,
-                                final IMantisStorageProvider mantisStorageProvider,
+                                final MantisStorageProvider mantisStorageProvider,
```
sundargates, line 106: I would prefer we don't make these changes in this diff - @hmitnflx is working on cleaning this abstraction.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/Jackson.java
```diff
@@ -39,7 +40,8 @@ public class Jackson {
     private static final ObjectMapper defaultObjectMapper = new ObjectMapper()
             .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
             .configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false)
-            .registerModule(new Jdk8Module());
+            .registerModule(new Jdk8Module())
+            .registerModule(new JavaTimeModule());
```
sundargates, line 44: This has some implications, I think. For example, with this change, timestamps would be serialized differently, which may not be what our clients expect. Would it be possible to avoid this?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/MasterApiAkkaService.java
fdc-ntflx: If the other PR is ready to go I'm fine to rebase.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/ArtifactID.java
```diff
@@ -0,0 +1,29 @@
...
+
+import lombok.RequiredArgsConstructor;
+import lombok.Value;
+
+/**
+ * Represents the job artifact that mantis can run.
+ */
+@RequiredArgsConstructor(staticName="of")
+@Value
+public class ArtifactID {
```
sundargates, line 27: Shouldn't this be part of the client? I would assume you would want to expose APIs around which artifacts are available in mantis, used by which job clusters, and so on.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
```diff
@@ -0,0 +1,45 @@
...
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonProperty;
+import java.time.Instant;
+import java.util.Map;
+import lombok.Builder;
+import lombok.Value;
+
+@Value
+@Builder
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class JobArtifact {
```
sundargates, line 29: Same as above.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
```diff
@@ -0,0 +1,45 @@
...
+    ArtifactID artifactID;
+    @JsonProperty("project")
+    String project;
+    @JsonProperty("version")
+    String version;
+    @JsonProperty("createdAt")
+    Instant createdAt;
+    // Potential values: spring-boot, guice
+    @JsonProperty("runtimeType")
+    String runtimeType;
```
sundargates, line 40: This feels too tied to Netflix. Did you have a chance to consider my earlier proposal of having generic tags instead?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/Jackson.java
fdc-ntflx: Are you sure this is the case?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/JobArtifact.java
fdc-ntflx: I don't think having a map for any extra config is a good idea. we need much better validation on which key is present, changing clients will be harder if we change the keys in the map etc etc. I think runtime type is also something other companies can leverage in case of migrations. thoughts?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/Jackson.java
fdc-ntflx: There are tests where we serialize and deserialize timestamps we should break if this new module had implications, no?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/MasterApiAkkaService.java
fdc-ntflx: spoke to @hmitnflx and he needs more time for testing the change. I will go head and merge this for now.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/Jackson.java
sundargates: I prefer to avoid this change if possible. I'm not so sure about the coverage of the tests themselves.

mantis-control-plane/mantis-control-plane-server/dependencies.lock
```diff
@@ -5,138 +5,8 @@
         }
     },
     "baseline-exact-dependencies-main": {
-        "com.github.spullara.cli-parser:cli-parser": {
```
sundargates, line 8: Why was this file updated if there is no build.gradle changes?

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/domain/JobArtifact.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.Value;
+
+@Value
+@Builder
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class JobArtifact {
+    @JsonProperty("artifactID")
+    ArtifactID artifactID;
+    @JsonProperty("project")
+    String project;
```
sundargates, line 33: nit: Add comments for each of the fields.

sundargates, line 33: nit: I also think artifact name is a better name for this field than project itself. It kind of aligns with the java artifact model as well.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/MantisStorageProvider.java
```diff
@@ -183,4 +186,12 @@ List<MantisWorkerMetadataWritable> getArchivedWorkers(String jobid)
     void deleteExpiredDisableTaskExecutorRequest(DisableTaskExecutorsRequest request) throws IOException;
 
     List<DisableTaskExecutorsRequest> loadAllDisableTaskExecutorsRequests(ClusterID clusterID) throws IOException;
+
+    void storeNewJobArtifact(JobArtifact jobArtifact) throws IOException;
```
sundargates, line 190: nit: addNewJobArtifact

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/MantisStorageProvider.java
```diff
@@ -183,4 +186,12 @@ List<MantisWorkerMetadataWritable> getArchivedWorkers(String jobid)
...
+
+    boolean jobArtifactExists(ArtifactID artifactID);
+
+    List<String> searchJobArtifactProjects(@Nullable String prefix) throws IOException;
```
sundargates, line 194: I'm guessing you want to list artifacts by name here?

sundargates, line 194: Add a comment on what the return value is. I would assume that this returns the artifact name but make it clear.

mantis-control-plane/mantis-control-plane-server/dependencies.lock
fdc-ntflx: nice catch. interesting... not sure what happened. let me revert

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/domain/JobArtifact.java
fdc-ntflx: agreed.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/MantisStorageProvider.java
fdc-ntflx: yes. I can better name that.

fdc-ntflx: +1.

sundargates (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/Jackson.java
fdc-ntflx: I think we have the following options:
- leave as is but we risk to make unwanted changes to other ser/deser of timestamps.
- use separate object mapper which registers `JavaTimeModule` 
- Change `Instant` to long

I prefer option 2, thoughts?

sundargates: The JsonSerializer already registers JavaTimeModule, I think. So I would prefer we use that. I'm trying to avoid new ObjectMapper initialization precisely because of this reason.

PR #285, opened 2022-11-03
Combines the two storage related interfaces IMantisStorageProvider and MantisStorageProvider

hmitnflx:
### Context
There are two interfaces IMantisStorageProvider and MantisStorageProvider in mantis codebase. This is a source of confusion since the implementation for the first one is effectively a data transformation from one instance type to another and the second one implements the storage specific nuance. 
I think it'd be easier to consolidate the two implementations that assumes a key-value implementation. That way new storages could be added easily!

### Checklist


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/IMantisStorageProvider.java
```diff
@@ -27,6 +27,7 @@
 import io.mantisrx.server.master.resourcecluster.TaskExecutorID;
 import io.mantisrx.server.master.resourcecluster.TaskExecutorRegistration;
 import java.io.IOException;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import rx.Observable;
```
calvin681, line 33: I think this is a very nice clean up. I suggest adding details javadoc comment here and in MantisStorageProvider explaining the differences between these 2 interfaces. For me, that has been a source of confusion.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/ExecuteStageRequest.java
```diff
@@ -62,9 +62,9 @@ public class ExecuteStageRequest implements Serializable {
     private final int stage;
     private final int totalNumStages;
     private final int metricsPort;
-    private final List<Integer> ports = new LinkedList<Integer>();
+    private final List<Integer> ports = new LinkedList<>();
```
sundargates, line 65: nit: Can we change this to an ArrayList?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/MantisStorageProvider.java
```diff
@@ -16,171 +16,86 @@
...
-import io.mantisrx.server.master.resourcecluster.ClusterID;
-import io.mantisrx.server.master.resourcecluster.TaskExecutorID;
-import io.mantisrx.server.master.resourcecluster.TaskExecutorRegistration;
 import java.io.IOException;
 import java.util.List;
-import rx.Observable;
+import java.util.Map;
 
 
 public interface MantisStorageProvider {
```
sundargates, line 24: A couple of questions:
1. Do we need this abstraction since this looks like it assumes that the underlying data will get stored in a KV store? 
2. If we need it, should we rename this to something like KVStore or something like that?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/NoopStorageProvider.java
```diff
@@ -16,150 +16,49 @@
...
-import io.mantisrx.server.master.resourcecluster.TaskExecutorRegistration;
-import io.mantisrx.shaded.com.google.common.collect.ImmutableList;
-import java.io.IOException;
-import java.util.ArrayList;
 import java.util.List;
-import rx.Observable;
+import java.util.Map;
 
 
 public class NoopStorageProvider implements MantisStorageProvider {
```
sundargates, line 23: I would maybe define this in the MantisStorageProvider itself and then have a static method MantisStorageProvider::noop() to get this instance.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/MantisJobStore.java
```diff
@@ -318,21 +253,14 @@ public Optional<IMantisWorkerMetadata> getArchivedWorker(String jobId, int worke
...
     public List<IMantisWorkerMetadata> getArchivedWorkers(String jobId) throws Exception {
-        if (logger.isTraceEnabled()) { logger.trace("Getting Archived workers for Job {}", jobId);}
-        List archivedWorkers = new ArrayList<>(archivedWorkersCache.getArchivedWorkerMap(jobId).values());
-        if (logger.isTraceEnabled()) {
-            logger.trace("Fetched archived {} workers for Job  {}", archivedWorkers.size(), jobId);
-        }
-        return archivedWorkers;
-
-
+        return new ArrayList<>(archivedWorkersCache.getArchivedWorkerMap(jobId).values());
```
sundargates, line 263: nit: use ImmutableList instead.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/MantisJobStore.java
```diff
@@ -275,34 +216,28 @@ private void archiveWorkersIfAny(IMantisJobMetadata mjmd) throws IOException {
     }
 
     public void archiveWorker(IMantisWorkerMetadata worker) throws IOException {
-        if (logger.isTraceEnabled()) { logger.trace("Archiving worker index=" + worker.getWorkerIndex()); }
```
sundargates, line 278: nit: Do you want to retain the logger.trace after removing the condition? maybe replace the string concatenation with interpolation for perf reasons.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/MantisStorageProviderAdapter.java
```diff
@@ -50,6 +66,27 @@
...
+    private static final String ACTIVE_ASGS_NS = "MantisActiveASGs";
+    private static final String TASK_EXECUTOR_REGISTRATION = "TaskExecutorRegistration";
+    private static final String DISABLE_TASK_EXECUTOR_REQUESTS = "MantisDisableTaskExecutorRequests";
+    private static final String CONTROLPLANE_NS = "mantis_controlplane";
+    private static final String NAMED_JOB_SECONDARY_KEY = "jobNameInfo";
+    private static final int WORKER_BATCH_SIZE = 1000;
+    private static final int WORKER_MAX_INDEX = 30000;
+
+    static {
+        mapper
```
sundargates, line 85: nit: use JsonSerializer from mantis-common-serde.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/MantisStorageProviderAdapter.java
```diff
@@ -50,6 +66,27 @@
 public class MantisStorageProviderAdapter implements IMantisStorageProvider {
```
sundargates, line 66: This looks like KVDal specific implementation. I would maybe rename this to KVDalAwareMantisStorageProvider

sundargates, line 66: What do you think about making this an abstract class and getting rid of the MantisStorageProvider interface (I'm referring to the kvdal methods such as get, upsert, etc...) methods and moving them here?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/IMantisStorageProvider.java
sundargates: +1

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/MantisStorageProviderAdapter.java
hmitnflx: I think that works better. In fact, I attempted down this road but it becomes a slightly bigger refactor in how the instance of this storage class is initialized here in a hard-coded whereas we create the instance of underlying `MantisStorageProvider` using reflection in a config driven manner.

How about I take a stab at it in a subsequent pass?

hmitnflx: done

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/MantisStorageProviderAdapter.java
hmitnflx: The serializer here depends on `DateTimeModule` in jackson. I'd avoid adding it to JsonSerializer, actually eliminate from here but would like to understand better on why it's needed.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/MantisJobStore.java
hmitnflx: are trace logs useful?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/MantisJobStore.java
hmitnflx: done

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/NoopStorageProvider.java
hmitnflx: done, good suggestion.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/MantisStorageProvider.java
hmitnflx: done, good suggestion.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/IMantisStorageProvider.java
calvin681: Can you add java doc?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/MantisStorageProvider.java
calvin681: so are you deleting this class?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/IMantisStorageProvider.java
calvin681: I suggestion I have would be to rename classes in this package XXXPersistenceProvider, and classes in the other package XXXStoreProvider. Drop the use of 'Storage'. Also This is a good opportunity to have consistent naming. So for interfaces, you can rename to `IKeyValueStoreProvider`.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/NoopStorageProvider.java
calvin681: Not used anymore right?

hmitnflx: ah yes, seems to have stuck around because of some recent rebase. thx!

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/IMantisPersistenceProvider.java
```diff
@@ -21,18 +21,25 @@
...
 import rx.Observable;
 
 
-public interface IMantisStorageProvider {
+/**
+ * A way to persist mantis master related metadata to a durable storage.
+ * See {@link KeyValueBasedPersistenceProvider} for how mantis job cluster,
+ * mantis job info is persisted to a key-value based storage (like cassandra)
+ */
+public interface IMantisPersistenceProvider {
```
Andyz26, line 42: FYI @sundargates has been pushing for moving all mantis interfaces not to have the "I" prefix.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
```diff
@@ -0,0 +1,718 @@
...
+ * The instance of this class needs an actual key-value storage
+ * implementation (implements KeyValueStorageProvider) to function.
+ *
+ * Look at {@code io.mantisrx.server.master.store.KeyValueStorageProvider}
+ * to see the features needed from the storage.
+ *   Effectively, an apache-cassandra like storage with primary key made
+ *   up of partition key and composite key and a way to iterate over all
+ *   partition keys in the table.
+ */
+public class KeyValueBasedPersistenceProvider implements IMantisPersistenceProvider {
```
Andyz26, line 87: Probably not for this PR but in the long term shall we consider breaking down these functionalities to each module (otherwise any new persistence interactions will be needed here and this class will grow to an unbounded size)? e.g. each logical component (jobCluster/jobArtifiact/job) take its own logic and only need to shared the keyValueStore instance.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/FileBasedPersistenceProvider.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public FileBasedPersistenceProvider(FileBasedStore sprovider) {
+        this(sprovider, noopEventPublisher);
+    }
+
+    public FileBasedPersistenceProvider(File stateDirectory, boolean eventPublisher) {
+        this(new FileBasedStore(stateDirectory), noopEventPublisher);
+    }
+
+    Optional<IJobClusterMetadata> loadJobCluster(String clusterName) throws IOException {
```
Andyz26, line 73: is this an override?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
Andyz26, line 87: Should this class be abstract?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/FileBasedPersistenceProvider.java
hmitnflx: No, this is a class used only in tests and had a completely separate implementation earlier. 
There were these two helpful methods defined used only in tests. I chose to keep them and simplify the implementation a bit.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/IMantisPersistenceProvider.java
hmitnflx: I'd like to take care of it in a separate PR.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
hmitnflx: no this is a concrete implementation. 
Another way of writing this code is to make this class abstract and move the interface methods from KeyValueStore here but then that'd lead to more bloating as all persistence use-cases would have to end up here.

hmitnflx: +1 this makes a lot of sense. 
there are 3 categories that I see so far - job metadata, task-compute-resource-management, and artifact data. There'll be more in the future.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/FileBasedPersistenceProvider.java
Andyz26: nit: add a doc or testing annotation.

Andyz26 (approved):

PR #291, opened 2022-11-11
Make pr comments sticky

fdc-ntflx:
### Context

### Checklist

### DIFF
--- a/.github/workflows/nebula-snapshot.yml
+++ b/.github/workflows/nebula-snapshot.yml
@@ -51,5 +51,6 @@ jobs:
             ```
             ${{ env.PR_STR }}
             ```
+          message-id: ${{ github.event.number }} # For sticky messages
           repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/nebula-snapshot.yml
```diff
@@ -51,5 +51,6 @@ jobs:
             ```
             ${{ env.PR_STR }}
             ```
+          message-id: ${{ github.event.number }} # For sticky messages
```
sundargates, line 54: What is the event number? Will this be the same for a given PR?

sundargates (approved):

.github/workflows/nebula-snapshot.yml
fdc-ntflx: I'm pretty confident that this is the number of a given PR (in this case, this pr will have number 291). I deduce this from - https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull-request-event-pull_request
- https://github.com/actions/checkout/issues/58

PR #293, opened 2022-11-18
Fix multiple HA sevice instance

Andyz26:
### Context

There are multiple places creating HA service instances and their own curator instance. A regression was introduced when some of the HA service creations no longer initialize/run their own curator service. 
The fix here is to consolidate and use a single HA service instance across all components and ensure the singleton instance will be ran.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/HighAvailabilityServicesUtil.java
+++ b/mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/HighAvailabilityServicesUtil.java
@@ -27,6 +27,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import lombok.extern.slf4j.Slf4j;
 import rx.Scheduler;
 import rx.Subscription;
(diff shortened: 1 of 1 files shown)


mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/HighAvailabilityServicesUtil.java
```diff
@@ -37,12 +38,17 @@
  */
 @Slf4j
 public class HighAvailabilityServicesUtil {
+  private static AtomicReference<HighAvailabilityServices> HAServiceInstanceRef = new AtomicReference<>();
```
hmitnflx, line 41: nit: you could make it final

hmitnflx (approved):

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/HighAvailabilityServicesUtil.java
```diff
@@ -37,12 +38,17 @@
  */
 @Slf4j
 public class HighAvailabilityServicesUtil {
+  private final static AtomicReference<HighAvailabilityServices> HAServiceInstanceRef = new AtomicReference<>();
```
sundargates, line 41: Should we move this to some *Registry class?

sundargates (approved):

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/HighAvailabilityServicesUtil.java
Andyz26: I feel like ideally we should avoid have leaf modules depending on HA service in general but inject the master client properly instead.

PR #296, opened 2022-11-23
Improve fetch performance for completed jobs

hmitnflx:
### Context
In an earlier diff, I think I slowed down loading all completed jobs causing timeouts during JobClusterActor init. This prevents master from launching if that job list is too big.
Speeding it up by using a different all rows API.
On closer inspection, I feel that completed jobs shouldn't be part of the JobClusterActor init and can be delayed—-ideally done async. I'll spend some time on that time next!

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
@@ -722,7 +722,7 @@ public class JobClusterActor extends AbstractActorWithTimers implements IJobClus
                     String.format("JobCluster %s initialized successfully. But is currently disabled",
                             initReq.jobClusterDefinition.getName()),initReq.jobClusterDefinition.getName(),
                     initReq.requestor), getSelf());
-            logger.info("Job expiry check frquency set to {}", expireFrequency);
+            logger.info("Job expiry check frequency set to {}", expireFrequency);
             setExpiredJobsTimer(expireFrequency);
 
             getContext().become(disabledBehavior);
@@ -767,13 +767,13 @@ public class JobClusterActor extends AbstractActorWithTimers implements IJobClus
             try {
                 cronManager = new CronManager(name, getSelf(), jobClusterMetadata.getJobClusterDefinition().getSLA());
             } catch (Exception e) {
-                logger.warn("Exception initializing cron {}", e);
+                logger.warn("Exception initializing cron", e);
             }
             initRunningJobs(initReq, sender);
 
             setExpiredJobsTimer(expireFrequency);
 
-            logger.info("Job expiry check frquency set to {}", expireFrequency);
+            logger.info("Job expiry check frequency set to {}", expireFrequency);
             try {
                 jobManager.addCompletedJobsToCache(initReq.completedJobsList);
             } catch(Exception e) {
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
@@ -711,7 +711,7 @@ public class JobClusterActor extends AbstractActorWithTimers implements IJobClus
                             jobStore.archiveJob(jobMeta);
                         }
                     } catch (Exception e) {
-                        logger.error("Exception {} archiving job {} during init ",e.getMessage(), jobMeta.getJobId());
+                        logger.error("Exception {} archiving job {} during init ",e.getMessage(), jobMeta.getJobId(), e);
                     }
                 }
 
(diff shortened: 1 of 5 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
```diff
@@ -421,23 +421,23 @@ public List<IJobClusterMetadata> loadAllJobClusters() throws IOException {
...
+            .values().stream()
+            .flatMap(x -> x.values().stream())
+            .map(data -> {
+                try {
+                    NamedJob.CompletedJob cj = mapper.readValue(data, NamedJob.CompletedJob.class);
+                    CompletedJob completedJob = DataFormatAdapter.convertNamedJobCompletedJobToCompletedJob(cj);
+                    successCount.getAndIncrement();
+                    return completedJob;
+                } catch (JsonProcessingException e) {
+                    logger.warn("failed to parse CompletedJob from {}", data, e);
```
Andyz26, line 434: Do we expect this to happen in any case? if this error does happen does it imply some regression and would it be better to fail loudly instead?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
```diff
@@ -421,23 +421,23 @@ public List<IJobClusterMetadata> loadAllJobClusters() throws IOException {
...
-                        CompletedJob completedJob = DataFormatAdapter.convertNamedJobCompletedJobToCompletedJob(cj);
-                        completedJobsList.add(completedJob);
-                        successCount.getAndIncrement();
-                    } catch (JsonProcessingException e) {
-                        logger.warn("failed to parse CompletedJob from {} {}", pkey, data, e);
-                        failedCount.getAndIncrement();
-                    }
-                });
-        }
+        final List<CompletedJob> completedJobsList = kvStore.getAllRows(NAMED_COMPLETEDJOBS_NS)
```
fdc-ntflx, line 424: I noticed that `getAllRows` still uses `getAllPartitionKeys`. is there any major performance difference between the old and new code? Also is the docstring on `getAllPartitionKeys` still valid? https://github.com/Netflix/mantis/blob/03685037d1a3903068054dcf4d9526e6a2fb8b4b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/store/KeyValueStore.java#L56

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
hmitnflx: it implies some regression but since completed jobs aren't critical for functionality I'm okay with this as a warning instead.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
hmitnflx: That's the default implementation but the inheritors can choose to override the implementation. 
Also, this change wouldn't make this function call fast enough, with the current implementation and table design. I'm disabling it since the older implementation wasn't doing anything either.

fdc-ntflx (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -3241,7 +3241,7 @@ public void purgeOldCompletedJobs(long tooOldCutOff, MantisJobStore jobStore) {
                         }
 
                     } catch (Exception e) {
-                        logger.warn("Unable to purge job {} due to {}", completedJob, e.getMessage());
+                        logger.warn("Unable to purge job {} due to {}", completedJob, e);
```
sundargates, line 3244: @hmitnflx Looks like we need both e.getMessage() and e here?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -3275,7 +3275,7 @@ void forcePurgeCompletedJobs(MantisJobStore jobStore) {
                             labelsCache.removeJobIdFromLabelCache(jobId.get());
                         }
                     } catch (Exception e) {
-                        logger.warn("Unable to purge job {} due to {}", completedJob, e.getMessage());
+                        logger.warn("Unable to purge job {} due to {}", completedJob, e);
```
sundargates, line 3278: Same here.

hmitnflx: ah yes

PR #311, opened 2022-12-10
I am making close() implementation optional. 

sundargates:
### Context

Making close() implementation optional. The reason it's made optional is that we want users to migrate to Titus without any API changes. I will change this in the 3.0.0 release so that we can close sources and sinks appropriately.

### Checklist


hmitnflx (approved):
nit: could you please add some context/motivation for this change? Also, super nitty, I'd suggest making the commit msg header passive-voice like 'make close() optional'.

calvin681 (approved):

PR #314, opened 2022-12-16
[Minor] Fix npe in metrics collector

Andyz26:
### Context

Fix NPE error event reported from control plane when workerId is not available in metrics collector.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/events/WorkerMetricsCollector.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/events/WorkerMetricsCollector.java
@@ -120,7 +120,11 @@ public class WorkerMetricsCollector extends AbstractScheduledService implements
             final WorkerId workerId = workerStatusEvent.getWorkerId();
             Preconditions.checkNotNull(jobWorkers.get(jobId));
             final IMantisWorkerMetadata metadata = jobWorkers.get(jobId).get(workerId);
-            Preconditions.checkNotNull(metadata);
+            if (metadata == null) {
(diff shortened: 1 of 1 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/events/WorkerMetricsCollector.java
```diff
@@ -120,7 +120,11 @@ public void process(WorkerStatusEvent workerStatusEvent) {
             final WorkerId workerId = workerStatusEvent.getWorkerId();
             Preconditions.checkNotNull(jobWorkers.get(jobId));
             final IMantisWorkerMetadata metadata = jobWorkers.get(jobId).get(workerId);
-            Preconditions.checkNotNull(metadata);
+            if (metadata == null) {
+                log.warn("Unknown workerId: {} for metrics collector in job: {}", workerId, jobId);
```
sundargates, line 124: In what situation will this be null? Can you raise it to the error level?

sundargates (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/events/WorkerMetricsCollector.java
Andyz26: When some workers are in-transition and being replace this can happen. And npe gets wired with radar auto alerts (email notification) when this happens.

Andyz26: There are other logs in control plane to track those cases already iirc.

PR #315, opened 2022-12-16
IcebergWriter uses table.io() to replace HadoopOutputFile

liuml07:
### Context

Currently we use HadoopOutputFile interface to open data file to write. This limits the Iceberg data file has to use a Hadoop file system to open for write. As Iceberg FileIO is more flexible than that, we should use Iceberg native IO interface for opening files. For example, in Netflix set up, we use `S3FileIO` implementation to write to secure tables.

### Checklist


liuml07:
CC: @jfz @sundargates @hmit for review

jfz (approved):

hmitnflx (approved):

PR #316, opened 2022-12-17
Fix metrics in non-mesos job

Andyz26:
### Context

* MetricsServer uses metric group ID as measurement name, which doesn't work when common tags are set.

* In the non-Mesos job (task executor initiated jobs) the current metrics publisher is no-op but the metrics publisher instance created from CoreConfiguration causes loader trouble in the multi-loader (Titus + SB) case. I am disabling the no-op publisher for now and will follow up to see what the proper override/injection is for this case.

### Checklist


Andyz26:
@sundargates let me know if you got any comments and I will address them in a follow-up. Need to merge this one to unblock nf side fix.

PR #317, opened 2022-12-21
[Preview] A dsl for writing new mantis jobs

hmitnflx:
### Context

### Checklist

- [ ] Write more tests


mantis-runtime/src/main/java/io/mantisrx/runtime/core/MantisStream.java
```diff
@@ -0,0 +1,104 @@
...
+import io.mantisrx.runtime.core.sources.SourceFunction;
+import io.mantisrx.runtime.parameter.ParameterDefinition;
+import java.time.Duration;
+import lombok.Getter;
+
+public interface MantisStream<T> {
+
+    <OUT> MantisStream<OUT> source(SourceFunction<OUT> sourceFunction);
+
+    MantisStream<Void> sink(SinkFunction<T> sinkFunction);
```
sundargates, line 34: What's the reason that this is producing an output? Shouldn't this be a void function?

For instance, I'm guessing you want users to avoid doing things like
```
MantisStream.source().sink().sink() etc...?
```

mantis-runtime/src/main/java/io/mantisrx/runtime/core/MantisStream.java
```diff
@@ -0,0 +1,104 @@
...
+
+    <OUT> MantisStream<OUT> create();
```
sundargates, line 36: What does this do? Maybe add some doc here since most users will only be looking at this interface.

mantis-examples/mantis-examples-sine-function/src/main/java/io/mantisrx/mantis/examples/sinefunction/SineFunction.java
```diff
@@ -0,0 +1,91 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class SineFunction {
+    public static void main(String[] args) {
+        final double amplitude = 5.0;
+        final double frequency = 1;
+        final double phase = 0.0;
+
+        MantisStreamImpl.init()
```
sundargates, line 44: IMO, the chain below looks cleaner. Also, not very sure if we should mix MantisStreams with the parameter definition itself. IMO, it should be part of the job definition for the time being. We should think about getting rid of it in the future. 
```
MantisStreamFactory
  .source(SourceFunction)
  .filter()
  .map()
  .keyBy()
  .tumblingWindow()
  .reduce()
  .sink() 
```

liuml07: yeah, agree decoupling the parameter definition from the stream building is cleaner.

mantis-runtime/src/main/java/io/mantisrx/runtime/core/MantisStream.java
```diff
@@ -0,0 +1,104 @@
...
+    <OUT> MantisStream<OUT> map(MapFunction<T, OUT> mapFn);
+    <OUT> MantisStream<OUT> flatMap(FlatMapFunction<T, OUT> flatMapFn);
+
+    MantisStream<T> materialize();
+
+    <K> KeyedMantisStream<K, T> keyBy(KeyByFunction<K, T> keyFn);
+
+    MantisStream<T> parameters(ParameterDefinition<?>... params);
+
+    void execute();
```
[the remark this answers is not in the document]
liuml07: If there are multiple sinks, one `sink()` call will need to append all of them?

mantis-runtime/src/main/java/io/mantisrx/runtime/core/MantisStream.java
```diff
@@ -0,0 +1,104 @@
...
+
+    interface KeyedMantisStream<K, IN> {
```
sundargates, line 49: Should this be a top-level interface instead? This way, people can avoid referring to these symbols like below. 
```
MantisStream.KeyedMantisStream<>
```

mantis-runtime/src/main/java/io/mantisrx/runtime/core/MantisStream.java
```diff
@@ -0,0 +1,104 @@
...
+        <OUT> KeyedMantisStream<K, OUT> map(MapFunction<IN, OUT> mapFn);
+
+        <OUT> KeyedMantisStream<K, OUT> flatMap(FlatMapFunction<IN, OUT> flatMapFn);
+
+        KeyedMantisStream<K, IN> filter(FilterFunction<IN> filterFn);
+
+        KeyedMantisStream<K, IN> window(MantisStream.WindowSpec spec);
```
sundargates, line 56: Can we prefix the operator name with the type of the window e.g., tumbingWindow?

mantis-runtime/src/main/java/io/mantisrx/runtime/core/MantisStream.java
```diff
@@ -0,0 +1,104 @@
...
+
+        <OUT> MantisStream<OUT> reduce(ReduceFunction<IN, OUT> reduceFn);
+    }
+
+    @Getter
+    class WindowSpec {
```
sundargates, line 62: This should be a top-level class, too, IMO.

Andyz26:
Could you add an OSS version doc about the plan/structure and link it here?

mantis-common/src/main/java/io/mantisrx/common/codec/Codecs.java
```diff
@@ -109,4 +116,31 @@ public byte[] encode(final byte[] value) {
             }
         };
     }
+
+    public static <T> Codec<T> javaSerializer() {
```
calvin681, line 120: do you need to specify T extends Serializable?

mantis-runtime/src/main/java/io/mantisrx/runtime/core/MantisStream.java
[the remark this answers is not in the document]
calvin681: I don't think we should perform execute as part of sink. I think we need to go further to break up the execute, we will need a build method that return a Job object. And execute should reuse old logic of either executing locally or as stages.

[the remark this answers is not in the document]
liuml07: I agree separating the "job graph" being built and being executed is more flexible. In future supporting multiple sinks will make it less straightforward if one `.sink()` triggers execution.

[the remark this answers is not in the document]
sundargates: I agree with decoupling the building of the job graph from executing/submitting it for execution.

mantis-runtime/src/main/java/io/mantisrx/runtime/core/MantisStream.java
liuml07: I think either way works. In Flink the window operator is called `window` and the "Window Spec" can be a specific window definition.

hmitnflx:
I have addressed core feedback and plan to merge this to proceed with further testing. I would address any new feedback in a subsequent PR since this change is pretty isolated and wouldn't affect production...

PR #319, opened 2023-01-02
[WIP] Make it easy to integrate with another distributed KV store like etcd for leader election

sundargates:
#### The diff still needs to be ready for review. I'm still iterating on it. I hope to wrap it up as soon as possible. 

This diff has two general purposes.
1. It takes a stab at replacing "skife.config" with https://github.com/lightbend/config (a much better config, more-used library).
2. It also takes a stab at replacing the zookeeper with another distributed kv store, such as etcd, for leader election/leader retrieval. 

I'd like to divide this PR into two smaller PRs so that it's easier to both review it as well as merge it. 

### Architecture

On the integration of a highly-available KV store for leader election/retrieval, these are the classes to look at. 
1. HighAvailabilityServices: This class is a factory for creating a leader elector service or a leader retrieval service.
2. LeaderElectorService: This service allows a contender to participate in the leader election and notifies the contender of state changes.
3. LeaderRetrievalService: This service allows a spectator to listen to leader election results. 

I have also added a zookeeper-specific integration for HighAvailabilityServices called ZookeeperHighAvailabilityServicesFactory in `mantis-control-plane-core`. Any new integration can be added by implementing HighAvailabilityServicesFactory and providing that className as the config value for `mantis.highAvailability.mode=class` and `mantis.highAvailability.class` set to the class name. 

#### Why not use Flink's Configuration library?
There are a few reasons why I have chosen the typesafe library over Flink's configuration classes.
1. Flink's configuration code is not a general-purpose configuration library. Instead, it's tied to flink and its internals. So, for instance, it's coupled with Flink's configurations, such as pipeline options. 
5. Typesafe config library, on the other hand, is a general-purpose config library used in several systems, including Akka. It has a reasonably small footprint with 0 dependencies and is decoupled from the runtime, such as guice and spring-boot. Based on these reasons, it's an excellent long-term choice. 

### Checklist


mantis-common/src/main/java/io/mantisrx/common/metrics/MetricsPublisherUtil.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package io.mantisrx.common.metrics;
+
+import com.typesafe.config.Config;
+
+public class MetricsPublisherUtil {
+
+    public static MetricsPublisher createMetricsPublisher(Config config) throws Exception {
+        final String clazz = config.getString("mantis.metricsPublisher.class");
```
calvin681, line 24: is this an existing property?

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/ClientServices.java
```diff
@@ -18,7 +18,7 @@
 import io.mantisrx.server.core.master.MasterMonitor;
 import io.mantisrx.server.master.resourcecluster.ClusterID;
 import io.mantisrx.server.master.resourcecluster.ResourceClusterGateway;
-import io.mantisrx.shaded.com.google.common.util.concurrent.Service;
+import java.io.Closeable;
 
 /**
  * HighAvailabilityServices is a container for a group of services which are considered to be highly available because
```
calvin681, line 24: Comments no longer match. Why the need to rename this to ClientServices?

calvin681: NVM, I see the refactor.

calvin681 (approved):

PR #321, opened 2023-01-14
Schedule TE in min sku

Andyz26:
### Context

Currently, the TE allocation is greedy and uses the first TE with a machine size big enough for the request. Switching the allocation logic to use the min size available for the request instead.

### Checklist


Andyz26:
@sundargates I will merge for rc for now. Feel free to comment and I will follow-up.

PR #323, opened 2023-01-24
Verifying if some constraints on the write stage path hold true

sundargates:
### Context

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
@@ -42,6 +42,7 @@ import io.mantisrx.shaded.com.fasterxml.jackson.databind.DeserializationFeature;
 import io.mantisrx.shaded.com.fasterxml.jackson.databind.ObjectMapper;
 import io.mantisrx.shaded.com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import io.mantisrx.shaded.com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import io.mantisrx.shaded.com.google.common.base.Preconditions;
 import io.mantisrx.shaded.com.google.common.collect.ImmutableList;
 import io.mantisrx.shaded.com.google.common.collect.Lists;
 import java.io.IOException;
@@ -53,6 +54,7 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


fdc-ntflx (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
```diff
@@ -2269,7 +2274,7 @@ private void resubmitWorker(JobWorker oldWorker) throws Exception {
         @Override
         public int scaleStage(MantisStageMetadataImpl stageMetaData, int numWorkers, String reason) {
             LOGGER.info("Scaling stage {} to {} workers", stageMetaData.getStageNum(), numWorkers);
-            final int oldNumWorkers = stageMetaData.getNumWorkers();
+            final int oldNumWorkers = stageMetaData.getAllWorkers().size();
```
Andyz26, line 2277: Can you check when the worker list has out of sync workers (e.g. worker index 5 when numWorker is 3) such workers will get closed?

Andyz26 (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
calvin681, line 2277: Is this safe? There is a valid case where if scaleStage is called multiple times quickly, the worker list is getting updated. So worker list size cannot be used to determine the difference.

calvin681: There is a way to avoid changing this line. Instead of scaling to stageMeta.getNumWorkers, go through the worker list, find the highest worker index from that list, and scale to that.

calvin681 (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
sundargates: scaleStage being multiple times quickly is not a concern because all the interactions with WorkerManager::scaleStage happen from within the JobActor. Furthermore, Akka guarantees that messages will be handled strictly in serial order. 

I'm afraid I have to disagree with the second idea here. Imagine the following scenario:
1. scaleDown called for the stage from 6 workers to 3 workers.
2. Let's say worker six got removed because workers are removed from 6 to 3 in that order.
3. Now, the machine goes down.
4. We will come back up and realize that three is not equal to 5 (number of active workers).
5. We will scale back to 5 instead of 3, which was the original intention. We are also not scaling back to the original state, which had six workers. 

Because of this, I prefer the current logic that realizes the intent and makes that a reality.

hmitnflx (approved):

Andyz26:
@sundargates do you still plan to merge it?

sundargates:
No this is not needed.

PR #331, opened 2023-02-06
Setting preconditions to ensure we don't start mantis with an invalid state

sundargates:
### Context

Currently, we ignore duplicate workers for the same (jobId/stageNum/workerIdx) tuple. However, when we ignore these duplicate workers, we are essentially running a mantis master with a state that may not reflect the previous state of the master. This diff aims to surface these issues and block deployments to make sure we at least know when such problems happen. The current thinking is we should never see these issues. But, if we do see them, then the expectation is that we run state cleaner to clean up the state and then restart the master.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/MantisJobMetadataImpl.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/MantisJobMetadataImpl.java
@@ -240,18 +240,19 @@ public class MantisJobMetadataImpl implements IMantisJobMetadata {
      public boolean addWorkerMetadata(int stageNum, JobWorker newWorker)
             throws InvalidJobException {
          if(logger.isTraceEnabled()) { logger.trace("Adding workerMetadata {} for stage {}", stageNum, newWorker); }
-        boolean result=true;
         if(!stageMetadataMap.containsKey(stageNum)) {
         	logger.warn("No such stage {}", stageNum);
         }
-        if(!((MantisStageMetadataImpl)stageMetadataMap.get(stageNum)).addWorkerIndex(newWorker))
-            result=false;
-        Integer integer = workerNumberToStageMap.put(newWorker.getMetadata().getWorkerNumber(), stageNum);
-        if(integer != null && integer != stageNum) {
-            logger.error(String.format("Unexpected to put worker number mapping from %d to stage %d for job %s, prev mapping to stage %d",
-            		newWorker.getMetadata().getWorkerNumber(), stageNum, newWorker.getMetadata().getJobId(), integer));
+        final boolean result = ((MantisStageMetadataImpl) stageMetadataMap.get(stageNum)).addWorkerIndex(newWorker);
+
+        if (result) {
+            Integer integer = workerNumberToStageMap.put(newWorker.getMetadata().getWorkerNumber(), stageNum);
+            if(integer != null && integer != stageNum) {
+                logger.error(String.format("Unexpected to put worker number mapping from %d to stage %d for job %s, prev mapping to stage %d",
+                    newWorker.getMetadata().getWorkerNumber(), stageNum, newWorker.getMetadata().getJobId(), integer));
+            }
+            if(logger.isTraceEnabled()) { logger.trace("Exit addworkerMeta {}", workerNumberToStageMap); }
         }
-         if(logger.isTraceEnabled()) { logger.trace("Exit addworkerMeta {}", workerNumberToStageMap); }
         return result;
     }
 
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
@@ -42,6 +42,7 @@ import io.mantisrx.shaded.com.fasterxml.jackson.databind.DeserializationFeature;
 import io.mantisrx.shaded.com.fasterxml.jackson.databind.ObjectMapper;
 import io.mantisrx.shaded.com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import io.mantisrx.shaded.com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import io.mantisrx.shaded.com.google.common.base.Preconditions;
 import io.mantisrx.shaded.com.google.common.collect.ImmutableList;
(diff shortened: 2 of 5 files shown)


calvin681 (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/MantisJobMetadataImpl.java
```diff
@@ -240,18 +240,19 @@ boolean replaceWorkerMetaData(int stageNum,
...
-        if(!((MantisStageMetadataImpl)stageMetadataMap.get(stageNum)).addWorkerIndex(newWorker))
-            result=false;
-        Integer integer = workerNumberToStageMap.put(newWorker.getMetadata().getWorkerNumber(), stageNum);
-        if(integer != null && integer != stageNum) {
-            logger.error(String.format("Unexpected to put worker number mapping from %d to stage %d for job %s, prev mapping to stage %d",
-            		newWorker.getMetadata().getWorkerNumber(), stageNum, newWorker.getMetadata().getJobId(), integer));
+        final boolean result = ((MantisStageMetadataImpl) stageMetadataMap.get(stageNum)).addWorkerIndex(newWorker);
+
+        if (result) {
+            Integer integer = workerNumberToStageMap.put(newWorker.getMetadata().getWorkerNumber(), stageNum);
```
hmitnflx, line 249: This is a behavior change where we aren't updating the map when `result = false`. Would `workerNumberToStageMap` have the mapping for an existing worker esp. during leader election or bootstrap from state?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/MantisJobMetadataImpl.java
```diff
@@ -240,18 +240,19 @@ boolean replaceWorkerMetaData(int stageNum,
...
+            if(integer != null && integer != stageNum) {
+                logger.error(String.format("Unexpected to put worker number mapping from %d to stage %d for job %s, prev mapping to stage %d",
+                    newWorker.getMetadata().getWorkerNumber(), stageNum, newWorker.getMetadata().getJobId(), integer));
+            }
+            if(logger.isTraceEnabled()) { logger.trace("Exit addworkerMeta {}", workerNumberToStageMap); }
```
hmitnflx, line 254: I've been removing trace logging since I haven't turned on `trace` logging for any sort of debugging. What do you think? If useful, I'd also not remove them.

hmitnflx (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
```diff
@@ -561,7 +569,14 @@ public Optional<IMantisJobMetadata> loadArchivedJob(String jobId) throws IOExcep
                 for (IMantisWorkerMetadata w : archivedWorkers) {
                     try {
                         MantisWorkerMetadataWritable wmw = DataFormatAdapter.convertMantisWorkerMetadataToMantisWorkerMetadataWritable(w);
-                        jmw.addWorkerMedata(w.getStageNum(), wmw, null);
+                        Preconditions.checkState(
```
hmitnflx, line 572: I'd avoid throwing an exception for archived job (or even archived worker) since the job isn't already running.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
```diff
@@ -371,7 +372,14 @@ public List<IMantisJobMetadata> loadAllJobs() throws IOException {
...
+                    Preconditions.checkState(
+                        jobMeta.addWorkerMedata(workerMeta.getStageNum(), workerMeta, null),
+                        "JobID=%s stage=%d workerIdx=%d has existing worker, existing=%s, new=%s",
+                        workerMeta.getJobId(),
+                        workerMeta.getStageNum(),
+                        workerMeta.getWorkerIndex(),
+                        jobMeta.getWorkerByIndex(workerMeta.getStageNum(), workerMeta.getWorkerIndex()).getWorkerId(),
+                        workerMeta.getWorkerId());
                 }
                 jobMetas.add(DataFormatAdapter.convertMantisJobWriteableToMantisJobMetadata(jobMeta, eventPublisher));
```
hmitnflx, line 384: Another thing worth logging (or even exception) is non-contiguous utilization of `workerIdx`. Mantis doesn't handle gaps very well and so proactively logging it when adding job to jobMetas would be a good idea.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/MantisJobMetadataImpl.java
sundargates: Yes, this is behavior change. But the thinking here is that this is more correct than the existing logic because if you cannot add the worker to the stage, then the workerNumberToStageMap should also reflect that.

sundargates: Let me know if the rationale here makes sense.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/MantisJobMetadataImpl.java
sundargates: I did not add this. These are existing logs.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
sundargates: This is a separate problem, and I will tackle this separately. But, for the time being, the script I have catches this issue.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
sundargates: I got rid of this code fully. WIll it have any impact?

PR #334, opened 2023-02-20
Upgrade to Mockito 3

### LINKED ISSUE #184: Upgrade code base to use 'Mockito' version 3.+
### Context

Upgrade the whole repo to use Mockito version 3+.
Currently, there are breaks in the existing code base when applying Mockito 3+.

HoussemNasri:
CLoses #184

### Checklist


sundargates (approved):
LGTM! Thanks for making this change.

sundargates:
If we have gotten rid of the old mockito dependency in every module, can we also get rid of the dependency in the root build.gradle file?

HoussemNasri:
Thank you for the review! I removed all references to mockitoCore2 and mockitoAll.

image

PR #341, opened 2023-02-25
[Back Compat] Wrap env parameter read + metrics tags

Andyz26:
### Context
* Make props read from system env wrapped with system properties read since these props are now in system props in TE.
* Add worker number + stage number tag names to support existing metrics dashboard/alerts.

### Checklist

### DIFF
--- a/mantis-runtime/src/main/java/io/mantisrx/runtime/executor/StageExecutors.java
+++ b/mantis-runtime/src/main/java/io/mantisrx/runtime/executor/StageExecutors.java
@@ -320,28 +320,18 @@ public class StageExecutors {
      *
      * @return
      */
-    private static int resolveStageConcurrency(int givenStageConcurrency) {
+    private static int resolveStageConcurrency(Context context, int givenStageConcurrency) {
         if (givenStageConcurrency == StageConfig.DEFAULT_STAGE_CONCURRENCY) {
             String jobParamPrefix = "JOB_PARAM_";
             String stageConcurrencyParam = jobParamPrefix + STAGE_CONCURRENCY;
             // Need to be compatible for both Mesos and TaskExecutor.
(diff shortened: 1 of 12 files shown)


calvin681:
There are many more places where getenv is used. We need to update all of them.

https://github.com/Netflix/mantis/blob/5f18638305a75ab0e489a940069b93702a1f0283/mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoScaler.java#L133

mantis-runtime/src/main/java/io/mantisrx/runtime/executor/StageExecutors.java
```diff
@@ -313,7 +324,9 @@ private static int resolveStageConcurrency(int givenStageConcurrency) {
         if (givenStageConcurrency == StageConfig.DEFAULT_STAGE_CONCURRENCY) {
             String jobParamPrefix = "JOB_PARAM_";
             String stageConcurrencyParam = jobParamPrefix + STAGE_CONCURRENCY;
-            String concurrency = System.getenv(stageConcurrencyParam);
+            // Need to be compatible for both Mesos and TaskExecutor.
+            String concurrency = System.getProperty(stageConcurrencyParam, System.getenv(stageConcurrencyParam));
```
sundargates, line 328: I think we should avoid both System::getEnv as well as System::getProperty.

Andyz26: this might be tricky in some places where they don't have any injection/context to use. Any thought on best practice here?

sundargates: As discussed offline, let's make use of the Context class to get this value.

Andyz26: fixed (read via context now).

PR #342, opened 2023-02-25
mantis-api is migrated to mantis

Sahil06012002:
mantis-api is migrated to mantis with zero changes in any of the repositories.


sundargates:
Thank you for this PR. Instead of making mantis-api a submodule, we prefer to take the codebase from mantis-api along with its commit history and add it to this repo.

sundargates (changes requested):
The submodule approach may not be how we want to migrate mantis-api to mantis.

PR #362, opened 2023-03-08
Adding the capability to set vtts for iceberg tables written to by mantis

sundargates:
### Context

Some mantis customers need support for vtts in their iceberg tables. This diff adds support for that. This diff was done in collaboration with @liuml07. 

### Checklist


mantis-connectors/mantis-connector-iceberg/src/main/java/io/mantisrx/connector/iceberg/sink/committer/IcebergCommitter.java
```diff
@@ -16,36 +16,79 @@
...
+            config.getDatabase(),
+            config.getTable(),
+            dataFiles.size());
+
+        Long lowWatermark = getCurrentWatermark(transaction.table());
+        for (MantisDataFile flinkDataFile : dataFiles) {
+            lowWatermark = minNullSafe(lowWatermark, flinkDataFile.getLowWatermark());
+        }
+
+        if (lowWatermark != null) {
```
liuml07, line 73: I think we need to check the max of the current watermark and the new watermark.

mantis-connectors/mantis-connector-iceberg/src/main/java/io/mantisrx/connector/iceberg/sink/writer/config/WriterConfig.java
```diff
@@ -34,6 +34,8 @@ public class WriterConfig extends SinkConfig {
     private final int writerMaximumPoolSize;
     private final Configuration hadoopConfig;
 
+    private final boolean watermarkEnabled;
```
liuml07, line 37: wondering if this property can be removed entirely. Writers always calculate the low watermark and always populate to committer stage. User can simply provide null value (default) to disable this.

mantis-connectors/mantis-connector-iceberg/src/main/java/io/mantisrx/connector/iceberg/sink/committer/config/CommitterProperties.java
```diff
@@ -28,6 +28,7 @@ private CommitterProperties() {
      * Iceberg committer frequency by time in milliseconds.
      */
     public static final String COMMIT_FREQUENCY_MS = "commitFrequencyMs";
+    public static final String WATERMARK_PROPERTY_KEY = "watermarkPropertyKey";
```
liuml07, line 31: Do we ask user to set per-region config value for this key?

mantis-connectors/mantis-connector-iceberg/src/main/java/io/mantisrx/connector/iceberg/sink/committer/IcebergCommitter.java
sundargates: good point - done!

mantis-connectors/mantis-connector-iceberg/src/main/java/io/mantisrx/connector/iceberg/sink/committer/config/CommitterProperties.java
sundargates: We can set this at the job level.

mantis-connectors/mantis-connector-iceberg/src/main/java/io/mantisrx/connector/iceberg/sink/committer/IcebergCommitter.java
```diff
@@ -16,36 +16,82 @@
...
+        AppendFiles tableAppender = transaction.newAppend();
+        dataFiles.stream().map(MantisDataFile::getDataFile).forEach(tableAppender::appendFile);
         tableAppender.commit();
+        log.info(
+            "Iceberg committer {}.{} appended {} data files to transaction",
+            config.getDatabase(),
+            config.getTable(),
+            dataFiles.size());
+
+        Long currentWatermark = getCurrentWatermark(transaction.table());
```
liuml07, line 69: nit: simply use `table` should be fine?

mantis-connectors/mantis-connector-iceberg/src/main/java/io/mantisrx/connector/iceberg/sink/committer/IcebergCommitter.java
```diff
@@ -16,36 +16,82 @@
...
+
+
+        transaction.commitTransaction();
         return table.currentSnapshot() == null ? new HashMap<>() : new HashMap<>(table.currentSnapshot().summary());
     }
+
+    @Nullable
+    private Long getCurrentWatermark(Table table) {
+        try {
+            return Long.parseLong(table.properties().get(config.getWatermarkPropertyKey()));
```
liuml07, line 91: I'm thinking if we need to refresh table? Otherwise the `table.properties()` may always return the old value. I'm not sure about this though. A safer approach is to create a `lowWatermark` field in this committer class, initialize it once in the constructor, and keep updating it for every commit. Then we don't need this `getCurrentWatermark` method any more as it's saved and maintained in the Iceberg committer class.

mantis-connectors/mantis-connector-iceberg/src/main/java/io/mantisrx/connector/iceberg/sink/committer/config/CommitterProperties.java
liuml07: In Flink the three jobs in different regions sets region-specific VTTS table properties. If we want to keep the same behavior in Mantis, we may need to add `region` to the key name?

mantis-connectors/mantis-connector-iceberg/src/main/java/io/mantisrx/connector/iceberg/sink/writer/config/WriterConfig.java
liuml07: I think we can depend on the watermarkPropertyKey being null or not to check if this is enabled? So that user will not need to input this config value but it's kinda derived.

mantis-connectors/mantis-connector-iceberg/src/main/java/io/mantisrx/connector/iceberg/sink/committer/IcebergCommitter.java
sundargates: Since we are sending the table from the current transaction, it should have the latest set of properties.

liuml07 (approved):

PR #367, opened 2023-03-17
Configurable Disable TE expiry

Andyz26:
### Context

Make disable TE request expiration configurable from request.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v1/ResourceClustersNonLeaderRedirectRoute.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v1/ResourceClustersNonLeaderRedirectRoute.java
@@ -265,9 +265,7 @@ public class ResourceClustersNonLeaderRedirectRoute extends BaseRoute {
             log.info("POST /api/v1/resourceClusters/{}/disableTaskExecutors called with body {}", clusterID, request);
             return withFuture(gateway.getClusterFor(clusterID).disableTaskExecutorsFor(
                 request.getAttributes(),
-                request.getExpirationDurationInHours() == null || request.getExpirationDurationInHours() <= 0L ?
-                    Instant.now().plus(Duration.ofHours(24)) :
-                    Instant.now().plus(Duration.ofHours(request.getExpirationDurationInHours()))));
+                Instant.now().plus(Duration.ofHours(request.getExpirationDurationInHours()))));
         });
     }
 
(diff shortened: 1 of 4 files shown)


fdc-ntflx (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/DisableTaskExecutorsRequest.java
```diff
@@ -22,4 +22,6 @@
 @Value
 public class DisableTaskExecutorsRequest {
     Map<String, String> attributes;
+
+    Long expirationDurationInHours;
```
sundargates, line 26: Let's use Duration DS for this. Again, I want us to avoid capturing semantics in var names. 
``` suggestion
    Duration expirationDurationInHours;
```

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v1/ResourceClustersNonLeaderRedirectRoute.java
```diff
@@ -263,7 +263,11 @@ private Route getTaskExecutorState(ClusterID clusterID, TaskExecutorID taskExecu
     private Route disableTaskExecutors(ClusterID clusterID) {
         return entity(Jackson.unmarshaller(DisableTaskExecutorsRequest.class), request -> {
             log.info("POST /api/v1/resourceClusters/{}/disableTaskExecutors called with body {}", clusterID, request);
-            return withFuture(gateway.getClusterFor(clusterID).disableTaskExecutorsFor(request.getAttributes(), Instant.now().plus(Duration.ofHours(24))));
+            return withFuture(gateway.getClusterFor(clusterID).disableTaskExecutorsFor(
+                request.getAttributes(),
+                request.getExpirationDurationInHours() == null || request.getExpirationDurationInHours() <= 0L ?
```
sundargates, line 268: If it's nullable, then we should capture this in the type. I would prefer that this is not nullable, btw since this is under our control.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/DisableTaskExecutorsRequest.java
Andyz26: This is for the JSON payload in user request and specifying the time format for the duration here is just painful (plus the easy behavior regression from Jackson version etc.)

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v1/ResourceClustersNonLeaderRedirectRoute.java
Andyz26: ok will add nonnull annotation to the request.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/DisableTaskExecutorsRequest.java
sundargates: I was thinking about this today, and I wanted to know if there was a reason why we want to have a configurable TTL per request. In other words, why not configure this at the control plane level?

Andyz26: Yeah I was planning to configure this from the caller side (upgrade workflow) config (also inside nfmantis) as the length of this duration shall be decided by the upgrade workflow logic.

PR #371, opened 2023-03-18
Support NotWritable timeout on SSE connection

calvin681:
### Context

This is an attempt to mitigate the stale connection issue from SSE server side. From observation, a stale connection is still reported as `active` but NOT `writable`. Therefore, we can allow the server to close the connection proactively if the channel remains not writable for an configurable amount of time.

### Checklist

### DIFF
--- a/mantis-common/src/main/java/io/mantisrx/common/MantisProperties.java
+++ b/mantis-common/src/main/java/io/mantisrx/common/MantisProperties.java
@@ -28,17 +28,18 @@ public class MantisProperties {
         env = System.getenv();
     }
 
+    @Deprecated
     public static MantisProperties getInstance() {
         return instance;
     }
 
+    @Deprecated
+    /**
+     * Use {@link #getProperty(String)} instead.
+     */
     public String getStringValue(String name) {
         if (name != null) {
-            if (env.containsKey(name)) {
-                return getProperty(name, env.get(name));
-            } else {
-                return getProperty(name);
-            }
+            return getProperty(name, env.get(name));
         } else {
             return null;
         }
@@ -46,10 +47,13 @@ public class MantisProperties {
     }
 
     public static String getProperty(String key) {
-        return System.getProperty(key, System.getenv(key));
+        return getProperty(key, null);
     }
 
     public static String getProperty(String key, String defaultVal) {
+        if (key == null) {
+            return null;
+        }
         String value = System.getProperty(key);
         if (value != null) {
             return value;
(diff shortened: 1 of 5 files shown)


mantis-common/src/main/java/io/mantisrx/common/MantisProperties.java
```diff
@@ -28,24 +28,40 @@ private MantisProperties() {
         env = System.getenv();
     }
 
+    @Deprecated
```
calvin681, line 31: @Andyz26 Please take a look. I fixed MantisProperties to properly fallback to System.getProperty and then System.getEnv. I believe you intend for this class to be used with static methods going forward. So I'm deprecating all the instance methods.

sundargates (commented):
Can we refactor PushServer in a separate diff or commit, as it's hard to differentiate between code and whitespace changes?

mantis-common/src/main/java/io/mantisrx/common/MantisProperties.java
```diff
@@ -28,24 +28,40 @@ private MantisProperties() {
...
+    /**
+     * Use {@link #getProperty(String)} instead.
+     */
     public String getStringValue(String name) {
-        if (name != null && env.containsKey(name)) {
-            return System.getProperty(name, env.get(name));
+        if (name != null) {
```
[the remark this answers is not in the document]
calvin681: Defensive to prevent NPE from map.get

mantis-network/src/main/java/io/reactivex/mantis/network/push/PushServer.java
```diff
@@ -235,128 +245,174 @@ protected Observable<Void> manageConnectionWithCompression(final DefaultChannelW
...
+            .addCounter("channelWritable")
+            .addCounter("channelNotWritable")
+            .addCounter("channelNotWritableTimeout")
+            .build();
+        metricsRegistry.registerAndGet(writableMetrics);
+        Counter channelWritableCounter = writableMetrics.getCounter("channelWritable");
+        Counter channelNotWritableCounter = writableMetrics.getCounter("channelNotWritable");
+        Counter channelNotWritableTimeoutCounter = writableMetrics.getCounter("channelNotWritableTimeout");
+
+        final Future<?> writableCheck;
```
calvin681, line 278: @sundargates  These lines are the only ones that are relevant. Everything else is not important.

mantis-common/src/main/java/io/mantisrx/common/MantisProperties.java
calvin681: You need System properties and env var because that's how some jobs get job parameters.

Andyz26 (approved):

PR #375, opened 2023-03-22
Pass classloader to ThreadFactoryBuilder

fdc-ntflx:
### Context

### Checklist


mantis-runtime/src/main/java/io/mantisrx/runtime/Context.java
```diff
@@ -159,5 +167,5 @@ public Observable<WorkerMap> getWorkerMapObservable() {
         return this.workerMapObservable;
     }
 
-
+    public ClassLoader getClassLoader() { return this.classLoader; }
```
Andyz26, line 170: Is the idea to have this "custom" cl as nullable? It might be a bit confusing for users to use it without knowing whether or not it's appropriate.

sundargates (approved):
LGTM! Thanks for finding this issue and fixing it!

PR #378, opened 2023-03-24
Close RpcSystem on TE shutdown

fdc-ntflx:
### Context

This PR aims to cleanup the temp folder where flink-rpc-akka jar is stored by `MantisAkkaRpcSystemLoader`. Both graceful and abrupt shutdown should be covered.

### Checklist

### DIFF
--- a/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutorStarter.java
+++ b/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutorStarter.java
@@ -53,6 +53,7 @@ import org.apache.flink.runtime.rpc.RpcUtils;
 public class TaskExecutorStarter extends AbstractIdleService {
     private final TaskExecutor taskExecutor;
     private final HighAvailabilityServices highAvailabilityServices;
+    private final RpcSystem rpcSystem;
 
     @Override
     protected void startUp() {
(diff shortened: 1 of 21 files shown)


mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutorStarter.java
```diff
@@ -68,6 +69,7 @@ protected void shutDown() throws Exception {
             .exceptionally(throwable -> null)
             .thenCompose(dontCare -> Services.stopAsync(highAvailabilityServices,
                 MoreExecutors.directExecutor()))
+            .thenRunAsync(rpcSystem::close)
```
sundargates, line 72: Instead of doing this in TaskExecutorStarter, why not do this from the TaskExecutor itself?

fdc-ntflx: +1

fdc-ntflx: we pass `RpcService` in `TaskExecutor` which doesn't have any reference to `RpcSystem`. I think this is the right way to close our custom rpc system.

Andyz26: does this gets run if the worker crashed?

Andyz26 (approved):

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutorStarter.java
sundargates, line 72: Should we use the caller thread, in this case, using MoreExecutors.directExecutor()?

fdc-ntflx: Yes it does.

PR #379, opened 2023-03-24
RuntimeTask host thread + Logging improvements on TE

Andyz26:
### Context

* Currently the runtime task is executed in the IO executor service pool and mixed with other IO ops. Move runtime task execution to its own thread and not share it with the IO thread pool.
* Logging improvements on TE lifecycle.

### Checklist

### DIFF
--- a/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
+++ b/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
@@ -172,14 +172,13 @@ public class TaskExecutor extends RpcEndpoint implements TaskExecutorGateway {
         log.info("Starting executor registration: {}", this.taskExecutorRegistration);
 
         this.ioExecutor =
-            Executors.newFixedThreadPool(
-                Hardware.getNumberCPUCores(),
+            Executors.newCachedThreadPool(
                 new ExecutorThreadFactory("taskexecutor-io"));
 
         this.runtimeTaskExecutor =
-            Executors.newFixedThreadPool(
-                1,
+            Executors.newCachedThreadPool(
                 new ExecutorThreadFactory("taskexecutor-runtime"));
+
         this.resourceManagerCxnIdx = 0;
         this.taskFactory = taskFactory == null ? new SingleTaskOnlyFactory() : taskFactory;
     }
--- a/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
+++ b/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
@@ -97,6 +97,7 @@ public class TaskExecutor extends RpcEndpoint implements TaskExecutorGateway {
     private final TaskExecutorRegistration taskExecutorRegistration;
     private final CompletableFuture<Void> startFuture = new CompletableFuture<>();
     private final ExecutorService ioExecutor;
+    private final ExecutorService runtimeTaskExecutor;
     private final ListenerCallQueue<Listener> listeners = new ListenerCallQueue<>();
 
     // the reason the MantisMasterGateway field is not final is because we expect the HighAvailabilityServices
@@ -174,6 +175,11 @@ public class TaskExecutor extends RpcEndpoint implements TaskExecutorGateway {
             Executors.newFixedThreadPool(
                 Hardware.getNumberCPUCores(),
                 new ExecutorThreadFactory("taskexecutor-io"));
+
+        this.runtimeTaskExecutor =
+            Executors.newFixedThreadPool(
+                1,
+                new ExecutorThreadFactory("taskexecutor-runtime"));
         this.resourceManagerCxnIdx = 0;
         this.taskFactory = taskFactory == null ? new SingleTaskOnlyFactory() : taskFactory;
     }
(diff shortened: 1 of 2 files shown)


mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
```diff
@@ -174,6 +175,11 @@ public TaskExecutor(
             Executors.newFixedThreadPool(
                 Hardware.getNumberCPUCores(),
                 new ExecutorThreadFactory("taskexecutor-io"));
+
+        this.runtimeTaskExecutor =
```
fdc-ntflx, line 179: what benefits we get in separating this?

fdc-ntflx:
What tests did you run to validate this change?

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
Andyz26: Having main runtime task mixed in IO thread pool can potentially cause expected block/wait with other network IO ops?

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
```diff
@@ -174,6 +175,11 @@ public TaskExecutor(
             Executors.newFixedThreadPool(
                 Hardware.getNumberCPUCores(),
```
fdc-ntflx, line 176: shall we do #cores - 1 now?

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
sundargates: Hmm, have we seen any actual problems associated with this? I'm not convinced we need a dedicated thread to start the runtime task. I agree this is a problem if the io executor was single-threaded. But it looks like it's not.

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
sundargates, line 176: Given that this pool is meant for IO-bound work, we can increase the number of threads to greater than the cores to address the concerns this diff is trying to solve.

Andyz26: any concern on isolating main runtime task from IO work? I agree it won't cause trouble sharing with IO pool most of the time but it will be cleaner on logging/debugging and safer in case some unexpected IO work took over too many threads in future.

Andyz26: the container can still get multiple cores?

sundargates: Wouldn't adding an extra thread in the above line have the same effect as what you are doing? To be clear, I suggest we increase the number of IO threads to something like 3x the number of cores. 

My main concern was about the complexity. When someone looks at this code, they need to reason why we need three thread pools, and the obvious question would be, why not just use the IO pool for starting the task?

Andyz26: From the code readability perspective, seeing IO-related work separated from the main execution thread is a pretty common practice right? I actually feel confused when I was analyzing the logs and saw some main entry operations were done on an IO executor thread and wondered if something went wrong with IO handling.

sundargates: I don't understand your point here - we already have the main thread executor, AFAICT. All the RPC interactions that change the state of the task executor do happen on this thread as expected. 

```
protected MainThreadExecutor getMainThreadExecutor() {
    return mainThreadExecutor;
}
```
You can see references to this executor in all the RPC interactions. Now, starting the task is not running on this executor because that's mainly IO work (such as starting the spring boot routine, etc...), and these do not change the state of the task executor.

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
```diff
@@ -174,6 +175,11 @@ public TaskExecutor(
...
                 new ExecutorThreadFactory("taskexecutor-io"));
+
+        this.runtimeTaskExecutor =
+            Executors.newFixedThreadPool(
+                1,
```
sundargates, line 181: If you merge this, I recommend increasing the number of threads to more than one.

sundargates (approved):

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
Andyz26: I see what you mean. In the TE right now:
* RPC service main thread which handles TE level interactions and state change (e.g. start task/cancel task).
* IO executor to handle resourceMgr connection IO, user code class loader setup, TE-level event queue dispatcher, and entry of the runtimeTask instance (e.g. create/maintain SB context).

- RuntimeTask level interactions cannot be part of the RPC service main thread as there are blocking operations. 

- In the meantime, the IO executor does handle a variety of work already and most of them IO related, but the actual runtimeTask service entry (e.g. SB context start) is not.

- Splitting the runtime executor means SB entry code will no longer be mixed with IO work in the same thread pool and logs, which will reduce the complexity to understand what happens in the runtimeTask lifecycle (avoid the pain of seeing all the interleaving logs on IO executor threads).

- Any future change introducing more IO work here or lifecycle changes or blocking local event dispatcher work in the runtime entry will not have unexpected blocking behavior sharing threads. This outweighs the overhead of one more executor in TE IMO.

PR #381, opened 2023-03-27
Add docs for mantis dsl

hmitnflx:
### Context

This adds docs relevant to mantis dsl added in #317

### Checklist


docs/docs/develop/writing-jobs/word-count.md
```diff
@@ -129,6 +111,13 @@ $ ../../gradlew execute
 
 ```
 
+### Old Implementation
+If you find the new operator API limiting, you can continue using existing RxJava based implementation. It's available in [WordCountJob.java](https://github.com/Netflix/mantis/blob/master/mantis-examples/mantis-examples-wordcount/src/main/java/com/netflix/mantis/examples/wordcount/WordCountJob.java).
+A few callouts using this approach are 
+1. keyBy is not distributed and is local to the worker
```
calvin681, line 117: We should have more documentation for things like these. Are they explained in more details in Java docs? If so, link to them. We need to have pretty detailed Java doc on each of these methods similar to RxJava Observable Java doc.

mantis-runtime/src/main/java/io/mantisrx/runtime/core/MantisStream.java
```diff
@@ -26,21 +27,97 @@
...
+     * (worker-thread). Materialize breaks this chaining and all the pending
+     * operators are combined into a single scalar stage.
+     * Any subsequent operators will go into a new scalar stage [with chaining]
+     *
+     * @return a new {@link MantisStream} instance
+     */
     MantisStream<T> materialize();
 
+    /**
+     * Partitions the elements of the stream based on the key extracted by
```
calvin681, line 115: Following previous comment, we should call out important details such as local vs distributed.

Issue #355, opened 2023-03-03
Upgrade JDK version to 17


sundargates:
### Context

We want mantis to be on Java17. Unfortunately, this would require several gradle changes. So, the goal of this task is to make these changes and ensure the builds succeed.

Adithya2907:
I think one of the reasons it fails for higher versions of JDK is the `DatatypeConverter` package. As [this](https://stackoverflow.com/questions/61104983/cannot-resolve-datatypeconverter-class-in-java) answer on StackOverflow explains:
```
The module javax.xml.bind has been put outside the jdk as of Java 9.

Also, as part of the donation of Java EE to the Eclipse Foundation, its new name is Jakarta EE and accordingly the namespace was renamed to Jakarta. So you have to make the following modifications :
```

sundargates:
@Adithya2907 would you like to pick this up?

Adithya2907:
Yeah, sure I'd like to pick this up.

cypherean:
Hey @Adithya2907 
Are you working on this?

Adithya2907:
Hi @cypherean , yes, I'm working on this. I've been trying to understand the codebase till now, and now I'll be migrating it to JDK 17. I'm seeing which all packages and libraries are failing due to the migration first

PR #390, opened 2023-03-29
Jdk migration of mantis-common to compile on JDK 17

Adithya2907:
### Context

Upgrade JDK version to 17. This PR upgrades the mantis-common module.
Part of https://github.com/Netflix/mantis/issues/355

- `../gradlew build` in the `mantis-common` module fails `org.mockito.exceptions.base.MockitoException at OperatorGroupByTest.java:1039 Caused by: java.lang.reflect.InaccessibleObjectException at OperatorGroupByTest.java` error.
- The reason for the error is using a version of the `Mockito` package not supported by JDK 17.
- Therefore, updated the dependency from `2.0.+` to `2.23.4` which solves this problem.

Attaching screenshot of success build of the `mantis-common` module using JDK 17.
image

### Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -148,6 +148,19 @@ subprojects {
         options.compilerArgs << "-Xlint:deprecation"
     }
 
+    if(JavaVersion.current() != JavaVersion.VERSION_1_8) {
+        targetCompatibility = JavaVersion.VERSION_17
+        tasks.withType(JavaCompile) {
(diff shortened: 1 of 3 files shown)


Andyz26 (changes requested):
We would prefer handling the Java version as a whole instead of a single module level. Please do not pursue such a change for now as this requires more compatibility validation coordination.

PR #400, opened 2023-03-31
Upgrade API: add forceUpgradeOnSameImage flag

Andyz26:
### Context

Add a flag in upgrade cluster API to support skipping on existing containers on the target image id.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/UpgradeClusterContainersRequest.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/UpgradeClusterContainersRequest.java
@@ -34,4 +34,6 @@ public class UpgradeClusterContainersRequest {
     MantisResourceClusterEnvType optionalEnvType;
 
     int optionalBatchMaxSize;
+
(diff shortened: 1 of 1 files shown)


fdc-ntflx (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/UpgradeClusterContainersRequest.java
```diff
@@ -34,4 +34,6 @@ public class UpgradeClusterContainersRequest {
     MantisResourceClusterEnvType optionalEnvType;
```
sundargates, line 34: What's the need for cluster env? Are we not following the island model wrt operations?

sundargates (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/UpgradeClusterContainersRequest.java
Andyz26: This optional metadata was introduced in early days but is currently ignored in nfmantis. I can track a clean up task around this.

Issue #384, opened 2023-03-29
Ensure JobActor has JobID as part of every emitted log line


sundargates:
### Context

We use the slf4j logger API to emit logs from within all the actors. All these actors have a unique way of identifying themselves. For instance, the JobActor represents an instance of a running Job with a unique JobID. Thus, it would be nice to ensure all the logs emitted from the JobActor contain this information. Today, this is done in an ad-hoc fashion. Some logs have the JobID, while others don't. It would be nice to develop a more structured way of solving this.

### Goals

Look at the below line from JobActor.java to understand what the problem is. 
```
LOGGER.info("Initializing Job {}", jobId);
```
We manually add the jobID in log line above. However, in other places, we fail to do this—for instance, look at the example below (also from JobActor.java).
```
LOGGER.info("Stored mantis job");
```

Instead, it would be nice to have a more structured way of solving this problem—something like the one below.

```
jobLogger.info("Whatever");
```
And the expectation is that jobLogger adds the jobID to every emitted logline by default.

kronsbein:
Hey @sundargates I’m interested in working on this, could you please assign me?

7ue5wu:
Hi @sundargates, Is this issue assigned? If not, I wanna take this issue.

Yosefnago:
What is the state of this issue?

PR #408, opened 2023-04-04
Ensure JobActor has JobID as part of every emitted log line #384

7ue5wu:
### Context

Logger adds the jobID to every emitted logline by default.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
@@ -77,6 +77,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
 import rx.Observable;
 import rx.schedulers.Schedulers;
 import rx.subjects.BehaviorSubject;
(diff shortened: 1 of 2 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -600,6 +601,7 @@ private Receive buildInitializedBehavior() {
                 // UNEXPECTED MESSAGES END //
                 .match(Terminated.class, this::onTerminated)
                 .matchAny(x -> {
+                    MDC.put("name", name);
```
sundargates, line 604: Doesn't this make assumptions about the thread context? Will these assumptions be held for Akka actors?

7ue5wu: Oh, I didn't consider it. So does it mean it is better to use ActorContext.getLog() to obtain the logger and clear the MDC after processing the current message?

sundargates: I don't think we want to use MDC for this.

7ue5wu: It seems I thought it the wrong way. Thank you for telling me and I will fix it after my busy final.

PR #411, opened 2023-04-06
JobActor::WorkerManager's logs contain the WorkerID that they correspond to

GautamGottipati:
### Context

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
@@ -36,6 +36,7 @@ import io.mantisrx.master.events.LifecycleEventPublisher;
 import io.mantisrx.master.events.LifecycleEventsProto;
 import io.mantisrx.master.jobcluster.WorkerInfoListHolder;
 import io.mantisrx.master.jobcluster.job.worker.*;
+import io.mantisrx.master.jobcluster.job.worker.WorkerManagerLogger;
 import io.mantisrx.master.jobcluster.proto.JobClusterManagerProto;
 import io.mantisrx.master.jobcluster.proto.JobClusterManagerProto.*;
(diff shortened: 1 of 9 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
```diff
@@ -1286,6 +1288,8 @@ class WorkerManager implements IWorkerManager {
                 .build();
         private volatile boolean stageAssignmentPotentiallyChanged;
 
+        private WorkerManagerLogger workerLogger;
```
Andyz26, line 1291: Introducing extra lifecycle for these loggers doesn't seem to provide enough benefits. Adding the extra metadata to the logger message is sufficient enough.

Andyz26 (changes requested):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
GautamGottipati: Hi, @Andyz26 can you give me some hints, like what metadata I could add and how.

Andyz26: The issue is about jobId at the job actor level and that is a different structure comparing to this workerManager. Here WorkerManager is not 1 to 1 mapping with a specific worker logically. There is no need to maintain a worker-level logger as long as the information needed is logged.

GautamGottipati: so you mean, I can directly add worker Id to Logger.info, where ever worker id is not mentioned.

Andyz26:
In this case yes.

PR #418, opened 2023-04-13
Make TE allocation strictly following core number

Andyz26:
### Context

Update the allocation logic in TE so that the scheduler will only put the requested node into a TE strictly using the smallest CPU cores to avoid having TEs spreading in different container sizes. Also, introduce an intermediate allocation request contract to help eventual decoupling/removal of Mesos classes.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
@@ -17,7 +17,6 @@
 package io.mantisrx.server.master.resourcecluster;
 
 import io.mantisrx.common.Ack;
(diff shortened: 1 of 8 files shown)


fdc-ntflx (approved):
LGTM!

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
```diff
@@ -72,7 +71,7 @@ public interface ResourceCluster extends ResourceClusterGateway {
      * @param workerId          worker id of the task that's going to run on the node.
      * @return task executor assigned for the particular task.
      */
-    CompletableFuture<TaskExecutorID> getTaskExecutorFor(MachineDefinition machineDefinition, WorkerId workerId);
```
sundargates, line 75: nit: What was the need to introduce a new type here?

PR #421, opened 2023-04-16
Fix compile warnings in mantis-control-plane-client module

### LINKED ISSUE #389: Fix compile warnings in mantis-control-plane-client module
There are compile warnings in the code due to:
- deprecated APIs
- missing `@Nullable/@NotNull` annotations
- incomplete javadocs
- lambda can be replaced with method references
- raw usage of generic types
- unnecessary type parameters (like `Map<String> m = new HashMap<String>()`)
- anonymous class can be replaced with lambdas
and more.

We don't need to solve all of them, but a good coding exercise is to address most those compile warnings when possible.

This tracks the effort for the `mantis-control-plane-client` module to make the code change scoped.

mabelbot:
### Context

Contains changes to resolve #389

Deprecation Warnings addressed (to be reviewed):
- HttpUtility.java:63: warning: [deprecation] getUri() in HttpRequest has been deprecated
- MantisMasterClientApi.java:778: warning: [deprecation] JsonCodec(Class<T>) in JsonCodec has been deprecated

Progress:
- Remove deprecated getUri() and replace with uri()
- Remove deprecated JsonCodec constructor and replace with JacksonCodecs (as indicated in javadoc) pojo method

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/MantisMasterClientApi.java
+++ b/mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/MantisMasterClientApi.java
@@ -27,7 +27,7 @@ import io.mantisrx.runtime.JobSla;
 import io.mantisrx.runtime.MantisJobDefinition;
 import io.mantisrx.runtime.MantisJobState;
 import io.mantisrx.runtime.WorkerMigrationConfig;
-import io.mantisrx.runtime.codec.JsonCodec;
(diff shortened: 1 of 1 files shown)


liuml07 (changes requested):
Is it better to take care all classes in the module? Ideally we can consolidate a single PR to address most of the compiler warnings. For e.g. in `MantisMasterClientApi` I can see multiple ones open. 

You can also check code warnings in IntelliJ IDE editor (open a .java file and check yellow marks on the right).

PR #422, opened 2023-04-18
Add API to enable/disable RC scaler

fdc-ntflx:
### Context

The goal of this PR is to introduce a new API to enable/disable scaler actor for a given SKU. This is useful when for example a new ASG is being provisioned to replace an existing one. During such operation we want to disable autoscaling for these ASGs until provisioning is done and the old ASG is marked as disabled. 

A timer is also added to re-enable the scaler in case something goes wrong and ASG provisioning couldn't complete successfully.  

This PR also aims to increase the cool down period for scaling down operations. The approach taken here is simplest possible which is hardcoding a factor of 5 over the configured cool down only for scale down operations.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v1/ResourceClustersNonLeaderRedirectRoute.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v1/ResourceClustersNonLeaderRedirectRoute.java
@@ -42,6 +42,7 @@ import io.mantisrx.master.resourcecluster.proto.ResourceClusterScaleRuleProto.Ge
 import io.mantisrx.master.resourcecluster.proto.ResourceClusterScaleRuleProto.GetResourceClusterScaleRulesResponse;
 import io.mantisrx.master.resourcecluster.proto.ScaleResourceRequest;
 import io.mantisrx.master.resourcecluster.proto.ScaleResourceResponse;
+import io.mantisrx.master.resourcecluster.proto.SetResourceClusterScalerStatusRequest;
 import io.mantisrx.master.resourcecluster.proto.UpgradeClusterContainersRequest;
 import io.mantisrx.master.resourcecluster.proto.UpgradeClusterContainersResponse;
 import io.mantisrx.server.master.config.ConfigurationProvider;
@@ -75,6 +76,7 @@ import lombok.extern.slf4j.Slf4j;
  * /api/v1/resourceClusters/{}/scaleSku                               (POST)
  * /api/v1/resourceClusters/{}/upgrade                                (POST)
  * /api/v1/resourceClusters/{}/disableTaskExecutors                   (POST)
+ * /api/v1/resourceClusters/{}/setScalerStatus                        (POST)
  * <p>
  * <p>
  * /api/v1/resourceClusters/{}/scaleRule                              (POST)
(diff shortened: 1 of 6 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterScalerActor.java
```diff
@@ -70,6 +70,9 @@ public class ResourceClusterScalerActor extends AbstractActorWithTimers {
     // Timer threshold of refreshing cluster scale rules from storage provider.
     private final Duration ruleSetRefreshThreshold;
 
+    // Timer to re-enable scaler
+    private final Duration enableScalerThreshold = Duration.ofMinutes(60);
```
Andyz26, line 74: Configurable from request?

Andyz26 (approved):
Any UT avaiable?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterScalerActor.java
fdc-ntflx: good idea.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
```diff
@@ -102,6 +102,15 @@ public interface ResourceCluster extends ResourceClusterGateway {
...
     CompletableFuture<Ack> disableTaskExecutorsFor(Map<String, String> attributes, Instant expiry);
 
+    /**
+     * Enables/Disables scaler for a given skuID of a given clusterID
+     *
+     * @param skuID skuID whom scaler will be enabled/disabled.
+     * @param enabled whether the scaler will be enabled/disabled.
+     * @return a future that completes when the underlying operation is registered by the system
+     */
+    CompletableFuture<Ack> setScalerStatus(ContainerSkuID skuID, Boolean enabled);
```
sundargates, line 112: Is this interface the right place for this method? It feels like we will be delegating to the scaler actor here.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v1/ResourceClustersNonLeaderRedirectRoute.java
```diff
@@ -269,6 +279,13 @@ private Route disableTaskExecutors(ClusterID clusterID) {
         });
     }
 
+    private Route setScalerStatus(ClusterID clusterID) {
+        return entity(Jackson.unmarshaller(SetResourceClusterScalerRequest.class), request -> {
+            log.info("POST /api/v1/resourceClusters/{}/setScalerStatus called with body {}", clusterID, request);
+            return withFuture(gateway.getClusterFor(clusterID).setScalerStatus(ContainerSkuID.of(request.getSkuId()), request.getEnabled()));
```
sundargates, line 285: I think we should be using ResourceClusterRouteHandler here instead.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterScalerActor.java
```diff
@@ -323,24 +363,37 @@ static class ClusterAvailabilityRule {
...
         }
-        public Optional<ScaleDecision> apply(UsageByGroupKey usage) {
-            // Cool down check
-            if (this.lastActionInstant.plusSeconds(this.scaleSpec.getCoolDownSecs()).compareTo(clock.instant()) > 0) {
-                log.debug("Scale CoolDown skip: {}, {}", this.scaleSpec.getClusterId(), this.scaleSpec.getSkuId());
-                return Optional.empty();
-            }
 
-            this.lastActionInstant = clock.instant();
+        public void resetLastActionInstant() {
```
sundargates, line 376: nit: I would update the ticker from within the DS whenever any field of the DS is being modified. This way, you don't have to make explicit calls for every interaction. Also, in the future, when you have new references, there's less chance of making a mistake.

sundargates (approved):
LGTM

PR #425, opened 2023-04-19
Handle TE invalid unAssigned state error

Andyz26:
### Context
[Found from prod log scan]
In the control plane transition, in-error/crashed TE can have an initially empty state and gets an illegal state error if the control plane decides to unassign its current work. There is no action to take for the control plane at that point as it hasn't been able to re-establish a connection with TE. If the TE has a discrepancy on task assignment state if it should get handled when it gets back in the heartbeat message instead.

Reduced the catch scope and move the error log into a warning.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
@@ -472,18 +472,19 @@ class ResourceClusterActor extends AbstractActorWithTimers {
     }
 
     private void onTaskExecutorAssignmentTimeout(TaskExecutorAssignmentTimeout request) {
-        try {
-            TaskExecutorState state = this.executorStateManager.get(request.getTaskExecutorID());
-            if (state.isRunningTask()) {
-                log.debug("TaskExecutor {} entered running state alraedy; no need to act", request.getTaskExecutorID());
-            } else {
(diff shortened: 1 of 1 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -472,18 +472,19 @@ private void onTaskExecutorAssignmentRequest(TaskExecutorAssignmentRequest reque
...
+            log.debug("TaskExecutor {} entered running state alraedy; no need to act", request.getTaskExecutorID());
+        } else {
+            try
+            {
                 boolean stateChange = state.onUnassignment();
                 if (stateChange) {
                     this.executorStateManager.markAvailable(request.getTaskExecutorID());
                 }
+            } catch (IllegalStateException e) {
+                log.warn("Failed to un-assign taskExecutor {}", request.getTaskExecutorID(), e);
```
sundargates, line 486: Check if it's unassigned already -> If yes, then no need to emit this log line.

Andyz26: The "availabilityState" is null in this case. Do you mean check the registration state and ignore on "unregistered" state?

sundargates: yes

PR #438, opened 2023-05-12
Adding the capability to submit the latest job cluster configuration as a job

sundargates:
### Context

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/proto/JobClusterManagerProto.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/proto/JobClusterManagerProto.java
@@ -1554,18 +1554,21 @@ public class JobClusterManagerProto {
         }
     }
 
+    @ToString
     public static final class SubmitJobRequest extends BaseRequest {
         private final Optional<JobDefinition> jobDefinition;
         private final String submitter;
         private final String clusterName;
         private final boolean isAutoResubmit;
+        private final boolean submitLatest;
 
         @JsonCreator
         @JsonIgnoreProperties(ignoreUnknown = true)
         public SubmitJobRequest(
                 @JsonProperty("name") final String clusterName,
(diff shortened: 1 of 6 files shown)


hmitnflx (approved):
This is a nice and much needed change!

I'd argue that we shouldn't have the other rollout mode at all. There's inherently no value in submitting a new job with old job config (which is different from cluster config).

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/proto/JobClusterManagerProto.java
```diff
@@ -1554,18 +1554,21 @@ public String toString() {
...
+        private final boolean submitLatest;
 
         @JsonCreator
         @JsonIgnoreProperties(ignoreUnknown = true)
         public SubmitJobRequest(
                 @JsonProperty("name") final String clusterName,
                 @JsonProperty("user") final String user,
-                @JsonProperty(value = "jobDefinition") final Optional<JobDefinition> jobDefinition) {
+                @JsonProperty(value = "jobDefinition") final Optional<JobDefinition> jobDefinition,
+                @JsonProperty("submitLatest") final boolean submitLatest) {
```
Andyz26, line 1571: If I understand correctly this flag is to submit from the current job cluster definition instead of the most recent job's definition (the latest job can have some overrides in its configs). maybe something like "submitLatestJobCluster".

Andyz26 (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/proto/JobClusterManagerProto.java
sundargates: Yes, that's the intention. However, we should not allow overrides in the job definition config in the long run - this way, what's shown on the job cluster definition is what gets run from the job.

PR #447, opened 2023-05-22
fix issue#446

7ue5wu:
### Context
When running the source job example by following the Mantis document, error happened.

Exception in thread "main" java.lang.NoClassDefFoundError: com/netflix/spectator/api/Registry
Caused by: java.lang.ClassNotFoundException: com.netflix.spectator.api.Registr

### Checklist

### DIFF
--- a/mantis-common/build.gradle
+++ b/mantis-common/build.gradle
@@ -30,8 +30,7 @@ dependencies {
     api "org.xerial.snappy:snappy-java:$snappyVersion"
     api "org.jctools:jctools-core:$jctoolsVersion"
 
-    // spectatorApi should be packaged at entry point level to avoid version conflicts.
-    compileOnly libraries.spectatorApi
+    implementation libraries.spectatorApi
 
     api libraries.mantisShaded
     api libraries.rxNettyShaded


hmitnflx (approved):

Andyz26 (changes requested):
@7ue5wu you might want to update https://github.com/Netflix/mantis/blob/master/mantis-runtime/build.gradle too.

7ue5wu:
> @7ue5wu you might want to update https://github.com/Netflix/mantis/blob/master/mantis-runtime/build.gradle too.

Hi Andy, I already updated this one too. Is it the right update? I am confused because it seems nothing different when running after I update this one.

Andyz26:
> > @7ue5wu you might want to update https://github.com/Netflix/mantis/blob/master/mantis-runtime/build.gradle too.
> 
> Hi Andy, I already updated this one too. Is it the right update? I am confused because it seems nothing different when running after I update this one.

A bit of context: these were made compileOnly to fix some dynamic load problem in our internal wrapper layer and this is no longer needed now. Also, I don't know if you still need to declare the dep in the runtime module anymore if it's using "implementation" directly in common (in that case you can remove the part in runtime).

7ue5wu:
> > > @7ue5wu you might want to update https://github.com/Netflix/mantis/blob/master/mantis-runtime/build.gradle too.
> > 
> > 
> > Hi Andy, I already updated this one too. Is it the right update? I am confused because it seems nothing different when running after I update this one.
> 
> A bit of context: these were made compileOnly to fix some dynamic load problem in our internal wrapper layer and this is no longer needed now. Also, I don't know if you still need to declare the dep in the runtime module anymore if it's using "implementation" directly in common (in that case you can remove the part in runtime).

It seems we should still keep this part temporarily because the ServerSentRequestHandler still needs the libraries for now.   So I didn't change mantis-runtime/build.gradle and remain it onlyCompile. And the synthetic-sourcejob can run successfully.

Andyz26:
> > > > @7ue5wu you might want to update https://github.com/Netflix/mantis/blob/master/mantis-runtime/build.gradle too.
> > > 
> > > 
> > > Hi Andy, I already updated this one too. Is it the right update? I am confused because it seems nothing different when running after I update this one.
> > 
> > 
> > A bit of context: these were made compileOnly to fix some dynamic load problem in our internal wrapper layer and this is no longer needed now. Also, I don't know if you still need to declare the dep in the runtime module anymore if it's using "implementation" directly in common (in that case you can remove the part in runtime).
> 
> It seems we should still keep this part temporarily because the ServerSentRequestHandler still needs the libraries for now. So I didn't change mantis-runtime/build.gradle and remain it onlyCompile. And the synthetic-sourcejob can run successfully.

Since the common will now bring in the lib file, using "compileOnly" in the runtime will not remove it from the dep chain. I just feel that it's now a bit confusing using "compileOnly" in runtime when the lib is still going to be included.

7ue5wu:
> > > > > @7ue5wu you might want to update https://github.com/Netflix/mantis/blob/master/mantis-runtime/build.gradle too.
> > > > 
> > > > 
> > > > Hi Andy, I already updated this one too. Is it the right update? I am confused because it seems nothing different when running after I update this one.
> > > 
> > > 
> > > A bit of context: these were made compileOnly to fix some dynamic load problem in our internal wrapper layer and this is no longer needed now. Also, I don't know if you still need to declare the dep in the runtime module anymore if it's using "implementation" directly in common (in that case you can remove the part in runtime).
> > 
> > 
> > It seems we should still keep this part temporarily because the ServerSentRequestHandler still needs the libraries for now. So I didn't change mantis-runtime/build.gradle and remain it onlyCompile. And the synthetic-sourcejob can run successfully.
> 
> Since the common will now bring in the lib file, using "compileOnly" in the runtime will not remove it from the dep chain. I just feel that it's now a bit confusing using "compileOnly" in runtime when the lib is still going to be included.

Hi! I have already changed both of them to "Implementation", I think that can make sense this time and the example can run successfully now.

PR #448, opened 2023-05-24
Cache job artifacts on TE right after registration (to speed up job initialization time)

fdc-ntflx:
### Context

The goal of this PR is to introduce a way to "pre"-cache job artifacts into Task Executors before jobs get scheduled in order to speed up job launch times (currently ~1 minute).

## Details 

The idea is to inform TEs of which job artifacts to download from blobstore right after they registered themselves with the resource cluster. The detailed semantic follows these steps:

1. We tell the RC actor the set of artifacts to be cached on a given cluster (in the future we will add a new endpoint for this, but for now the information is stored in kvdal manually)
2. The TEs registers itself to the RC
3. The RC actor tells the TE which artifacts to cache (right now this process is best-effort, I think it's totally fine if some TEs fail to pre-cache and end-up with longer than expected job launch latencies).

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/CacheJobArtifactsRequest.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/CacheJobArtifactsRequest.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2023 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.server.core;
+
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonCreator;
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonProperty;
+import java.io.Serializable;
+import java.net.URI;
+import java.util.List;
+import lombok.NonNull;
+import lombok.Value;
+
+@Value
+public class CacheJobArtifactsRequest implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    @NonNull
+    final List<URI> artifacts;
+
+    @JsonCreator
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    public CacheJobArtifactsRequest(@JsonProperty("artifacts") List<URI> artifacts) {
+        this.artifacts = artifacts;
+    }
+}
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
@@ -17,6 +17,7 @@
 package io.mantisrx.server.master.resourcecluster;
 
 import io.mantisrx.common.Ack;
+import io.mantisrx.server.core.domain.ArtifactID;
 import io.mantisrx.server.core.domain.WorkerId;
 import io.mantisrx.server.worker.TaskExecutorGateway;
 import java.time.Instant;
@@ -62,6 +63,8 @@ public interface ResourceCluster extends ResourceClusterGateway {
     CompletableFuture<List<TaskExecutorID>> getUnregisteredTaskExecutors();
 
     CompletableFuture<ResourceOverview> resourceOverview();
+
+    CompletableFuture<Ack> addNewJobArtifactsToCache(ClusterID clusterID, List<ArtifactID> artifacts);
 
     /**
      * Can throw {@link NoResourceAvailableException} wrapped within the CompletableFuture in case there
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/worker/TaskExecutorGateway.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/worker/TaskExecutorGateway.java
@@ -16,6 +16,7 @@
 package io.mantisrx.server.worker;
 
 import io.mantisrx.common.Ack;
+import io.mantisrx.server.core.CacheJobArtifactsRequest;
 import io.mantisrx.server.core.ExecuteStageRequest;
 import io.mantisrx.server.core.domain.WorkerId;
 import java.util.concurrent.CompletableFuture;
@@ -35,6 +36,14 @@ public interface TaskExecutorGateway extends RpcGateway {
      *                                     in case there's already an existing task that's running on the task executor.
      */
     CompletableFuture<Ack> submitTask(ExecuteStageRequest request);
+
+    /**
+     * instruct the task executor on which job artifact to cache in order to speed up job initialization time.
+     *
+     * @param request List of job artifacts that need to be cached.
+     * @return Ack in any case (this task is best effort).
+     */
+    CompletableFuture<Ack> cacheJobArtifacts(CacheJobArtifactsRequest request);
 
     /**
      * cancel the currently running task and get rid of all of the associated resources.
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v1/JobArtifactsRoute.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v1/JobArtifactsRoute.java
@@ -116,8 +116,6 @@ public class JobArtifactsRoute extends BaseRoute {
                 try {
                     final CompletionStage<UpsertJobArtifactResponse> response = jobArtifactRouteHandler.upsert(new JobArtifactProto.UpsertJobArtifactRequest(jobArtifact));
 
-                    // TODO(fdichiara): introduce job artifact pre-fetching in task executors.
-
                     return completeAsync(
                         response,
                         resp -> complete(
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
@@ -29,6 +29,8 @@ import io.mantisrx.common.WorkerConstants;
 import io.mantisrx.master.resourcecluster.metrics.ResourceClusterActorMetrics;
 import io.mantisrx.master.resourcecluster.proto.GetClusterIdleInstancesRequest;
 import io.mantisrx.master.resourcecluster.proto.GetClusterIdleInstancesResponse;
+import io.mantisrx.server.core.CacheJobArtifactsRequest;
+import io.mantisrx.server.core.domain.ArtifactID;
 import io.mantisrx.server.core.domain.WorkerId;
 import io.mantisrx.server.master.persistence.MantisJobStore;
 import io.mantisrx.server.master.resourcecluster.ClusterID;
@@ -46,12 +48,14 @@ import io.mantisrx.server.master.resourcecluster.TaskExecutorReport.Available;
 import io.mantisrx.server.master.resourcecluster.TaskExecutorReport.Occupied;
 import io.mantisrx.server.master.resourcecluster.TaskExecutorStatusChange;
 import io.mantisrx.server.master.scheduler.JobMessageRouter;
+import io.mantisrx.server.worker.TaskExecutorGateway;
 import io.mantisrx.server.worker.TaskExecutorGateway.TaskNotFoundException;
 import io.mantisrx.shaded.com.google.common.base.Preconditions;
 import io.mantisrx.shaded.com.google.common.collect.Comparators;
 import io.mantisrx.shaded.com.google.common.collect.ImmutableMap;
 import io.vavr.Tuple;
 import java.io.IOException;
+import java.net.URI;
 import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
@@ -100,6 +104,8 @@ class ResourceClusterActor extends AbstractActorWithTimers {
 
     private final ResourceClusterActorMetrics metrics;
 
+    private final HashSet<ArtifactID> jobArtifactsToCache = new HashSet<>();
+
     static Props props(final ClusterID clusterID, final Duration heartbeatTimeout, Duration assignmentTimeout, Duration disabledTaskExecutorsCheckInterval, Clock clock, RpcService rpcService, MantisJobStore mantisJobStore, JobMessageRouter jobMessageRouter) {
         return Props.create(ResourceClusterActor.class, clusterID, heartbeatTimeout, assignmentTimeout, disabledTaskExecutorsCheckInterval, clock, rpcService, mantisJobStore, jobMessageRouter);
     }
@@ -132,6 +138,8 @@ class ResourceClusterActor extends AbstractActorWithTimers {
     @Override
     public void preStart() throws Exception {
         super.preStart();
+        fetchJobArtifactsToCache();
+
         List<DisableTaskExecutorsRequest> activeRequests =
             mantisJobStore.loadAllDisableTaskExecutorsRequests(clusterID);
         for (DisableTaskExecutorsRequest request : activeRequests) {
@@ -186,7 +194,29 @@ class ResourceClusterActor extends AbstractActorWithTimers {
                 .match(ExpireDisableTaskExecutorsRequest.class, this::onDisableTaskExecutorsRequestExpiry)
                 .match(GetTaskExecutorWorkerMappingRequest.class, req -> sender().tell(getTaskExecutorWorkerMapping(req.getAttributes()), self()))
                 .match(PublishResourceOverviewMetricsRequest.class, this::onPublishResourceOverviewMetricsRequest)
+                .match(CacheJobArtifactsOnTaskExecutorRequest.class, this::onCacheJobArtifactsOnTaskExecutorRequest)
+                .match(AddNewJobArtifactsToCacheRequest.class, this::onAddNewJobArtifactsToCacheRequest)
                 .build();
+    }
+
+    private void onAddNewJobArtifactsToCacheRequest(AddNewJobArtifactsToCacheRequest req) {
+        try {
+            mantisJobStore.addNewJobArtifactsToCache(req.getClusterID(), req.getArtifacts());
+            jobArtifactsToCache.addAll(req.artifacts);
+        } catch (IOException e) {
+            log.warn("Cannot add new job artifacts {} to cache in cluster: {}", req.getArtifacts(), req.getClusterID(), e);
+        }
+    }
+
+    private void fetchJobArtifactsToCache() {
+        try {
+            mantisJobStore.getJobArtifactsToCache(clusterID)
+                .stream()
+                .map(ArtifactID::of)
+                .forEach(jobArtifactsToCache::add);
+        } catch (IOException e) {
+            log.warn("Cannot refresh job artifacts to cache in cluster: {}", clusterID, e);
+        }
     }
 
     private GetClusterIdleInstancesResponse onGetClusterIdleInstancesRequest(GetClusterIdleInstancesRequest req) {
@@ -393,6 +423,9 @@ class ResourceClusterActor extends AbstractActorWithTimers {
                 updateHeartbeatTimeout(registration.getTaskExecutorID());
             }
             log.info("Successfully registered {} with the resource cluster {}", registration.getTaskExecutorID(), this);
+            if (!jobArtifactsToCache.isEmpty() && isJobArtifactCachingEnabled()) {
+                self().tell(new CacheJobArtifactsOnTaskExecutorRequest(taskExecutorID, clusterID), self());
+            }
             sender().tell(Ack.getInstance(), self());
         } catch (Exception e) {
             sender().tell(new Status.Failure(e), self());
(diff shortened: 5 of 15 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v1/ResourceClustersLeaderExclusiveRoute.java
```diff
@@ -96,8 +98,12 @@ private Route registerTaskExecutor(ClusterID clusterID) {
                 "POST /api/v1/resourceClusters/{}/actions/registerTaskExecutor called {}",
                 clusterID,
                 request);
+            final String uri = "s3://netflix.s3.genpop.prod/mantis/jobs/raven-connector-job-thin-1.0.147.zip";
```
Andyz26, line 101: this seems to be very specific/hardcoded?

Andyz26:
Can you update the description about what this is for?

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
```diff
@@ -44,6 +44,8 @@ public interface ResourceCluster extends ResourceClusterGateway {
      */
     String getName();
 
+    CompletableFuture<Ack> downloadArtifactsOnTaskExecutor(TaskExecutorID taskExecutorID);
```
sundargates, line 47: What is the semantics you are thinking of for the caching story? I was surprised to find out about this endpoint on the RC.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/worker/TaskExecutorGateway.java
```diff
@@ -36,6 +37,8 @@ public interface TaskExecutorGateway extends RpcGateway {
      */
     CompletableFuture<Ack> submitTask(ExecuteStageRequest request);
 
+    CompletableFuture<Ack> downloadArtifact(DownloadArtifactsRequest request);
```
sundargates, line 40: s/downloadArtifact/cacheArtifact

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
fdc-ntflx: My idea is to persist the mapping RC -> artifacts and get it when TEs get registered. In the future we can probably evolve the structure and make it more sophisticated. But for now this mapping I guess is fine.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/worker/TaskExecutorGateway.java
```diff
@@ -36,6 +37,8 @@ public interface TaskExecutorGateway extends RpcGateway {
      */
     CompletableFuture<Ack> submitTask(ExecuteStageRequest request);
 
+    CompletableFuture<Ack> cacheJobArtifacts(CacheJobArtifactsRequest request);
```
Andyz26, line 40: nit: cache implies it can be cleanup/evicted but in this case, it's more persistent?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -393,7 +397,10 @@ private void onTaskExecutorRegistration(TaskExecutorRegistration registration) {
                 updateHeartbeatTimeout(registration.getTaskExecutorID());
             }
             log.info("Successfully registered {} with the resource cluster {}", registration.getTaskExecutorID(), this);
+            self().tell(new CacheJobArtifactsOnTaskExecutorRequest(taskExecutorID, clusterID), self());
+            log.info("[fdc-91] 1");
```
Andyz26, line 401: clean up?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -600,6 +607,21 @@ private void updateHeartbeatTimeout(TaskExecutorID taskExecutorID) {
             heartbeatTimeout);
     }
 
+    private void onCacheJobArtifactsOnTaskExecutorRequest(CacheJobArtifactsOnTaskExecutorRequest request) {
+        log.info("[fdc-91] onCacheJobArtifactsRequest: {}", request);
```
Andyz26, line 611: maybe convert to debug + metrics?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -393,7 +397,10 @@ private void onTaskExecutorRegistration(TaskExecutorRegistration registration) {
                 updateHeartbeatTimeout(registration.getTaskExecutorID());
             }
             log.info("Successfully registered {} with the resource cluster {}", registration.getTaskExecutorID(), this);
+            self().tell(new CacheJobArtifactsOnTaskExecutorRequest(taskExecutorID, clusterID), self());
+            log.info("[fdc-91] 1");
             sender().tell(Ack.getInstance(), self());
+            log.info("[fdc-91] 2");
```
Andyz26, line 403: ditto

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -600,6 +607,21 @@ private void updateHeartbeatTimeout(TaskExecutorID taskExecutorID) {
             heartbeatTimeout);
     }
 
+    private void onCacheJobArtifactsOnTaskExecutorRequest(CacheJobArtifactsOnTaskExecutorRequest request) {
+        log.info("[fdc-91] onCacheJobArtifactsRequest: {}", request);
+        TaskExecutorState state = this.executorStateManager.get(request.getTaskExecutorID());
+        if (state != null) {
+            TaskExecutorGateway gateway = state.getGateway().join();
+            try {
+                List<URI> artifacts = mantisJobStore.getJobArtifactsToCache(request.getClusterID());
```
Andyz26, line 616: shall the actor store/cache this instead of making a call to store on every TE registration? Also it can use this state to determine whether it needs to trigger an cacheRequest message in the registration call directly?

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/ClassLoaderHandle.java
```diff
@@ -47,6 +47,8 @@ public interface ClassLoaderHandle extends Closeable {
     UserCodeClassLoader getOrResolveClassLoader(Collection<URI> requiredJarFiles,
                                                 Collection<URL> requiredClasspaths) throws IOException;
 
+    void cacheJobArtifacts(Collection<URI> artifacts);
```
Andyz26, line 50: for your consideration: this API seems a bit awkward to live in classloaderHandle as it has no relation to any class loader? Maybe move to its own class?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
```diff
@@ -689,6 +691,15 @@ public List<JobArtifact> listJobArtifacts(String name, String version) throws IO
             }).filter(Objects::nonNull).collect(Collectors.toList());
     }
 
+    @Override
+    public List<URI> listJobArtifactsToCache(ClusterID clusterID) throws IOException {
```
Andyz26, line 695: how does these get updated? I don't see a write path.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
fdc-ntflx: yeah logs will go away once I finish testing. no worries

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
fdc-ntflx: I don't see that many benefits in caching this very tiny info in memory which I'm sure can be retrieved from the datastore in milliseconds.

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/BlobStoreAwareClassLoaderHandle.java
```diff
@@ -93,6 +93,15 @@ public UserCodeClassLoader getOrResolveClassLoader(Collection<URI> requiredFiles
         return SimpleUserCodeClassLoader.create(classLoader);
     }
 
+    @Override
+    public void cacheJobArtifacts(Collection<URI> artifacts) {
+        try {
+            getResolvedUrls(artifacts, blobStore);
+        } catch (IOException | URISyntaxException e) {
+            log.warn("Failed to download artifacts: {}. Trace: {}", artifacts, e);
```
liuml07, line 101: nit: if a Throwable is the last argument of the log line, you don't need a place holder for it.

```
log.warn("Failed to download artifacts: {}. ", artifacts, e);
```

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/ClassLoaderHandle.java
fdc-ntflx: good point. I'm leaving this to a later refactor on our internal `DefaultClassLoaderHandle` where I want to separate it from handling the blobstore. After that I can move this functionality into the new blob store manager.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
fdc-ntflx: Right now manually. will add a write path on another PR so we can move on with the migration

liuml07: I guess this only returns a list for caching, and write path is request handling?

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/CacheJobArtifactsRequest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private final List<URI> artifacts;
+
+    @JsonCreator
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    public CacheJobArtifactsRequest(@JsonProperty("artifacts") List<URI> artifacts) {
+        this.artifacts = artifacts;
+    }
+
+    public List<URI> getArtifacts() {
```
liuml07, line 42: nit: use lombok.Getter for the class?

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/CacheJobArtifactsRequest.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+@EqualsAndHashCode
+@ToString
+public class CacheJobArtifactsRequest implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    private final List<URI> artifacts;
```
liuml07, line 34: nit: `@NonNull`?

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/worker/TaskExecutorGateway.java
Andyz26: maybe just add a document to this API for now?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
Andyz26: In current logic during the upgrade this read will happen to every TE including the ones don't have any cache request. Just seems to be many unnecessary cycles concurrently in IO executor though.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
Andyz26: @liuml07 yeah I am referring to that there is no API to write/update this list.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/CacheJobArtifactsRequest.java
```diff
@@ -0,0 +1,45 @@
...
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonProperty;
+import java.io.Serializable;
+import java.net.URI;
+import java.util.List;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+@EqualsAndHashCode
+@ToString
+public class CacheJobArtifactsRequest implements Serializable {
```
Andyz26, line 30: does this need to be Serializable? what about using "@Data"?

fdc-ntflx: All messages sent over the wire must be serializable in akka (or at least that's what I remember from reading the docs).

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
```diff
@@ -689,6 +691,21 @@ public List<JobArtifact> listJobArtifacts(String name, String version) throws IO
             }).filter(Objects::nonNull).collect(Collectors.toList());
     }
 
+    @Override
+    public void addNewJobArtifactToCache(ClusterID clusterID, ArtifactID artifactID) throws IOException {
+        kvStore.upsert(
+            JOB_ARTIFACTS_TO_CACHE_PER_CLUSTER_ID_NS,
+            clusterID.getResourceID(),
+            clusterID.getResourceID(),
```
Andyz26, line 699: two cluster id?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -393,7 +426,10 @@ private void onTaskExecutorRegistration(TaskExecutorRegistration registration) {
                 updateHeartbeatTimeout(registration.getTaskExecutorID());
             }
             log.info("Successfully registered {} with the resource cluster {}", registration.getTaskExecutorID(), this);
+            self().tell(new CacheJobArtifactsOnTaskExecutorRequest(taskExecutorID, clusterID), self());
```
Andyz26, line 429: no need to trigger this request if no valid cache request?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
fdc-ntflx: added cache.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/CacheJobArtifactsRequest.java
```diff
@@ -0,0 +1,40 @@
...
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonCreator;
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonProperty;
+import java.io.Serializable;
+import java.util.List;
+import lombok.Data;
+import lombok.NonNull;
+
+@Data
+public class CacheJobArtifactsRequest implements Serializable {
```
sundargates, line 28: nit: Can we use @Value instead?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/JobArtifactRouteHandler.java
```diff
@@ -36,4 +37,9 @@ public interface JobArtifactRouteHandler {
      * Search job artifact names by given prefix. Returns only the names for faster lookups.
      */
     CompletionStage<JobArtifactProto.ListJobArtifactsByNameResponse> listArtifactsByName(final JobArtifactProto.ListJobArtifactsByNameRequest request);
+
+    /**
+     * Upsert mapping resource cluster - job artifact to be cached.
+     */
+    CompletionStage<BaseResponse> upsertJobArtifactToCache(JobArtifactProto.JobArtifactToCacheRequest request);
```
sundargates, line 44: Do you think the “which artifacts to cache?” semantics should be under the ResourceCluster component instead of the ResourceCluster route handler endpoint?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/JobArtifactRouteHandlerImpl.java
```diff
@@ -76,4 +77,17 @@ public CompletionStage<JobArtifactProto.UpsertJobArtifactResponse> upsert(JobArt
             }
         });
     }
+
+    @Override
+    public CompletionStage<BaseResponse> upsertJobArtifactToCache(JobArtifactProto.JobArtifactToCacheRequest request) {
+        return CompletableFuture.supplyAsync(() -> {
+            try {
+                mantisStorageProvider.addNewJobArtifactToCache(request.getClusterID(), request.getArtifactID());
```
sundargates, line 85: Do we need to send a message to the resource cluster actor instead and have it deal with the persistence, etc..?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v1/JobArtifactsRoute.java
```diff
@@ -85,6 +86,14 @@ public Route constructRoutes() {
...
                                 () -> get(this::listJobArtifactsByNameRoute)
                             )
+                        ),
+                        // /api/v1/jobArtifacts/cache
+                        path(
+                            "cache",
+                            () -> pathEndOrSingleSlash(
+                                // POST - cache job artifact on resource cluster
+                                () -> post(this::cacheJobArtifactOnResourceCluster)
+                            )
```
sundargates, line 96: I think the ResourceClusterRoute is a more appropriate location for this endpoint.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -100,8 +106,10 @@ class ResourceClusterActor extends AbstractActorWithTimers {
 
     private final ResourceClusterActorMetrics metrics;
 
-    static Props props(final ClusterID clusterID, final Duration heartbeatTimeout, Duration assignmentTimeout, Duration disabledTaskExecutorsCheckInterval, Clock clock, RpcService rpcService, MantisJobStore mantisJobStore, JobMessageRouter jobMessageRouter) {
-        return Props.create(ResourceClusterActor.class, clusterID, heartbeatTimeout, assignmentTimeout, disabledTaskExecutorsCheckInterval, clock, rpcService, mantisJobStore, jobMessageRouter);
+    private final ConcurrentMap<ArtifactID, ArtifactID> jobArtifactsToCache = new ConcurrentHashMap<>();
```
sundargates, line 109: nit: This can be converted to a HashMap I'm guessing after your current set of changes.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -393,7 +428,12 @@ private void onTaskExecutorRegistration(TaskExecutorRegistration registration) {
                 updateHeartbeatTimeout(registration.getTaskExecutorID());
             }
             log.info("Successfully registered {} with the resource cluster {}", registration.getTaskExecutorID(), this);
+            if (!jobArtifactsToCache.isEmpty()) {
```
sundargates, line 431: nit: Can you check if the fast property enabled or not?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
sundargates, line 699: It seems that only one artifact can be stored per resource cluster. Can you please confirm if this is the case and, if so, modify it to allow for multiple artifacts to be cached per resource cluster? Thank you.

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/BlobStoreAwareClassLoaderHandle.java
```diff
@@ -93,6 +95,16 @@ public UserCodeClassLoader getOrResolveClassLoader(Collection<URI> requiredFiles
         return SimpleUserCodeClassLoader.create(classLoader);
     }
 
+    @Override
+    public void cacheJobArtifacts(Collection<ArtifactID> artifacts) {
+        try {
+            final List<URI> artifactURIs = artifacts.stream().map(x -> URI.create(blobStore.getPrefix() + x.getResourceID())).collect(Collectors.toList());
+            getResolvedUrls(artifactURIs, blobStore);
+        } catch (IOException | URISyntaxException e) {
+            log.warn("Failed to download artifacts: {}", artifacts);
```
sundargates, line 104: nice

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/BlobStore.java
```diff
@@ -31,6 +31,8 @@
 public interface BlobStore extends Closeable {
     File get(URI blobUrl) throws IOException;
 
+    default URI getPrefix() { return URI.create(""); };
+
```
sundargates, line 35: I would rather not expose this method as it is more of an implementation detail of a specific blob store, and its unnecessary exposure increases the surface area.

sundargates (approved):
LGTM overall!

PR #450, opened 2023-05-31
Making sure caching works properly in the BlobStore

sundargates:
### Context
When downloading artifacts, we should check if the file exists in the target filesystem instead of the source filesystem. 

### Checklist


fdc-ntflx (approved):

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/HadoopFileSystemBlobStore.java
```diff
@@ -41,11 +41,11 @@ public class HadoopFileSystemBlobStore implements BlobStore {
     public File get(URI blobUrl) throws IOException {
         final Path src = new Path(blobUrl);
         final Path dest = new Path(getStorageLocation(blobUrl));
-        if (!fileSystem.exists(dest)) {
+        File destFile = new File(dest.toUri().getPath());
```
hmitnflx, line 44: nit: having `final` would have been nice!

PR #451, opened 2023-06-01
Add source client connection pool options

Andyz26:
### Context

Add configurable options for the source client connection pool.
Move the default source client to disable the connection pool to avoid memory leak on pool cleanup.

### Checklist

### DIFF
--- a/mantis-runtime/src/main/java/io/mantisrx/runtime/source/http/impl/HttpClientFactories.java
+++ b/mantis-runtime/src/main/java/io/mantisrx/runtime/source/http/impl/HttpClientFactories.java
@@ -41,6 +41,15 @@ public class HttpClientFactories {
         return new SSEClientFactory();
     }
 
+    /**
+     * [Deprecated] read timeouts is not longer supported in SSE client factory.
+     * @param readTimeout
+     */
+    @Deprecated
+    public static HttpClientFactory<ByteBuf, ServerSentEvent> sseClientFactory(int readTimeout) {
+        return new SSEClientFactory(readTimeout);
+    }
+
     public static HttpClientFactory<ByteBuf, ServerSentEvent> sseClientFactory(
         boolean enableConnectionPooling,
         boolean enableIdleConnectionCleanup) {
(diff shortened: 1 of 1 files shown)


mantis-runtime/src/main/java/io/mantisrx/runtime/source/http/impl/HttpClientFactories.java
```diff
@@ -25,67 +25,108 @@
...
+        boolean enableIdleConnectionCleanup) {
+        return new DefaultHttpClientFactory(enableConnectionPooling, enableIdleConnectionCleanup);
+    }
+
     public static HttpClientFactory<ByteBuf, ServerSentEvent> sseClientFactory() {
         return new SSEClientFactory();
     }
 
-    public static HttpClientFactory<ByteBuf, ServerSentEvent> sseClientFactory(int readTimeout) {
-        return new SSEClientFactory(readTimeout);
```
fdc-ntflx, line 40: are you deleting this because not used?

mantis-runtime/src/main/java/io/mantisrx/runtime/source/http/impl/HttpClientFactories.java
```diff
@@ -25,67 +25,108 @@
...
+        boolean enableConnectionPooling,
+        boolean enableIdleConnectionCleanup) {
+        return new SSEClientFactory(enableConnectionPooling, enableIdleConnectionCleanup);
     }
 
-
     private static class DefaultHttpClientFactory implements HttpClientFactory<ByteBuf, ByteBuf> {
 
         private final HttpClient.ClientConfig clientConfig;
+        private final boolean enableConnectionPooling;
```
fdc-ntflx, line 53: do you know what impact enabling/disabling these options will have on the jobs? Any particular side effects we should be aware about?

fdc-ntflx (approved):

mantis-runtime/src/main/java/io/mantisrx/runtime/source/http/impl/HttpClientFactories.java
Andyz26: In general connection pooling helps in certain cases for perf/resource control but in the actual use cases like the one in the source job it's essentially just connecting to a given ip+port per client instance thus the benefit seems insignificant.

PR #453, opened 2023-06-05
Fix double TE registration against RC actor

fdc-ntflx:
### Context

Looks like the TE registers itself twice with the resource cluster actor in rapid succession which could lead to race conditions on caching artifacts.

* The first registration happens during [onStart](https://github.com/Netflix/mantis/blob/058657080e1341667b0fa38fb04b95d8160d560d/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java#L188)
* The second one when ZK registers the new resource cluster (which in the case of first bootstrap of the TE is the same as the one registered during the first registration

In order to avoid double registration for the same RC the High avail service can simply verify whether the RC has actually changed before trying to reconnect to it.

Note: I'm not sure if there are edge cases I'm not considering. I can't think of anything at the moment.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/HighAvailabilityServicesUtil.java
+++ b/mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/HighAvailabilityServicesUtil.java
@@ -15,6 +15,9 @@
  */
 package io.mantisrx.server.master.client;
 
+import io.mantisrx.common.metrics.Counter;
+import io.mantisrx.common.metrics.Metrics;
+import io.mantisrx.common.metrics.MetricsRegistry;
 import io.mantisrx.server.core.CoreConfiguration;
 import io.mantisrx.server.core.master.LocalMasterMonitor;
 import io.mantisrx.server.core.master.MasterDescription;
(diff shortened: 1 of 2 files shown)


mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/HighAvailabilityServicesUtil.java
```diff
@@ -182,9 +182,14 @@ public void register(ResourceLeaderChangeListener<ResourceClusterGateway> change
               .observeOn(scheduler)
               .subscribe(nextDescription -> {
                 log.info("nextDescription={}", nextDescription);
+
+                ResourceClusterGatewayClient newResourceClusterGatewayClient = new ResourceClusterGatewayClient(clusterID, nextDescription);
+                if (newResourceClusterGatewayClient.equals(currentResourceClusterGateway)) {
+                    return;
```
Andyz26, line 188: nit: add metrics on both cases.

Andyz26 (approved):

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/HighAvailabilityServicesUtil.java
```diff
@@ -182,9 +182,14 @@ public void register(ResourceLeaderChangeListener<ResourceClusterGateway> change
               .observeOn(scheduler)
               .subscribe(nextDescription -> {
                 log.info("nextDescription={}", nextDescription);
+
+                ResourceClusterGatewayClient newResourceClusterGatewayClient = new ResourceClusterGatewayClient(clusterID, nextDescription);
```
sundargates, line 186: It would be more appropriate to compare the descriptions rather than the clients themselves in terms of semantics. I’m concerned that if the logic for clients being equal changes in the future, it could cause problems for this logic.

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceClusterGatewayClient.java
```diff
@@ -119,4 +121,13 @@ private AsyncHttpClient buildCloseableHttpClient() {
...
   }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        ResourceClusterGatewayClient gatewayClient = (ResourceClusterGatewayClient) o;
+        return clusterID.equals(gatewayClient.clusterID) &&
+            masterDescription.equals(gatewayClient.masterDescription);
+    }
```
sundargates, line 132: Do you still need this?

PR #455, opened 2023-06-07
replace the metrics in HttpServerListener with Micrometer

7ue5wu:
### Context

Try to replace the metrics part in HttpServerListener with Micrometer metrics and wanna get some early feedback and advice.

Explanation:
In the previous implementation, we have the concepts GroupId(name, tags) and MetricId(groupname, metricname, tags).
However, in the Micrometer, it cannot name the Metrics with GroupId, so I try to add the group name as a prefix in metric name and connect them with the underscore. 
If we need the group name or the metric(event) name when publishing, we can extract it from the whole metric name.
And we can still use them to form a GroupId(groupname + tags). and MetricId(metricName + tag) if needed.

### Checklist

### DIFF
--- a/mantis-common/build.gradle
+++ b/mantis-common/build.gradle
@@ -41,6 +41,7 @@ dependencies {
     api libraries.slf4jLog4j12
     implementation libraries.commonsIo
     implementation 'net.jcip:jcip-annotations:1.0'
+    implementation 'io.micrometer:micrometer-core:1.11.0'
 
     testImplementation libraries.spectatorApi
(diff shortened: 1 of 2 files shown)


Andyz26:
@hmit I suppose these change cannot be merged in master directly as it will break the dashboard etc. once picked up by nfmantis? Shall we keep all metrics changes to a feature branch for now?

hmitnflx:
@Andyz26 this is an initial PR to get early feedback on implementation details for micrometer migration. The other piece to make the metrics inter-operational is still being worked out.

PR #462, opened 2023-06-14
Send heartbeat failure before shutting down TE

fdc-ntflx:
### Context

The purpose of this pull request (PR) is to ensure that a worker is promptly resubmitted in two specific situations: when a shutdown signal is detected by the Task Executor (TE) and when an exception occurs before the job enters the Launched state.

The TE can receive shutdown signals in various ways, with one well-known example being during Titus container migration. During container migration, Titus sends a sigkill signal to the running container, allowing it time to gracefully shut down. It is crucial for the TE to communicate with the master in a timely manner, informing it of the need to reschedule the worker due to the impending migration.

Similarly, if an exception is raised during task preparation, the TE must ask the master to reschedule the TE in another container. One such case occurs when caching artifacts fails due to the download of an artifact for a new job request.

Currently, without this change, TEs affected by these two issues are only resubmitted after a heartbeat timeout, which can take up to 15 minutes. However, with this PR, the resubmission for these specific cases will only take a few seconds.

### Checklist

### DIFF
--- a/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
+++ b/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
@@ -23,6 +23,7 @@ import io.mantisrx.common.JsonSerializer;
 import io.mantisrx.common.WorkerPorts;
 import io.mantisrx.common.metrics.netty.MantisNettyEventsListenerFactory;
 import io.mantisrx.runtime.MachineDefinition;
+import io.mantisrx.runtime.MantisJobState;
 import io.mantisrx.runtime.loader.ClassLoaderHandle;
 import io.mantisrx.runtime.loader.RuntimeTask;
 import io.mantisrx.runtime.loader.SinkSubscriptionStateHandler;
@@ -109,12 +110,14 @@ public class TaskExecutor extends RpcEndpoint implements TaskExecutorGateway {
     // represents the current connection to the resource manager.
     private ResourceManagerGatewayCxn currentResourceManagerCxn;
     private TaskExecutorReport currentReport;
-    private RuntimeTask currentTask;
     private Subscription currentTaskStatusSubscription;
     private int resourceManagerCxnIdx;
     private Throwable previousFailure;
 
     private final TaskFactory taskFactory;
+
+    private RuntimeTask currentTask;
+    private ExecuteStageRequest currentRequest;
 
     public TaskExecutor(
         RpcService rpcService,
@@ -481,6 +484,8 @@ public class TaskExecutor extends RpcEndpoint implements TaskExecutorGateway {
 
     private void prepareTask(WrappedExecuteStageRequest wrappedRequest) {
         try {
+            this.currentRequest = wrappedRequest.getRequest();
+
             UserCodeClassLoader userCodeClassLoader = this.taskFactory.getUserCodeClassLoader(
                 wrappedRequest.getRequest(), classLoaderHandle);
             ClassLoader cl = userCodeClassLoader.asClassLoader();
@@ -501,6 +506,10 @@ public class TaskExecutor extends RpcEndpoint implements TaskExecutorGateway {
             }, 0, TimeUnit.MILLISECONDS);
         } catch (Exception ex) {
             log.error("Failed to submit task, request: {}", wrappedRequest.getRequest(), ex);
+            final Status failedStatus = new Status(currentRequest.getJobId(), currentRequest.getStage(), currentRequest.getWorkerIndex(), currentRequest.getWorkerNumber(),
+                Status.TYPE.INFO, "stage " + currentRequest.getStage() + " worker index=" + currentRequest.getWorkerIndex() + " number=" + currentRequest.getWorkerNumber() + " failed during initialization",
+                MantisJobState.Failed);
+            updateExecutionStatus(failedStatus);
             listeners.enqueue(getTaskFailedEvent(null, ex));
         }
         finally {
--- a/mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/Heartbeat.java
+++ b/mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/Heartbeat.java
@@ -28,6 +28,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.LinkedBlockingQueue;
 import javax.annotation.Nullable;
+import lombok.Getter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -35,9 +36,13 @@ import org.slf4j.LoggerFactory;
 class Heartbeat {
 
     private static final Logger logger = LoggerFactory.getLogger(Heartbeat.class);
+    @Getter
     private final String jobId;
+    @Getter
     private final int stageNumber;
+    @Getter
     private final int workerIndex;
+    @Getter
     private final int workerNumber;
     private final ConcurrentMap<String, String> payloads;
     private final BlockingQueue<PayloadPair> singleUsePayloads = new LinkedBlockingQueue<>();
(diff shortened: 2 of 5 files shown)


mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/Heartbeat.java
```diff
@@ -71,7 +71,11 @@ void addSingleUsePayload(String name, String value) {
         singleUsePayloads.offer(new PayloadPair(name, value));
     }
 
-    Status getCurrentHeartbeatStatus() {
+    private String getWorkerStringPrefix(int stageNum, int index, int number) {
+        return "stage " + stageNum + " worker index=" + index + " number=" + number;
```
Andyz26, line 75: nit: string.format

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/WorkerExecutionOperationsNetworkStage.java
```diff
@@ -622,7 +627,9 @@ private WorkerConsumer connectToObservableAtPreviousStages(Observable<JobSchedul
 
     @Override
     public void shutdownStage() throws IOException  {
-        logger.info("Shutdown initiated");
+        if (jobStatus != null) {
+            jobStatus.onNext(heartbeatRef.get().getFailedHeartbeatStatus());
```
Andyz26, line 631: should shutdown always reflect as failed? what about regular shutdown?

fdc-ntflx: Good question. I wasn't able to split the case failure vs regular shutdown which frankly would be ideal, however sending this signal to master during regular shutdown won't do anything because the worker has already been archived. The job actor will simply log `Job XYZ Terminal event from Unknown worker ZYX. Ignoring`.

sundargates, line 631: Wouldn’t it be cleaner to create the Status directly instead of altering the Heartbeat data structure?

fdc-ntflx: That's a fair ask. Let me explain why I implemented this in such way. 2 reasons mainly:
1. There is no reference to worker metadata in this class aside from within heartbeat instance.
2. I imagined heartbeat class to be more a status generator instead where heartbeat is "only a specific  type" but in such case we should probably rename the class.

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
```diff
@@ -501,6 +505,10 @@ private void prepareTask(WrappedExecuteStageRequest wrappedRequest) {
             }, 0, TimeUnit.MILLISECONDS);
         } catch (Exception ex) {
             log.error("Failed to submit task, request: {}", wrappedRequest.getRequest(), ex);
+            final Status failedStatus = new Status(currentRequest.getJobId(), currentRequest.getStage(), currentRequest.getWorkerIndex(), currentRequest.getWorkerNumber(),
+                Status.TYPE.INFO, "stage " + currentRequest.getStage() + " worker index=" + currentRequest.getWorkerIndex() + " number=" + currentRequest.getWorkerNumber() + " failed during initialization",
```
sundargates, line 509: nit: String.format for the message itself. Also, could you create the message pattern as a constant? This way, the message is clear.

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
```diff
@@ -116,6 +117,8 @@ public class TaskExecutor extends RpcEndpoint implements TaskExecutorGateway {
 
     private final TaskFactory taskFactory;
 
+    private ExecuteStageRequest currentRequest;
```
sundargates, line 120: nit: Move this below `RuntimeTask currentTask;` This way, all the state about the current runtime task is captured clearly.

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/WorkerExecutionOperationsNetworkStage.java
```diff
@@ -479,7 +484,7 @@ private void setupSubscriptionStateHandler(ExecuteStageRequest executeStageReque
     }
 
     private String getWorkerStringPrefix(int stageNum, int index, int number) {
-        return "stage " + stageNum + " worker index=" + index + " number=" + number;
+        return String.format("stage=%s worker index=%s number=%s", stageNum, index, number);
```
hmitnflx, line 487: the format is being changed here,
original would render `getWorkerStringPrefix(1,1,1)`
as `stage 1 worker index=1 number=1`
new would be `stage=1 worker index=1 number=1`

I hope nobody is parsing this to determine worker properties...

fdc-ntflx: Good catch!

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/TaskExecutor.java
```diff
@@ -451,6 +454,7 @@ <T> CompletableFuture<T> callInMainThread(Callable<CompletableFuture<T>> tSuppli
     public CompletableFuture<Ack> submitTask(ExecuteStageRequest request) {
 
         log.info("Received request {} for execution", request);
+        this.currentRequest = request;
```
sundargates, line 457: nit: I would also move this to prepareTask() since that's where the rest of the business logic is handled with respect to the state changes.

hmitnflx (approved):
lgtm

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/WorkerExecutionOperationsNetworkStage.java
```diff
@@ -106,6 +106,7 @@ public class WorkerExecutionOperationsNetworkStage implements WorkerExecutionOpe
     private final List<Closeable> closeables = new ArrayList<>();
     private final ScheduledExecutorService scheduledExecutorService;
     private final ClassLoader classLoader;
+    private Observer<Status> jobStatus;
```
sundargates, line 109: nit: jobStatusObserver/jobStatusListener.

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/WorkerExecutionOperationsNetworkStage.java
```diff
@@ -622,7 +627,13 @@ private WorkerConsumer connectToObservableAtPreviousStages(Observable<JobSchedul
 
     @Override
     public void shutdownStage() throws IOException  {
-        logger.info("Shutdown initiated");
+        if (jobStatus != null) {
+            final Heartbeat heartbeat = heartbeatRef.get();
+            final Status status = new Status(heartbeat.getJobId(), heartbeat.getStageNumber(), heartbeat.getWorkerIndex(), heartbeat.getWorkerNumber(),
+                Status.TYPE.INFO, getWorkerStringPrefix(heartbeat.getStageNumber(), heartbeat.getWorkerIndex(), heartbeat.getWorkerNumber()) + " shutdown",
```
sundargates, line 633: nit: It would be nice to put all these message patterns in the same place.

sundargates (approved):

sundargates:
Overall, looks good! When testing, please confirm the behavior when the worker is being asked to shutdown from the control plane.

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/WorkerExecutionOperationsNetworkStage.java
fdc-ntflx: +1

PR #467, opened 2023-06-28
Adding capability to compute and emit the costs of Mantis Jobs in mantis API

sundargates:
### Context

We need the capability to compute the costs of Mantis jobs and emit them as part of the Mantis Job Metadata. This PR adds support for that. 

### Checklist


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/Costs.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package io.mantisrx.server.master.domain;
+
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonIgnore;
+import lombok.Value;
+
+@Value
+public class Costs {
```
sundargates, line 23: Add more comments to this.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/config/MasterConfiguration.java
```diff
@@ -360,6 +360,10 @@ default Duration getSchedulerIntervalBetweenRetries() {
     @Default("60000") // 1 minute
     int getAssignmentIntervalInMs();
 
+    @Config("mantis.job.costsCalculator.class")
+    @Default("io.mantisrx.master.jobcluster.job.CostsCalculator")
```
sundargates, line 364: Change this default value.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/config/MantisExtensionFactory.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.server.master.config;
+
+import java.util.Properties;
+import lombok.val;
+
+public class MantisExtensionFactory {
```
sundargates, line 22: Add documentation

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/config/MantisExtension.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.server.master.config;
+
+import java.util.Properties;
+
+@FunctionalInterface
+public interface MantisExtension<T> {
```
sundargates, line 22: Add documentation

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/config/MantisExtensionFactory.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @SuppressWarnings({"unchecked"})
+    public static <T extends MantisExtension> T createObject(String className, Properties properties) throws Exception {
+        val extension = (MantisExtension<T>) Class.forName(className).newInstance();
```
calvin681, line 26: why do we use lombok val? doesn't seem necessary.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/Costs.java
calvin681: Is this class actually necessary? why do we need our own wrapper around a double?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/config/MasterConfiguration.java
calvin681: Should this value be "CostsCalculatorFactory"? All these layers of abstraction make it really confusing. I'm not seeing the value. Can we just have a CostsCalculatorFactory interface and ask for an impl class in the config?

sundargates: Fair enough. I'll get rid of it. I was hoping to have some structure around how we initialize extendable components. The other diff (https://github.com/Netflix/mantis/pull/319/files) uses this idea/interface for a number of components (search for MantisExtension). But if you think having this defined in documentation/code is better than having an explicit interface, we can eliminate it.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/Costs.java
sundargates: Today, this covers the total cost. Imagine we need to break this down further into cpuCosts, memCosts, diskCosts, etc... Having this value class will help wrap all that without adding more fields to the JobMetadata.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/config/MantisExtensionFactory.java
sundargates: Sure, I can get rid of it.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/Costs.java
calvin681: OK, that's a fair point.

calvin681 (approved):

PR #486, opened 2023-07-19
RC Scaler: Delta Calculation instead of Desired Cluster Size

fdc-ntflx:
### Context

The intention of this PR is to modify the functionality of the RC scaler. Instead of determining the desired cluster size, the proposed change suggests that it should calculate the delta based on the available nodes. The provider implementation would then determine the final desired cluster size.

The primary motivation behind this change is that the scaler currently lacks an accurate perception of the ASG size. Specifically, the total worker count doesn't account for ghost workers, resulting in a computed cluster size that is smaller than the actual size. Let me provide an example to illustrate this:

- Current cluster size: 100
- Number of ghost workers: 10
- Number of available workers: 0
- Cluster scale rule, minimum idle count: 10

Based on these figures, we anticipate the scaler to adjust the cluster to a desired size of 110. However, that is not the case. Let's examine why:

- Cluster size observed by the scaler: 90
- Computed desired size: current_cluster_size + min_idle_count = 90 + 10 = 100

The computed desired cluster size is 100, which coincides with the current size. Consequently, no changes will be applied to the cluster.

By allowing the provider to determine the actual cluster size based on the delta (in this instance, 10), the provider has the ability to fetch the actual ASG state and accurately update its desired size.

While resolving the issue of ghost workers and being able to replenish them is undoubtedly necessary, it is possible that the replacement of a worker using this mechanism would take several minutes, rendering autoscaling ineffective during this period. Therefore, I believe there is merit in pursuing this proposal even after addressing the ghost workers issue.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/GetClusterIdleInstancesRequest.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/GetClusterIdleInstancesRequest.java
@@ -28,5 +28,4 @@ public class GetClusterIdleInstancesRequest {
     ContainerSkuID skuId;
 
     int maxInstanceCount;
-    int desireSize;
 }
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/GetClusterIdleInstancesRequest.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/GetClusterIdleInstancesRequest.java
@@ -26,6 +26,8 @@ import lombok.Value;
 public class GetClusterIdleInstancesRequest {
     ClusterID clusterID;
     ContainerSkuID skuId;
-
-    int maxInstanceCount;
+    int scaleUpCount;
+    int scaleDownCount;
+    int registeredCount;
+    int idleCount;
 }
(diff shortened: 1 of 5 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/GetClusterIdleInstancesRequest.java
```diff
@@ -26,7 +26,8 @@
 public class GetClusterIdleInstancesRequest {
     ClusterID clusterID;
     ContainerSkuID skuId;
-
-    int maxInstanceCount;
-    int desireSize;
+    int scaleUpCount;
```
Andyz26, line 29: why do we need scaleUp count when fetching idle list (to scale down)?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/GetClusterIdleInstancesRequest.java
```diff
@@ -26,7 +26,8 @@
...
+    int scaleDownCount;
+    int registeredCount;
```
Andyz26, line 31: dito

sundargates:
I wonder if we need to complicate our scaler approach because of the problem of orphaned instances. I recommend we work on solving that problem first. We can take a pulse check on this approach after that issue is resolved. If even after solving the issue of ghost instances, we see some gaps, we can look into this IMO.

PR #490, opened 2023-07-25
[WIP] Add disable by TE id endpoint

fdc-ntflx:
### Context

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
@@ -156,7 +156,7 @@ public interface ResourceCluster extends ResourceClusterGateway {
      * @param taskExecutorID ID whom task executor will be disabled.
      * @return a future that completes when the underlying operation is registered by the system
      */
-    CompletableFuture<Ack> disableTaskExecutor(ClusterID clusterID, TaskExecutorID taskExecutorID);
+    CompletableFuture<TaskExecutorOnDisabled> disableTaskExecutor(ClusterID clusterID, TaskExecutorID taskExecutorID);
 
     /**
      * Enables/Disables scaler for a given skuID of a given clusterID
@@ -236,5 +236,10 @@ public interface ResourceCluster extends ResourceClusterGateway {
         @Nullable
         WorkerId workerId;
         long lastHeartbeatInMs;
+    }
+
+    @Value
+    class TaskExecutorOnDisabled {
+        boolean isDisabled;
     }
 }
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
@@ -156,7 +156,7 @@ public interface ResourceCluster extends ResourceClusterGateway {
      * @param taskExecutorID ID whom task executor will be disabled.
      * @return a future that completes when the underlying operation is registered by the system
      */
-    CompletableFuture<Ack> disableTaskExecutor(TaskExecutorID taskExecutorID);
+    CompletableFuture<Ack> disableTaskExecutor(ClusterID clusterID, TaskExecutorID taskExecutorID);
 
     /**
      * Enables/Disables scaler for a given skuID of a given clusterID
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
@@ -156,7 +156,7 @@ public interface ResourceCluster extends ResourceClusterGateway {
      * @param taskExecutorID ID whom task executor will be disabled.
      * @return a future that completes when the underlying operation is registered by the system
      */
-    CompletableFuture<TaskExecutorOnDisabled> disableTaskExecutor(ClusterID clusterID, TaskExecutorID taskExecutorID);
+    CompletableFuture<TaskExecutorOnDisabled> disableTaskExecutor(TaskExecutorID taskExecutorID);
 
     /**
      * Enables/Disables scaler for a given skuID of a given clusterID
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
@@ -151,6 +151,14 @@ public interface ResourceCluster extends ResourceClusterGateway {
     CompletableFuture<Ack> disableTaskExecutorsFor(Map<String, String> attributes, Instant expiry);
 
     /**
+     * Disables task executor that match the ID passed
+     *
+     * @param taskExecutorID ID whom task executor will be disabled.
+     * @return a future that completes when the underlying operation is registered by the system
+     */
+    CompletableFuture<Ack> disableTaskExecutor(TaskExecutorID taskExecutorID);
+
+    /**
      * Enables/Disables scaler for a given skuID of a given clusterID
      *
      * @param skuID   skuID whom scaler will be enabled/disabled.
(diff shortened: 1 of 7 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -192,6 +195,8 @@ public Receive createReceive() {
                 .match(ResourceOverviewRequest.class, this::onResourceOverviewRequest)
                 .match(TaskExecutorInfoRequest.class, this::onTaskExecutorInfoRequest)
                 .match(TaskExecutorGatewayRequest.class, this::onTaskExecutorGatewayRequest)
+                .match(DisableTaskExecutorRequest.class, this::onDisableTaskExecutorRequest)
+                .match(DeleteDisabledTaskExecutorRequest.class, this::onDeleteDisabledTaskExecutorRequest)
```
Andyz26, line 199: who/how is this action used?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -141,6 +143,7 @@ static Props props(final ClusterID clusterID, final Duration heartbeatTimeout, D
     public void preStart() throws Exception {
         super.preStart();
         fetchJobArtifactsToCache();
+        fetchDisabledTaskExecutors();
```
Andyz26, line 146: how does the current disable TE on attributes get persisted? can these be merged?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
fdc-ntflx: Titus Migration logic will call `onDisableTaskExecutorRequest` while the reaper (yet to be implemented) will call `onDeleteDisabledTaskExecutorRequest` after killing the container using spinnaker API.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
fdc-ntflx: Great question. There is another set called `activeDisableTaskExecutorsRequests` which stores disable TE requests  by attributes. Although it can be merged with the new set I'm proposing, however they seem to be fulfilling a slightly difference goal. The existing map is for ASG updates, while the new one is for containers migration. While for the existing set we expect very few requests and it's fine to have an expiration time, for the new one we can have hundreds of simultaneous requests and we only want to stored them until an external signal reaches the control plane (so we don't really need the expiration logic at all. 

I believe long-term we should move to a structure more similar to the one I'm proposing anyway (related to the discussion we had today).

That said I'm open to merge existing and new sets.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
Andyz26: "onDeleteDisabledTaskExecutorRequest": what does this do in RCActor?

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
```diff
@@ -150,6 +150,19 @@ CompletableFuture<TaskExecutorID> getTaskExecutorFor(
...
+     *
+     * @param taskExecutorID ID whom task executor will be disabled.
+     * @return a future that completes when the underlying operation is registered by the system
+     */
+    CompletableFuture<TaskExecutorOnDisabled> disableTaskExecutor(TaskExecutorID taskExecutorID);
+
+    /**
+     * TODO:
+     */
+    CompletableFuture<Ack> deleteDisabledTaskExecutor(TaskExecutorID taskExecutorID);
```
sundargates, line 164: Can we use the expiry to get rid of these requests? It feels more elegant.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
```diff
@@ -150,6 +150,19 @@ CompletableFuture<TaskExecutorID> getTaskExecutorFor(
      */
     CompletableFuture<Ack> disableTaskExecutorsFor(Map<String, String> attributes, Instant expiry);
 
+    /**
+     * Disables task executor that match the ID passed
+     *
+     * @param taskExecutorID ID whom task executor will be disabled.
+     * @return a future that completes when the underlying operation is registered by the system
+     */
+    CompletableFuture<TaskExecutorOnDisabled> disableTaskExecutor(TaskExecutorID taskExecutorID);
```
sundargates, line 159: nit: Ack is more consistent with the rest of the requests.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -395,6 +406,42 @@ private void findAndMarkDisabledTaskExecutors(CheckDisabledTaskExecutors r) {
...
+
+            final TaskExecutorState state = executorStateManager.get(request.getTaskExecutorID());
+            log.info("Trying to disable task executor {} with state {}", request.getTaskExecutorID(), state);
+
+            if (state == null) {
+                // If the TE is unknown by mantis, delete it from state
+                mantisJobStore.deleteDisabledTaskExecutor(request.getTaskExecutorID());
+                disabledTaskExecutors.remove(request.taskExecutorID);
+                sender().tell(new TaskExecutorOnDisabled(false), self());
+            } else {
```
sundargates, line 431: What if the task executor registers after this message gets processed? what should the semantics be in that case?

sundargates: It appears that we have different semantics for this than when the disabled request is for an ASG.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/IMantisPersistenceProvider.java
```diff
@@ -165,6 +165,12 @@ void storeAndUpdateWorkers(final IMantisWorkerMetadata existingWorker, final IMa
 
     List<DisableTaskExecutorsRequest> loadAllDisableTaskExecutorsRequests(ClusterID clusterID) throws IOException;
 
+    void storeNewDisabledTaskExecutor(TaskExecutorID taskExecutorID) throws IOException;
+
+    void deleteDisabledTaskExecutor(TaskExecutorID taskExecutorID) throws IOException;
+
+    List<TaskExecutorID> loadAllDisabledTaskExecutors(ClusterID clusterID) throws IOException;
+
```
sundargates, line 173: At the storage layer, we can use the existing methods to keep this simple and reduce the surface area.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v1/ResourceClustersNonLeaderRedirectRoute.java
```diff
@@ -83,6 +83,7 @@
  * /api/v1/resourceClusters/{}/scaleSku                               (POST)
  * /api/v1/resourceClusters/{}/upgrade                                (POST)
  * /api/v1/resourceClusters/{}/disableTaskExecutors                   (POST)
+ * /api/v1/resourceClusters/{}/disableTaskExecutors/{}                (POST)
```
sundargates, line 86: Semantically, is this consistent with the REST conventions?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
sundargates: Could we use the same model for external and storage purposes but store them differently internally for processing purposes? This way, we get the best of both worlds. We will expose less surface area and fewer endpoints they need to think about and manage while we get the efficiency of having two different models.

sundargates (approved):

fdc-ntflx:
Closed in favor of : https://github.com/Netflix/mantis/pull/493

PR #501, opened 2023-07-29
Support TE generation in scheduler 

Andyz26:
### Context

Introduce new metadata "generation" in the task executor state so that the scheduler will always try to use a TE on newer generation during scheduling. Thus during any TE upgrade/migration, the newly deployed TEs will get the migrated-out assignments and avoid repeated migration if these migrated-out workers were assigned to other older generation TEs that would get upgraded shortly after.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManager.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManager.java
@@ -35,9 +35,23 @@ import org.apache.commons.lang3.tuple.Pair;
  * A component to manage the states of {@link TaskExecutorState} for a given {@link ResourceClusterActor}.
  */
 public interface ExecutorStateManager {
-    void putIfAbsent(TaskExecutorID taskExecutorID, TaskExecutorState state);
-    void markAvailable(TaskExecutorID taskExecutorID);
-    void markUnavailable(TaskExecutorID taskExecutorID);
+    /**
+     * Store and track the given task executor's state inside this {@link ExecutorStateManager} if there is no existing
+     * state already. Ignore the given state instance if there is already a state associated with the given ID.
+     * @param taskExecutorID TaskExecutorID
+     * @param state new task executor state
+     * @return whether the given task executor becomes available.
+     */
+    boolean onTaskExecutorStateAssigned(TaskExecutorID taskExecutorID, TaskExecutorState state);
+
+    /**
+     * Try to mark the given task executor as available if its tracked state is available.
+     * @param taskExecutorID TaskExecutorID
+     * @return whether the given task executor becomes available.
+     */
+    boolean tryMarkAvailable(TaskExecutorID taskExecutorID);
+
+    void tryMarkUnavailable(TaskExecutorID taskExecutorID);
 
     @Nullable
     TaskExecutorState get(TaskExecutorID taskExecutorID);
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
@@ -16,6 +16,7 @@
 
 package io.mantisrx.master.resourcecluster;
 
+import io.mantisrx.common.WorkerConstants;
 import io.mantisrx.master.resourcecluster.ResourceClusterActor.GetActiveJobsRequest;
 import io.mantisrx.master.resourcecluster.ResourceClusterActor.GetClusterUsageRequest;
 import io.mantisrx.master.resourcecluster.ResourceClusterActor.TaskExecutorAssignmentRequest;
@@ -27,21 +28,26 @@ import io.mantisrx.server.core.domain.WorkerId;
 import io.mantisrx.server.master.resourcecluster.ContainerSkuID;
 import io.mantisrx.server.master.resourcecluster.ResourceCluster.ResourceOverview;
 import io.mantisrx.server.master.resourcecluster.TaskExecutorID;
+import io.mantisrx.server.master.resourcecluster.TaskExecutorRegistration;
 import io.mantisrx.shaded.com.google.common.cache.Cache;
 import io.mantisrx.shaded.com.google.common.cache.CacheBuilder;
+import java.util.Comparator;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.NavigableSet;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.SortedMap;
+import java.util.TreeSet;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
+import lombok.Builder;
+import lombok.Value;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.tuple.Pair;
 
@@ -53,7 +59,7 @@ public class ExecutorStateManagerImpl implements ExecutorStateManager {
      * Cache the available executors ready to accept assignments. Note these executors' state are not strongly
      * synchronized and requires state level check when matching.
      */
-    private final SortedMap<Double, Set<TaskExecutorID>> executorByCores = new ConcurrentSkipListMap<>();
+    private final SortedMap<Double, NavigableSet<TaskExecutorHolder>> executorByCores = new ConcurrentSkipListMap<>();
 
     private final Cache<TaskExecutorID, TaskExecutorState> archivedState = CacheBuilder.newBuilder()
         .maximumSize(10000)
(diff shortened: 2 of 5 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManager.java
```diff
@@ -35,9 +35,23 @@
...
-    void markAvailable(TaskExecutorID taskExecutorID);
-    void markUnavailable(TaskExecutorID taskExecutorID);
+    /**
+     * Store and track the given task executor's state inside this {@link ExecutorStateManager} if there is no existing
+     * state already. Ignore the given state instance if there is already a state associated with the given ID.
+     * @param taskExecutorID TaskExecutorID
+     * @param state new task executor state
+     * @return whether the given task executor becomes available.
+     */
+    boolean onTaskExecutorStateAssigned(TaskExecutorID taskExecutorID, TaskExecutorState state);
```
fdc-ntflx, line 45: is the return value ever used?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -63,55 +69,63 @@ public class ExecutorStateManagerImpl implements ExecutorStateManager {
...
-    public void markAvailable(TaskExecutorID taskExecutorID) {
+    public boolean tryMarkAvailable(TaskExecutorID taskExecutorID) {
         if (!this.taskExecutorStateMap.containsKey(taskExecutorID)) {
             log.warn("marking invalid executor as available: {}", taskExecutorID);
-            return;
+            return false;
         }
 
         TaskExecutorState taskExecutorState = this.taskExecutorStateMap.get(taskExecutorID);
-        if (taskExecutorState.getRegistration() == null) {
```
fdc-ntflx, line 94: It looks like with the old logic we would allow TEs with non-available state to be added to `executorByCores`. I'm wondering why.

fdc-ntflx (approved):
Overall lgtm!

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManager.java
Andyz26: No. It's a refactor on this API return type from void so that it makes more sense to return a bool about whether the operation is successful now after the logic change but the usages currently don't need to use this state.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManager.java
```diff
@@ -35,9 +35,27 @@
  * A component to manage the states of {@link TaskExecutorState} for a given {@link ResourceClusterActor}.
  */
 public interface ExecutorStateManager {
```
sundargates, line 37: nit: this interface need not be public.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManager.java
```diff
@@ -35,9 +35,27 @@
...
+     * @param taskExecutorID TaskExecutorID
+     * @return whether the given task executor becomes available.
+     */
+    boolean tryMarkAvailable(TaskExecutorID taskExecutorID);
+
+    /**
+     * Try to mark the given task executor as unavailable.
+     * @param taskExecutorID TaskExecutorID
+     */
+    void tryMarkUnavailable(TaskExecutorID taskExecutorID);
```
sundargates, line 58: nit: why not return Boolean in this case too?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManager.java
```diff
@@ -35,9 +35,27 @@
...
-    void markAvailable(TaskExecutorID taskExecutorID);
-    void markUnavailable(TaskExecutorID taskExecutorID);
+    /**
+     * Store and track the given task executor's state inside this {@link ExecutorStateManager} if there is no existing
+     * state already. Ignore the given state instance if there is already a state associated with the given ID.
+     * @param taskExecutorID TaskExecutorID
+     * @param state new task executor state
+     * @return whether the given task executor becomes available.
+     */
+    boolean onTaskExecutorStateAssigned(TaskExecutorID taskExecutorID, TaskExecutorState state);
```
sundargates, line 45: nit: onTaskExecutorAssigned

What's the use for the boolean return value in this case?

Andyz26: just to be consistent as the state change triggers the TE update in the executor manager.

mantis-control-plane/mantis-control-plane-server/src/test/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerTests.java
```diff
@@ -181,11 +189,162 @@ public void testGetBestFit() {
...
+        stateManager.tryMarkUnavailable(TASK_EXECUTOR_ID_3);
         bestFitO =
             stateManager.findBestFit(new TaskExecutorAssignmentRequest(
                 TaskExecutorAllocationRequest.of(WORKER_ID, MACHINE_DEFINITION_2), CLUSTER_ID));
 
         assertFalse(bestFitO.isPresent());
     }
+
+    @Test
+    public void testTaskExecutorHolderCreation() {
```
sundargates, line 201: Thank you for adding the tests!

sundargates (approved):
LGTM!

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -63,62 +69,71 @@ public class ExecutorStateManagerImpl implements ExecutorStateManager {
...
+    public void trackIfAbsent(TaskExecutorID taskExecutorID, TaskExecutorState state) {
         this.taskExecutorStateMap.putIfAbsent(taskExecutorID, state);
         if (this.archivedState.getIfPresent(taskExecutorID) != null) {
             log.info("Reviving archived executor: {}", taskExecutorID);
             this.archivedState.invalidate(taskExecutorID);
         }
 
-        // add to buckets. however new executors won't have valid registration at this moment and requires marking
-        // again later when registration is ready.
+        tryMarkAvailable(taskExecutorID, state);
```
sundargates, line 79: nit:
I would change this code as below.
```
if (state.isAvailable()) {
  Preconditions.checkState(markAsAvailable());
}
```

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -63,62 +69,71 @@ public class ExecutorStateManagerImpl implements ExecutorStateManager {
...
+
+    /**
+     * Add to buckets. however new executors won't have valid registration at this moment and requires marking
+     * again later when registration is ready.
+     * @param taskExecutorID taskExecutorID
+     * @param state state
+     * @return whether the target executor is marked as available.
+     */
+    public boolean tryMarkAvailable(TaskExecutorID taskExecutorID, TaskExecutorState state) {
         if (state.isAvailable() && state.getRegistration() != null) {
```
sundargates, line 90: Semantically, it does not make sense that we are calling this method to make the task executor available, and the state could in fact be non-available. I would probably do the checks from the caller.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -63,62 +69,71 @@ public class ExecutorStateManagerImpl implements ExecutorStateManager {
...
-            log.info("Marking executor {} as available for matching.", taskExecutorID);
+            TaskExecutorHolder teHolder = TaskExecutorHolder.of(taskExecutorID, state.getRegistration());
```
sundargates, line 91: I don't think we need this new DS. We should use the TaskExecutorState with a different comparator.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
sundargates, line 90: These checks are called happening as part of the caller today. Look at usages of ExecutorStageManager::markAvailable in the ResourceClusterActor. Doing these checks twice seems unnecessary.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManager.java
```diff
@@ -34,10 +34,27 @@
...
+     * @param state new task executor state
+     */
+    void trackIfAbsent(TaskExecutorID taskExecutorID, TaskExecutorState state);
+
+    /**
+     * Try to mark the given task executor as available if its tracked state is available.
+     * @param taskExecutorID TaskExecutorID
+     * @return whether the given task executor becomes available.
+     */
+    boolean tryMarkAvailable(TaskExecutorID taskExecutorID);
```
sundargates, line 51: If you look at TaskExecutorState and its use, it looks like this class is designed as a listener entity. Hence, the methods do not return any output and are designed as effects.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -53,7 +59,7 @@ public class ExecutorStateManagerImpl implements ExecutorStateManager {
      * Cache the available executors ready to accept assignments. Note these executors' state are not strongly
      * synchronized and requires state level check when matching.
      */
-    private final SortedMap<Double, Set<TaskExecutorID>> executorByCores = new ConcurrentSkipListMap<>();
+    private final SortedMap<Double, NavigableSet<TaskExecutorHolder>> executorByCores = new ConcurrentSkipListMap<>();
```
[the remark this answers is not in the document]
Andyz26: TaskExecutorState has no id in it. There is another map to associate TE id -> TE state.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
Andyz26: TaskExecutorState has no ID in it. In the current state inside this ExecutorStateManagerImpl there are tracked on TE id -> its state separately and I think tracking ID + gen here should be enough and there is no strong benefit to merge states in.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
Andyz26: I am trying to treat ExecutorStateManagerImpl as a sort of black box thus checks inside are more of a guard to future changes/usages if it gets used somewhere else.

Andyz26: let me switch this to private. Should be used inside the TaskExecutorImpl only.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
Andyz26: we don't need to fail if the given state is not available yet right? (checkState throws exception)

PR #503, opened 2023-07-31
Make default: inherit worker count from previous job if exists

hmitnflx:
### Context

When launching a new job, default to worker count from previous job if one exists.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobDefinition.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobDefinition.java
@@ -133,12 +133,13 @@ public class JobDefinition {
     }
 
     public boolean requireInheritInstanceCheck() {
-        return this.schedulingInfo != null && this.deploymentStrategy != null && this.getDeploymentStrategy().requireInheritInstanceCheck();
+        return this.schedulingInfo != null &&
+            (this.deploymentStrategy == null || this.getDeploymentStrategy().requireInheritInstanceCheck());
     }
 
     public boolean requireInheritInstanceCheck(int stageNum) {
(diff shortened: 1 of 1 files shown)


Andyz26:
I wonder if there could be breaking cases on multi-instance jobCluster e.g. job cluster running multiple jobs with auto-scaler and new job instance used a huge auto-scaled job instance count instead of the schedulingInfo spec count?

sundargates (commented):
Can we add some tests to codify these behaviors?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobDefinition.java
```diff
@@ -133,12 +133,13 @@ public void validate(boolean schedulingInfoOptional) throws InvalidJobException
     }
 
     public boolean requireInheritInstanceCheck() {
-        return this.schedulingInfo != null && this.deploymentStrategy != null && this.getDeploymentStrategy().requireInheritInstanceCheck();
+        return this.schedulingInfo != null &&
```
sundargates, line 136: I just want to confirm that my understanding is correct. Are we considering the absence of a deploymentStrategy as indicating that we need to have the same worker count as the previous job?

sundargates:
> I wonder if there could be breaking cases on multi-instance jobCluster e.g. job cluster running multiple jobs with auto-scaler and new job instance used a huge auto-scaled job instance count instead of the schedulingInfo spec count?

+1 on this. Why can we not add a deploymentStrategy for these jobs that need this?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobDefinition.java
hmitnflx: yes

PR #506, opened 2023-08-01
fix agent overrides

Andyz26:
### Context

* Fix agent property overrides: move env var based config value to part of worker config directly. Remove agent entry overrides.
* Re-enable test container test.

### Checklist

### DIFF
--- a/mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfiguration.java
+++ b/mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfiguration.java
@@ -38,7 +38,7 @@ public interface WorkerConfiguration extends CoreConfiguration {
     // ------------------------------------------------------------------------
     //  Task Executor machine related configurations
     // ------------------------------------------------------------------------
-    @Config("mantis.taskexecutor.id")
+    @Config({"mantis.taskexecutor.id", "MANTIS_AGENT_ID"})
     @DefaultNull
     String getTaskExecutorId();
 
@@ -46,7 +46,7 @@ public interface WorkerConfiguration extends CoreConfiguration {
         return getExternalAddress();
     }
 
-    @Config("mantis.taskexecutor.cluster-id")
+    @Config({"mantis.taskexecutor.cluster-id", "RESOURCE_CLUSTER_ID"})
     @Default("DEFAULT_CLUSTER")
     String getClusterId();
 
@@ -96,7 +96,7 @@ public interface WorkerConfiguration extends CoreConfiguration {
     // ------------------------------------------------------------------------
     //  RPC related configurations
     // ------------------------------------------------------------------------
-    @Config("mantis.taskexecutor.rpc.external-address")
+    @Config({"mantis.taskexecutor.rpc.external-address", "HOSTNAME"})
     @Default("localhost")
     String getExternalAddress();
 
--- a/mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfiguration.java
+++ b/mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfiguration.java
@@ -46,7 +46,7 @@ public interface WorkerConfiguration extends CoreConfiguration {
         return getExternalAddress();
     }
 
-    @Config({"mantis.taskexecutor.cluster-id", "RESOURCE_CLUSTER_ID"})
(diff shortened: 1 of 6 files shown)


mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfiguration.java
```diff
@@ -96,7 +96,7 @@ default Time getHeartbeatInterval() {
     // ------------------------------------------------------------------------
     //  RPC related configurations
     // ------------------------------------------------------------------------
-    @Config("mantis.taskexecutor.rpc.external-address")
+    @Config({"mantis.taskexecutor.rpc.external-address", "HOSTNAME"})
```
calvin681, line 99: maybe use `MANTIS_TASKEXECUTOR_RPC_EXTERNAL_ADDRESS` to prevent unintentionally using a value. Same for the other configs. Using the same config name should also reduce confusion.

Andyz26: this one is a bit tricky as the HOSTNAME is dynamic but the agent needs it at startup (so in test I cannot get this value till it's started but the agent needs it to start entry point). 
Let me add "MANTIS_TASKEXECUTOR_RPC_EXTERNAL_ADDRESS" in front of "HOSTNAME" and keep it as a last fallback option.

Andyz26: updated both

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfiguration.java
```diff
@@ -96,7 +96,7 @@ default Time getHeartbeatInterval() {
     // ------------------------------------------------------------------------
     //  RPC related configurations
     // ------------------------------------------------------------------------
-    @Config("mantis.taskexecutor.rpc.external-address")
+    @Config({"mantis.taskexecutor.rpc.external-address", "MANTIS_TASKEXECUTOR_RPC_EXTERNAL_ADDRESS", "HOSTNAME"})
```
calvin681, line 99: remove `HOSTNAME`

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfiguration.java
```diff
@@ -38,15 +38,15 @@ public interface WorkerConfiguration extends CoreConfiguration {
...
+    @Config({"mantis.taskexecutor.id", "MANTIS_AGENT_ID"})
     @DefaultNull
     String getTaskExecutorId();
 
     default String getTaskExecutorHostName() {
         return getExternalAddress();
     }
 
-    @Config("mantis.taskexecutor.cluster-id")
+    @Config({"mantis.taskexecutor.cluster-id", "MANTIS_RESOURCE_CLUSTER_ID"})
```
calvin681, line 49: i think keeping consistent and use `MANTIS_TASKEXECUTOR_CLUSTER_ID` would be better

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfiguration.java
```diff
@@ -38,15 +38,15 @@ public interface WorkerConfiguration extends CoreConfiguration {
     // ------------------------------------------------------------------------
     //  Task Executor machine related configurations
     // ------------------------------------------------------------------------
-    @Config("mantis.taskexecutor.id")
+    @Config({"mantis.taskexecutor.id", "MANTIS_AGENT_ID"})
```
calvin681, line 41: i think keeping consistent and use `MANTIS_TASKEXECUTOR_ID` would be better

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfiguration.java
Andyz26: I cannot remove HOSTNAME here as I need to translate the env to the address config and I cannot set it outside the container as env (I cannot get this value from outside before starting the agent container).

Andyz26: If this feels too risky I can revive the env override and fork dockerfile.

calvin681 (approved):

PR #510, opened 2023-08-03
Fix getTaskExecutorStatus failure on invalid TE ID

Andyz26:
### Context
Calling GetTaskExecutorStatus gets 500 internal error + actor crash error.
Fix:
* Separate the calls and return the proper type for the actor ask.
* Add UTs.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
@@ -228,14 +228,11 @@ public interface ResourceCluster extends ResourceClusterGateway {
         @Nullable
         WorkerId workerId;
         long lastHeartbeatInMs;
+    }
 
-        public static TaskExecutorStatus NotFound = new TaskExecutorStatus(
-            null,
-            false,
-            false,
-            false,
-            false,
-            null,
-            -1);
+    class NotFoundException extends Exception {
+        public NotFoundException(String message) {
+            super(message);
+        }
     }
 }
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
@@ -230,8 +230,8 @@ public interface ResourceCluster extends ResourceClusterGateway {
         long lastHeartbeatInMs;
     }
 
-    class NotFoundException extends Exception {
-        public NotFoundException(String message) {
+    class ResourceNotFoundException extends Exception {
+        public ResourceNotFoundException(String message) {
             super(message);
         }
     }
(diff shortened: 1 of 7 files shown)


fdc-ntflx (approved):
thanks for another important fix Andy.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
```diff
@@ -228,5 +228,14 @@ class TaskExecutorStatus {
         @Nullable
         WorkerId workerId;
         long lastHeartbeatInMs;
+
+        public static TaskExecutorStatus NotFound = new TaskExecutorStatus(
```
sundargates, line 232: Why not throw a known Exception instead?

sundargates (changes requested):
I prefer that we throw exceptions instead since it's more consistent with the rest of the ResourceCluster interface.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
sundargates: I believe that utilizing the same DS for exceptional cases can be problematic. It requires every caller to perform additional checks such as “if (case1) do this else if (case2) do that” and so on. However, if the issue is exceptional in nature and it is automatically surfaced/manifested as such, it becomes more intuitive from the client’s point of view.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -636,11 +638,13 @@ private ResourceOverview getResourceOverview() {
         return this.executorStateManager.getResourceOverview();
     }
 
-    private TaskExecutorStatus getTaskExecutorStatus(TaskExecutorID taskExecutorID) {
+    @Nonnull
+    private Object getTaskExecutorStatus(TaskExecutorID taskExecutorID) throws ResourceNotFoundException {
```
sundargates, line 642: nit: It's not throwing the exception.

sundargates: After your change, the only caller for this method is within the Akka context. Thus, you might as well change the signature to be reflective of that. 
```
private void onGetTaskExecutorStatus() {
  // send the message directly here. 
  getSender().tell();
}
```

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v1/BaseRoute.java
```diff
@@ -322,7 +323,12 @@ Boolean parseBoolean(String val) {
     protected  <T> Route withFuture(CompletableFuture<T> tFuture) {
         return onComplete(tFuture,
             t -> t.fold(
-                throwable -> complete(StatusCodes.INTERNAL_SERVER_ERROR, throwable, Jackson.marshaller()),
+                throwable -> {
+                    if (throwable instanceof ResourceNotFoundException) {
```
sundargates, line 327: I would move ResourceNotFoundException to some proto package (that's user-facing) to imply that this is part of the API and then have TaskExecutorNotFoundException extend it.

sundargates (approved):

PR #513, opened 2023-08-04
Avoid throwing npe when either TE gateway can't be reached or if caching job artifacts fails 

fdc-ntflx:
### Context

Upon master restart, the resource cluster actor is repeatedly crashing due to NPE thrown during caching job artifact routine. Although it's currently unknown why that's the case (we should spend a bit of time in the future to figure that out), for the time being it's important we never crash the actor in this non-critical function. 

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
@@ -724,9 +724,22 @@ class ResourceClusterActor extends AbstractActorWithTimers {
     private void onCacheJobArtifactsOnTaskExecutorRequest(CacheJobArtifactsOnTaskExecutorRequest request) {
         TaskExecutorState state = this.executorStateManager.get(request.getTaskExecutorID());
         if (state != null) {
-            TaskExecutorGateway gateway = state.getGateway().join();
-            List<URI> artifacts = jobArtifactsToCache.stream().map(artifactID -> URI.create(artifactID.getResourceID())).collect(Collectors.toList());
-            gateway.cacheJobArtifacts(new CacheJobArtifactsRequest(artifacts));
+            if (state.getGateway() != null) {
+                TaskExecutorGateway gateway = state.getGateway().join();
+                if (gateway != null) {
+                    // TODO(fdichiara): store URI directly to avoid remapping for each TE
+                    List<URI> artifacts = jobArtifactsToCache.stream().map(artifactID -> URI.create(artifactID.getResourceID())).collect(Collectors.toList());
+                    try {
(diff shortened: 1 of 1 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -723,10 +723,14 @@ private void updateHeartbeatTimeout(TaskExecutorID taskExecutorID) {
 
     private void onCacheJobArtifactsOnTaskExecutorRequest(CacheJobArtifactsOnTaskExecutorRequest request) {
         TaskExecutorState state = this.executorStateManager.get(request.getTaskExecutorID());
-        if (state != null) {
+        if (state != null && state.getGateway() != null) {
```
Andyz26, line 726: nit: warn log on gateway being null.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -723,10 +723,14 @@ private void updateHeartbeatTimeout(TaskExecutorID taskExecutorID) {
...
             TaskExecutorGateway gateway = state.getGateway().join();
-            List<URI> artifacts = jobArtifactsToCache.stream().map(artifactID -> URI.create(artifactID.getResourceID())).collect(Collectors.toList());
-            gateway.cacheJobArtifacts(new CacheJobArtifactsRequest(artifacts));
+            if (gateway != null) {
+                List<URI> artifacts = jobArtifactsToCache.stream().map(artifactID -> URI.create(artifactID.getResourceID())).collect(Collectors.toList());
```
Andyz26, line 729: nit: shall we save the URI in cache map directly to avoid the remapping for each TE?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -723,10 +723,14 @@ private void updateHeartbeatTimeout(TaskExecutorID taskExecutorID) {
...
+                gateway.cacheJobArtifacts(new CacheJobArtifactsRequest(artifacts));
```
Andyz26, line 730: shall we try/catch this call? If failed should the RC actor crash and restart?

Andyz26:
FYI please update subject + desc.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
fdc-ntflx: let me add a todo.

PR #514, opened 2023-08-05
Improving Job Artifact Caching: Immediate Downloads, Configurable Clusters, and Efficient Cleanup Mechanism

fdc-ntflx:
### Context

This PR is intended to improve the job artifact caching in these areas:

- Immediate downloads of newly registered job artifacts are ensured by _available_ Task Executors, eliminating the need for the first Task Executor restart. This approach is likely to speed up job initializations following the registration of a new job artifact.
- By attaching their clusters to a fast property (for now - we can have an API in the future), users have the option to adapt their job clusters for artifact caching. Master will ensure that the most recent artifact is always cached for the clusters configured in this way. This process occurs during each job launch. If any job's artifact has not been previously encountered, it will be included in the caching process after being added to the artifacts-to-cache list.
- To prevent the Task Executor from issues such as disk overuse, slowed startup times, and others that might arise from overloading with too many artifacts, a limit has been placed on the number of simultaneous artifact downloads. Hence, Task Executors won't download more than X artifacts concurrently.

#### Cleanup unused job artifacts : 

- A purging procedure must be established to continually maintain a suitably compact list of job artifacts used for caching. This implies that unused artifacts should not be cached. I haven't decided yet on how to implement this. An external cleaning mechanism, which aligns required cache artifacts based on active jobs and configured job clusters, is likely to be a quicker and less fault-prone solution than probably implementing it within master. Since we can already delete artifacts via api this cleanup mechanism is not super urgent. 

### Tests

Tested successfully in staging.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
@@ -54,6 +54,7 @@ import io.mantisrx.server.worker.TaskExecutorGateway;
 import io.mantisrx.server.worker.TaskExecutorGateway.TaskNotFoundException;
 import io.mantisrx.shaded.com.google.common.base.Preconditions;
 import io.mantisrx.shaded.com.google.common.collect.Comparators;
+import io.mantisrx.shaded.com.google.common.collect.ImmutableList;
 import io.mantisrx.shaded.com.google.common.collect.ImmutableMap;
 import io.vavr.Tuple;
 import java.io.IOException;
@@ -62,6 +63,8 @@ import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -110,8 +113,11 @@ class ResourceClusterActor extends AbstractActorWithTimers {
 
     private final HashSet<ArtifactID> jobArtifactsToCache = new HashSet<>();
 
-    static Props props(final ClusterID clusterID, final Duration heartbeatTimeout, Duration assignmentTimeout, Duration disabledTaskExecutorsCheckInterval, Clock clock, RpcService rpcService, MantisJobStore mantisJobStore, JobMessageRouter jobMessageRouter) {
-        return Props.create(ResourceClusterActor.class, clusterID, heartbeatTimeout, assignmentTimeout, disabledTaskExecutorsCheckInterval, clock, rpcService, mantisJobStore, jobMessageRouter);
+    private final int maxJobArtifactsToCache;
+    private final String jobClustersWithArtifactCachingEnabled;
+
+    static Props props(final ClusterID clusterID, final Duration heartbeatTimeout, Duration assignmentTimeout, Duration disabledTaskExecutorsCheckInterval, Clock clock, RpcService rpcService, MantisJobStore mantisJobStore, JobMessageRouter jobMessageRouter, int maxJobArtifactsToCache, String jobClustersWithArtifactCachingEnabled) {
+        return Props.create(ResourceClusterActor.class, clusterID, heartbeatTimeout, assignmentTimeout, disabledTaskExecutorsCheckInterval, clock, rpcService, mantisJobStore, jobMessageRouter, maxJobArtifactsToCache, jobClustersWithArtifactCachingEnabled);
     }
 
     ResourceClusterActor(
@@ -122,7 +128,9 @@ class ResourceClusterActor extends AbstractActorWithTimers {
         Clock clock,
         RpcService rpcService,
         MantisJobStore mantisJobStore,
-        JobMessageRouter jobMessageRouter) {
+        JobMessageRouter jobMessageRouter,
+        int maxJobArtifactsToCache,
+        String jobClustersWithArtifactCachingEnabled) {
         this.clusterID = clusterID;
         this.heartbeatTimeout = heartbeatTimeout;
         this.assignmentTimeout = assignmentTimeout;
@@ -134,6 +142,8 @@ class ResourceClusterActor extends AbstractActorWithTimers {
         this.mantisJobStore = mantisJobStore;
         this.activeDisableTaskExecutorsByAttributesRequests = new HashSet<>();
         this.disabledTaskExecutors = new HashSet<>();
+        this.maxJobArtifactsToCache = maxJobArtifactsToCache;
+        this.jobClustersWithArtifactCachingEnabled = jobClustersWithArtifactCachingEnabled;
 
         this.executorStateManager = new ExecutorStateManagerImpl();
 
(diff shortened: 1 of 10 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -723,13 +738,14 @@ private void updateHeartbeatTimeout(TaskExecutorID taskExecutorID) {
 
     private void onCacheJobArtifactsOnTaskExecutorRequest(CacheJobArtifactsOnTaskExecutorRequest request) {
         TaskExecutorState state = this.executorStateManager.get(request.getTaskExecutorID());
-        if (state != null) {
+        if (state != null && state.isAvailable()) {
```
sundargates, line 741: Shouldn't this be `isRegistered`?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -209,14 +209,29 @@ public Receive createReceive() {
...
+                refreshTaskExecutorJobArtifactCache();
+            }
             sender().tell(Ack.getInstance(), self());
         } catch (IOException e) {
             log.warn("Cannot add new job artifacts {} to cache in cluster: {}", req.getArtifacts(), req.getClusterID(), e);
         }
     }
 
+    private void refreshTaskExecutorJobArtifactCache() {
+        getTaskExecutors(ExecutorStateManager.isAvailable).getTaskExecutors().forEach(taskExecutorID ->
```
sundargates, line 231: Should this be the set of registered TEs instead?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java
```diff
@@ -132,6 +145,21 @@ private void onScheduleRequestEvent(ScheduleRequestEvent event) {
         pipe(assignedFuture, getContext().getDispatcher()).to(self());
     }
 
+    private void addNewJobArtifactsToCache(URL jobJarUrl) {
```
Andyz26, line 148: is it possible to add some UTs around these logic?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -209,14 +209,29 @@ public Receive createReceive() {
...
-            jobArtifactsToCache.addAll(req.artifacts);
+            List<ArtifactID> newArtifacts = req.artifacts
+                .stream()
+                .filter(artifactID -> !jobArtifactsToCache.contains(artifactID))
+                .collect(Collectors.toList());
+
+            if (!newArtifacts.isEmpty()) {
+                log.info("Storing and caching new artifacts: {}", newArtifacts);
+
+                mantisJobStore.addNewJobArtifactsToCache(req.getClusterID(), newArtifacts);
```
sundargates, line 220: I would probably change the semantics of this as follows:
```
s <- set of artifacts to be cached 

addNewArtifacts(delta):
  s <- s + delta
  store s
  refresh()
```

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -209,14 +209,29 @@ public Receive createReceive() {
...
+                jobArtifactsToCache.addAll(newArtifacts);
```
Andyz26, line 221: shall we track/alert the size of cached artifacts in e.g. metrics?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -209,14 +209,29 @@ public Receive createReceive() {
...
+                refreshTaskExecutorJobArtifactCache();
```
Andyz26, line 222: Any potential impact on TE startup if the cache list is long or contains a few large artifacts?

Andyz26 (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
fdc-ntflx: why? let's say the actor crashes before storing the updated `s`, the delta will get lost. Is that what we want?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java
```diff
@@ -382,4 +410,25 @@ private static class Noop {
...
         return "Retry-Schedule-Request-For" + workerId.toString();
     }
+
+    private Set<String> getJobClustersWithArtifactCacheEnabled() {
+        String jobClusters =
+            ServiceRegistry.INSTANCE.getPropertiesService()
+                .getStringValue("mantis.artifactCaching.jobClusters", "");
+        return new HashSet<>(Arrays.asList(jobClusters.split(",")));
+    }
+
```
sundargates, line 420: The semantics of this feels very specific to this implementation. I wonder if there's a way for us to make this more extendable.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
fdc-ntflx: Good question. From the TE point of view we want  to cache job artifacts only if the TE can still receive a job. If that's not the case we shouldn't since the cache is ephemeral in nature and will get purged at every restart. With that we also make caching very efficient when we add new job artifacts since we only ask available TEs to download at the same time, avoiding a burst of unnecessary calls to s3.  wdyt?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
fdc-ntflx: see answer below

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
sundargates: sure, you can change the ordering as you deem fit. But, the main point was about setting the goal state of the cache.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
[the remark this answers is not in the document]
fdc-ntflx: not sure I follow.

fdc-ntflx: done.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
fdc-ntflx: Today TEs can receive job requests while downloading job artifacts so the answer to your question is no. However this can be a problem in the future if we want to only accept requests _after_ all job artifacts have been downloaded.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java
```diff
@@ -132,6 +145,21 @@ private void onScheduleRequestEvent(ScheduleRequestEvent event) {
...
+            final ArtifactID artifact =  ArtifactID.of(urlString.substring(urlString.lastIndexOf('/') + 1));
+            resourceCluster.addNewJobArtifactsToCache(
+                    ClusterID.of(resourceCluster.getName()), Collections.singletonList(artifact))
+                .exceptionally(e -> {
+                    log.warn("Error while adding new job artifacts to cache", e);
+                    return Ack.getInstance();
+                });
+        } catch (Exception e) {
+            log.warn("Error while adding new job artifacts to cache", e);
+        }
```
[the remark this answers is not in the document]
fdc-ntflx: I can't think of anything. let me remove this.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -209,14 +209,29 @@ public Receive createReceive() {
...
+            }
             sender().tell(Ack.getInstance(), self());
         } catch (IOException e) {
             log.warn("Cannot add new job artifacts {} to cache in cluster: {}", req.getArtifacts(), req.getClusterID(), e);
         }
     }
 
+    private void refreshTaskExecutorJobArtifactCache() {
```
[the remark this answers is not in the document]
sundargates: yes, but there's no protection against it.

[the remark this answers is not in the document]
fdc-ntflx: that's fair.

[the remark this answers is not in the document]
fdc-ntflx: will add a todo.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
sundargates: We should decouple the TEs implementation details from the control plane's semantics. The current semantics is that we want to ensure all the TEs cache a set of common job artifacts. I think we should make these calls to the TEs based on that. The implementation details of how the TE manages the cache are independent, IMO.

PR #521, opened 2023-08-11
Handle TE gateway timeout

Andyz26:
### Context

Gateway registration runs as a detached future and the failure there is caused exception leak to the common thread pool and is not connected back to the state.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
@@ -345,7 +345,7 @@ class ResourceClusterActor extends AbstractActorWithTimers {
         } else {
             try {
                 if (state.isRegistered()) {
-                    sender().tell(state.getGateway().join(), self());
+                    sender().tell(state.getGateway(), self());
                 } else {
                     sender().tell(new Status.Failure(new Exception("")), self());
                 }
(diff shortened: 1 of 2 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorState.java
```diff
@@ -100,10 +100,10 @@ boolean onRegistration(TaskExecutorRegistration registration) {
             this.registration = registration;
             this.gateway =
                 rpcService.connect(registration.getTaskExecutorAddress(), TaskExecutorGateway.class)
-                    .whenComplete((gateway, throwable) -> {
-                        if (throwable != null) {
-                            log.error("Failed to connect to the gateway", throwable);
-                        }
+                    .exceptionally((throwable) -> {
```
fdc-ntflx, line 103: curious, why not keeping `whenComplete`?

Andyz26: whenComplete throws the exception to the executing thread.

Andyz26: and here we don't need that un-controlled behavior.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorState.java
```diff
@@ -100,10 +100,10 @@ boolean onRegistration(TaskExecutorRegistration registration) {
...
+                        log.error("Failed to connect to the gateway onRegistration, reset to unregistered.", throwable);
+                        onDisconnection();
```
fdc-ntflx, line 105: why disconnect vs reconnect?

Andyz26: this is logic inside taskExecutorState and chain the detached future error to another detached future error is doable but I prefer to have the upper actor to make these decisions based on registration/heartbeat events.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorState.java
```diff
@@ -100,10 +100,9 @@ boolean onRegistration(TaskExecutorRegistration registration) {
             this.registration = registration;
             this.gateway =
                 rpcService.connect(registration.getTaskExecutorAddress(), TaskExecutorGateway.class)
-                    .whenComplete((gateway, throwable) -> {
-                        if (throwable != null) {
-                            log.error("Failed to connect to the gateway", throwable);
-                        }
+                    .exceptionally((throwable) -> {
+                        log.error("Failed to connect to the gateway onRegistration, reset to unregistered.", throwable);
```
fdc-ntflx, line 104: I am under the impression that an exception is supposed to be thrown when onRegistration encounters a failure. In fact, onTaskExecutorRegistration incorporates a try/catch block around it. If we decide not to throw any exceptions now and state.onRegistration(registration); gives a true value, could this result in a semantic shift? Are all possible scenarios addressed correctly in this situation?

Andyz26: the try-catch on RC actor won't be available to catch this exception at all as it wasn't part of the returned value.

Andyz26: the exception will happen to another thread later time and RC actor won't see it either.

fdc-ntflx (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorState.java
```diff
@@ -251,8 +252,11 @@ TaskExecutorRegistration getRegistration() {
         return this.registration;
     }
 
-    protected CompletableFuture<TaskExecutorGateway> getGateway() {
-        return this.gateway;
+    protected TaskExecutorGateway getGateway() throws ExecutionException, InterruptedException {
+        if (this.gateway == null) {
+            throw new IllegalStateException("gateway is null");
```
sundargates, line 257: isn't NullPointerException semantically more correct?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -495,6 +495,7 @@ private void onTaskExecutorRegistration(TaskExecutorRegistration registration) {
             }
             sender().tell(Ack.getInstance(), self());
         } catch (Exception e) {
+            log.error("onTaskExecutorRegistration exception caught", e);
```
sundargates, line 498: we don't need this log.error since this is already being emitted when connection fails.

sundargates (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorState.java
Andyz26: NPE tends to be harder to read in logs as it doesn't point to the exact null object. Semantically this is illegal state too if a "client" try to get the gateway while it's has failed/disconnected?

PR #528, opened 2023-08-16
Put submitted vs launched behind a config/flag

hmitnflx:
### Context

This might be a more serious change and it'd be better to roll this out gradually separate from resource cluster upgrade. Putting the change behind a config (with default `false`).

### Checklist

### DIFF
--- a/mantis-control-plane/Dockerfile
+++ b/mantis-control-plane/Dockerfile
@@ -10,6 +10,9 @@ MAINTAINER Mantis Developers <mantis-oss-dev@netflix.com>
 #    apt-get install -y mesos=1.3.2-2.0.1 && \
 #    apt-get clean
 
+RUN mkdir -p /apps/mantis/mantis-control-plane-server/bin
+RUN mkdir -p /apps/mantis/mantis-control-plane-server/lib
+
 COPY ./mantis-control-plane-server/build/install/mantis-control-plane-server/bin/* /apps/mantis/mantis-control-plane-server/bin/
 COPY ./mantis-control-plane-server/build/install/mantis-control-plane-server/lib/* /apps/mantis/mantis-control-plane-server/lib/
 


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/MasterApiAkkaService.java
```diff
@@ -93,6 +93,7 @@ public class MasterApiAkkaService extends BaseService {
     private final Materializer materializer;
     private final ExecutorService executorService;
     private final CountDownLatch serviceLatch = new CountDownLatch(1);
+    private final boolean namedJobsReferToLaunched;
```
sundargates, line 96: nit: Can we wrap this in a config/settings class?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v0/JobDiscoveryRoute.java
```diff
@@ -134,14 +139,22 @@ private Route getJobDiscoveryRoutes() {
                                                     "/namedjobs/{} called",
                                                     jobCluster);
                                             jobClusterInfoStreamGET.increment();
-                                            JobClusterManagerProto.GetLastLaunchedJobIdStreamRequest req =
+                                            CompletionStage<JobDiscoveryRouteProto.JobClusterInfoResponse> jobClusterInfoRespCS;
```
sundargates, line 142: nit: final

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/config/MasterConfiguration.java
```diff
@@ -386,4 +386,8 @@ default Duration getHeartbeatInterval() {
     default Duration getMaxAssignmentThreshold() {
         return Duration.ofMillis(getAssignmentIntervalInMs());
     }
+
+    @Config("mantis.route.namedjobs.useLaunched")
+    @Default("false")
+    boolean getNamedJobsReferToLaunched();
```
sundargates, line 392: Do we not need this to be a fast property?

sundargates (approved):
LGTM!

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/config/MasterConfiguration.java
hmitnflx: fast property would be nice but I think having it as a config is also okay.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/MasterApiAkkaService.java
hmitnflx: umm, you okay if I pass the MasterConfiguration as-is instead of just this property?

PR #538, opened 2023-08-24
Implement a jittered exponential backoff retry policy for registration/heartbeat failures on the task executor

fdc-ntflx:
### Context

This PR adds the following things:

* Retry with random jittered exponential backoff - to prevent thundering herd of requests to master - to TE heartbeat. Retry configs can be changed
* Retry with random jittered exponential backoff - to prevent thundering herd of requests to master - to TE registration. Retry configs can be changed
* Counters: heartbeat timeouts, heartbeat failures, TE registration failures, TE disconnection failures 

Also ResourceManagerGatewayCxn has been moved to its own class since it's growing large.

Tests have been updated.

### Checklist

### DIFF
--- a/mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfigurationWritable.java
+++ b/mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfigurationWritable.java
@@ -52,6 +52,12 @@ public class WorkerConfigurationWritable implements WorkerConfiguration {
     int heartbeatInternalInMs;
     int tolerableConsecutiveHeartbeatFailures;
     int heartbeatTimeoutMs;
+    long heartbeatRetryInitialDelayMs;
+    long heartbeatRetryMaxDelayMs;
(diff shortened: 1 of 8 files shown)


mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfigurationWritable.java
```diff
@@ -168,6 +174,36 @@ public int heartbeatTimeoutMs() {
         return this.heartbeatTimeoutMs;
     }
 
+    @Override
+    public long heartbeatRetryInitialDelayMs() {
```
Andyz26, line 178: Doesn't need to be for this PR: seems like we can consider separate worker/agent config vs worker runtime (for the actual job worker) config. These ones don't need to be passed into the job worker I guess.

Andyz26 (approved):
[idea]: you can mock/test e2e in testcontainers to simulate the master down scenario (probably a test not in the main build but only run manually).

PR #539, opened 2023-08-24
Improve actor and http api server metrics

Andyz26:
### Context

* New metrics in http server/routes.
* Attach metered mailbox to resource cluster actor.

### Checklist

### DIFF
--- a/mantis-common/src/main/java/io/mantisrx/common/metrics/LoggingMetricsPublisher.java
+++ b/mantis-common/src/main/java/io/mantisrx/common/metrics/LoggingMetricsPublisher.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2023 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.common.metrics;
+
+import io.mantisrx.shaded.com.google.common.base.Joiner;
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Properties;
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * This implementation will print the metric group in log and is only recommended using in Test/Dev only.
+ * To use this publisher override config "mantis.metricsPublisher.class" in the master-*.properties file.
+ */
+@Slf4j
+public class LoggingMetricsPublisher extends MetricsPublisher {
+    public static final String LOGGING_ENABLED_METRICS_GROUP_ID_LIST_KEY =
+            "MANTIS_LOGGING_ENABLED_METRICS_GROUP_ID_LIST";
+    private Set<String> loggingEnabledMetricsGroupId = new HashSet<>();
+
+    public LoggingMetricsPublisher(Properties properties) {
+        super(properties);
+
+        String key = properties.getProperty(LOGGING_ENABLED_METRICS_GROUP_ID_LIST_KEY,
+                System.getenv(LOGGING_ENABLED_METRICS_GROUP_ID_LIST_KEY));
+        log.info("LOGGING_ENABLED_METRICS_GROUP_ID_LIST_KEY: {}", key);
+
+        if (key != null) {
+            this.loggingEnabledMetricsGroupId =
+                    Arrays.stream(key
+                            .toLowerCase().split(";")).collect(Collectors.toSet());
+            log.info("[Metrics Publisher] enable logging for: {}",
+                    Joiner.on(',').join(this.loggingEnabledMetricsGroupId));
+        }
+    }
+
+    @Override
+    public void publishMetrics(long timestamp,
+            Collection<Metrics> currentMetricsRegistered) {
+        log.info("Printing metrics from: {}", Instant.ofEpochMilli(timestamp));
+        currentMetricsRegistered.stream()
+                .filter(ms -> this.loggingEnabledMetricsGroupId.contains(ms.getMetricGroupId().id().toLowerCase()))
+                .map(ms -> ms.counters().entrySet())
+                .flatMap(Collection::stream)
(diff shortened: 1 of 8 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/MasterMain.java
```diff
@@ -365,6 +367,7 @@ public static void main(String[] args) {
         }
 
         try {
+            SpectatorRegistryFactory.setRegistry(new DefaultRegistry());
```
sundargates, line 370: why is this needed? Even if it's required, wouldn't it make sense to do this in nfmantis similar to worker-titus.

mantis-common/src/main/java/io/mantisrx/common/metrics/LoggingMetricsPublisher.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * This implementation will print the selected metric groups in log.
+ * To use this publisher override config "mantis.metricsPublisher.class" in the master-*.properties file.
+ */
+@Slf4j
+public class LoggingMetricsPublisher extends MetricsPublisher {
```
sundargates, line 34: just curious - why is this required? we are anyways publishing these metrics today via spectator - what's the need to log it.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/MasterApiAkkaService.java
```diff
@@ -209,18 +208,22 @@ private void startAPIServer() {
...
+                }))
+                .run(materializer)
+                .exceptionally(failure -> {
+                    System.err.println("API service exited, committing suicide !" + failure.getMessage());
+                    logger.info("Master API service exited in error, committing suicide !");
+                    system.terminate();
+                    System.exit(2);
+                    return null;
+                });
         logger.info("Starting Mantis Master API on port {}", port);
```
sundargates, line 227: What is this for?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/MasterApiMetrics.java
```diff
@@ -24,13 +24,15 @@ public class MasterApiMetrics {
     private final Counter resp2xx;
     private final Counter resp4xx;
     private final Counter resp5xx;
+    private final Counter incomingRequestCount;
```
sundargates, line 27: It looks like this is across all requests. Is there a way to emit this with the route path as a tag?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -117,7 +117,8 @@ class ResourceClusterActor extends AbstractActorWithTimers {
     private final String jobClustersWithArtifactCachingEnabled;
 
     static Props props(final ClusterID clusterID, final Duration heartbeatTimeout, Duration assignmentTimeout, Duration disabledTaskExecutorsCheckInterval, Clock clock, RpcService rpcService, MantisJobStore mantisJobStore, JobMessageRouter jobMessageRouter, int maxJobArtifactsToCache, String jobClustersWithArtifactCachingEnabled) {
-        return Props.create(ResourceClusterActor.class, clusterID, heartbeatTimeout, assignmentTimeout, disabledTaskExecutorsCheckInterval, clock, rpcService, mantisJobStore, jobMessageRouter, maxJobArtifactsToCache, jobClustersWithArtifactCachingEnabled);
+        return Props.create(ResourceClusterActor.class, clusterID, heartbeatTimeout, assignmentTimeout, disabledTaskExecutorsCheckInterval, clock, rpcService, mantisJobStore, jobMessageRouter, maxJobArtifactsToCache, jobClustersWithArtifactCachingEnabled)
```
sundargates, line 120: This can be configured directly in `application.conf`.

mantis-common/src/main/java/io/mantisrx/common/metrics/LoggingMetricsPublisher.java
sundargates: If this is only expected to be used for testing purposes, I recommend that you move this class to the test package and make it package private.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/MasterMain.java
sundargates: If this is for test containers, I would recommend moving this there since this module is not expected to be used as the leaf node.

Andyz26: in default masterMain-> main the spectator's default composite registry is empty thus no metrics is actually logged. This is needed for any app using this entry to be able to actually log metrics there.

mantis-common/src/main/java/io/mantisrx/common/metrics/LoggingMetricsPublisher.java
Andyz26: It is a debug mode but to use this it needs to be configured via the core config (this interface is one of these created in config instead of injection).
Also in OSS domain, I think it's okay to have a logging-based mode option.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/MasterApiAkkaService.java
Andyz26: Remove deprecated akka api + log incrementIncomingRequestCount.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/MasterApiMetrics.java
Andyz26: This is used at the initial connection point so the path could be anything (cardinality not controlled by us). we have some route-level metrics already at each route level but it's up to the route to decide whether to log so I want something in general for all connections (plus track if something keeps requesting for invalid paths).

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
Andyz26: let's keep it consistent with other usages e.g. jobManagerClusterActor?

PR #536, opened 2023-08-22
[1] Connection Monitor to close Dangling Connections

kriti-sc:
PR [1] Connection Monitor: https://github.com/Netflix/mantis/pull/536
PR [2] Test: https://github.com/Netflix/mantis/pull/550
PR [3] Metrics: https://github.com/Netflix/mantis/pull/551

This PR is a part of GSoC 2023, fixing the issue detailed [here](https://netflix.github.io/mantis/updates/gsoc2023-ideas/#mantis-connection-monitor). 

The SSE connection b/w mantis-api and worker, for instances where the worker does not produce any data, does not close. 
For instances where the worker does not produce data, [this](https://github.com/Netflix/mantis/blob/d0ead57078c60031d34794ba199b9b17392f428f/mantis-network/src/main/java/io/reactivex/mantis/network/push/PushServer.java#L316) callback is not called. 
```
channel.closeFuture().addListener(new GenericFutureListener<io.netty.util.concurrent.Future<Void>>() {
      @Override
      public void operationComplete(io.netty.util.concurrent.Future<Void> future) throws Exception {
          connectionManager.remove(connection);
          connectionCleanup(heartbeatSubscription, connectionClosedCallback, metaMsgSubscription);
          // Callback from the channel is closed, we don't need to check channel status anymore.
          if (writableCheck != null) {
              writableCheck.cancel(false);
          }
      }
  });
```

We expose the underlying connection object created [in this client.submit()](https://github.com/Netflix/mantis/blob/d0ead57078c60031d34794ba199b9b17392f428f/mantis-server/mantis-server-worker-client/src/main/java/io/mantisrx/server/worker/client/SseWorkerConnection.java#L216) call with the aim of explicitly closing it.

We create a new client implementation class that `MantisHttpClientImpl` [here](https://github.com/kriti-sc/mantis/blob/608a9733d996aabbd2d4207d768919bb997bc821/mantis-server/mantis-server-worker-client/src/main/java/io/mantisrx/server/worker/client/MantisHttpClientImpl.java#L39) inherits the originally used `HttpClientImpl`. The client mentioned above used this new implementation. 

[...]

PR #550, opened 2023-09-01
[2] Connection Monitor Tests

kriti-sc:
PR [1] Connection Monitor: https://github.com/Netflix/mantis/pull/536
PR [2] Test: https://github.com/Netflix/mantis/pull/550
PR [3] Metrics: https://github.com/Netflix/mantis/pull/551


mantis-server/mantis-server-worker-client/src/test/java/io/mantisrx/server/worker/client/SseWorkerConnectionTest.java
```diff
@@ -91,4 +105,274 @@ public void testStreamContentDrops() throws Exception {
...
+        client.trackConnection(dummyChannel);
+        logger.info("Connection tracker size: {}", client.connectionTrackerSize());
+        assertEquals(1, client.connectionTrackerSize());
+
+        client.closeConn();
+
+        logger.info("Connection tracker size: {}", client.connectionTrackerSize());
+        assertEquals(0, client.connectionTrackerSize());
+        logger.info("Channel active: {}, channel open: {}, channel writable: {}", dummyChannel.isActive(), dummyChannel.isOpen(), dummyChannel.isWritable());
+        assertFalse(dummyChannel.isActive());
```
calvin681, line 374: These assertions are pointless no? They are hardcoded to return false. We can just remove.

mantis-server/mantis-server-worker-client/src/test/java/io/mantisrx/server/worker/client/SseWorkerConnectionTest.java
```diff
@@ -91,4 +105,274 @@ public void testStreamContentDrops() throws Exception {
...
+        assertTrue(workerConnection.client instanceof MantisHttpClientImpl);
+
+
+        // check MantisHttpClientImpl.connect usage
+        logger.info("MantisHttpClientImpl.connect() called: {}", client.isObservableConectionSet());
+        assertTrue(client.isObservableConectionSet());
+
+
+        // check connection monitor functionality of MantisHttpClientImpl
+        Channel dummyChannel = new Channel() {
```
calvin681, line 148: It may be cleaner to just use a mock https://site.mockito.org/

mantis-server/mantis-server-worker-client/src/test/java/io/mantisrx/server/worker/client/SseWorkerConnectionTest.java
kriti-sc: I don't fully understand what you mean. 
I am closing the connection via the `closeConn` method and then testing if the original channel does get closed. If it does not, then the assertion will fail.

calvin681: Does dummyChannel.isActive() always return false? I don't see where it could be set to true.

kriti-sc: That is a great point. Removed those assertions.

calvin681 (approved):

PR #560, opened 2023-09-21
Make blob store thread safe

fdc-ntflx:
### Context

### Checklist

### DIFF
--- a/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/BlobStore.java
+++ b/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/BlobStore.java
@@ -24,6 +24,7 @@ import java.util.concurrent.locks.Lock;
 import lombok.AccessLevel;
 import lombok.RequiredArgsConstructor;
 import net.lingala.zip4j.ZipFile;
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
 
 /**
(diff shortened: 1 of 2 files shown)


mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/BlobStore.java
```diff
@@ -93,7 +99,7 @@ public File get(URI blobUrl) throws IOException {
                     String destDirStr = getUnzippedDestDir(z);
                     File destDir = new File(destDirStr);
                     if (destDir.exists()) {
-                        FileUtils.deleteDirectory(destDir);
+                        return destDir;
```
Andyz26, line 102: how do we ensure not using a corrupted dir from previous failed unzipping?

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/HadoopFileSystemBlobStore.java
```diff
@@ -41,22 +39,14 @@ public class HadoopFileSystemBlobStore implements BlobStore {
 
     private final File localStoreDir;
 
-    final Striped<Lock> locks = Striped.lock(1024);
```
Andyz26, line 44: is it possible now without lock two threads can try downloading the target zip from s3 same time?

Andyz26 (approved):

PR #562, opened 2023-10-06
Fix unexpected sink side unsubscribe behavior 

Andyz26:
### Context

There is a race condition on the subscription handler where the current state can be empty if the subscription is established before the handler service is started, which causes non-perpetual jobs to silently fail at startup.

- Added handling + logs to this case.
- Added extra test in integration test.
- Removed sinkSubscriptionHandlerFactory on TaskExecutor level (not used anymore).

### Checklist

### DIFF
--- a/mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/SubscriptionStateHandlerImpl.java
+++ b/mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/SubscriptionStateHandlerImpl.java
@@ -53,9 +53,7 @@ class SubscriptionStateHandlerImpl extends AbstractScheduledService implements S
 
     @Override
     public void startUp() {
-        if (currentState.get() == null) {
-            currentState.compareAndSet(null, SubscriptionState.of(clock));
-        }
+        currentState.compareAndSet(null, SubscriptionState.of(clock));
         log.info("SubscriptionStateHandlerImpl service started.");
     }
 
--- a/mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/SubscriptionStateHandlerImpl.java
+++ b/mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/SubscriptionStateHandlerImpl.java
@@ -56,6 +56,7 @@ class SubscriptionStateHandlerImpl extends AbstractScheduledService implements S
         if (currentState.get() == null) {
             currentState.compareAndSet(null, SubscriptionState.of(clock));
         }
+        log.info("SubscriptionStateHandlerImpl service started.");
     }
 
     @Override
@@ -92,11 +93,7 @@ class SubscriptionStateHandlerImpl extends AbstractScheduledService implements S
 
     @Override
     public void onSinkSubscribed() {
-        // sink subscription can happen before the startup of this service.
-        if (currentState.get() == null) {
-            currentState.compareAndSet(null, SubscriptionState.of(clock));
-        }
-
+        Preconditions.checkNotNull(currentState.get(), "currentState is not intialized onSinkSubscribed.");
         currentState.updateAndGet(SubscriptionState::onSinkSubscribed);
     }
 
--- a/mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/SubscriptionStateHandlerImpl.java
+++ b/mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/SubscriptionStateHandlerImpl.java
@@ -53,7 +53,9 @@ class SubscriptionStateHandlerImpl extends AbstractScheduledService implements S
 
     @Override
     public void startUp() {
-        currentState.set(SubscriptionState.of(clock));
+        if (currentState.get() == null) {
+            currentState.compareAndSet(null, SubscriptionState.of(clock));
+        }
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/SubscriptionStateHandlerImpl.java
```diff
@@ -81,11 +83,20 @@ protected void runOneIteration() {
...
+            return;
+        }
         currentState.updateAndGet(SubscriptionState::onSinkUnsubscribed);
     }
 
     @Override
     public void onSinkSubscribed() {
+        // sink subscription can happen before the startup of this service.
+        if (currentState.get() == null) {
+            currentState.compareAndSet(null, SubscriptionState.of(clock));
```
liuml07, line 97: same question, I was thinking the if-check is not necessary. `currentState.compareAndSet(null, SubscriptionState.of(clock));` works just fine.

mantis-runtime/src/main/java/io/mantisrx/runtime/executor/SinkPublisher.java
```diff
@@ -88,31 +87,29 @@ public void start(StageConfig<?, T> stage,
...
                     }
                 })
-                .doOnUnsubscribe(new Action0() {
-                    @Override
-                    public void call() {
-                        logger.info("Sink subscriptions clean up, action=" + onUnsubscribeAction);
-                        if (onUnsubscribeAction != null)
-                            onUnsubscribeAction.call();
-                    }
+                .doOnCompleted(() -> logger.info("Sink observable subscription completed."))
```
liuml07, line 101: nit: is other information useful in the log like subscribe etc.

liuml07 (approved):
LGTM but a second review is appreciated. I have limited context.

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/SubscriptionStateHandlerImpl.java
```diff
@@ -53,7 +53,9 @@ class SubscriptionStateHandlerImpl extends AbstractScheduledService implements S
 
     @Override
     public void startUp() {
-        currentState.set(SubscriptionState.of(clock));
+        if (currentState.get() == null) {
```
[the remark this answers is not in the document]
calvin681: seems like you are expecting scenarios where startUp can be called multiple times?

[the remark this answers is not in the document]
Andyz26: The currentState can be set at startUp or the first update from subscribe event (currently there is no set at subscribe so if subscribe happens before startup it hits NPE).

mantis-runtime/src/main/java/io/mantisrx/runtime/executor/SinkPublisher.java
Andyz26: subscribe?

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/SubscriptionStateHandlerImpl.java
```diff
@@ -81,11 +83,20 @@ protected void runOneIteration() {
 
     @Override
     public void onSinkUnsubscribed() {
+        if (currentState.get() == null) {
```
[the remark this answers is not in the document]
Andyz26: currentState should never be unset back to null in this instance, thus no race on getting non-null then reading null.

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/SubscriptionStateHandlerImpl.java
Andyz26: yeah it's not for thread safety but just a minor save to avoid the extra SubscriptionState.of(clock) creation.

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/SubscriptionStateHandlerImpl.java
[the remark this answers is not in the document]
sundargates: We need to ensure that none of the members of the class are called before startUp, as this would violate the semantics of a Service class. To prevent this, I suggest setting up preconditions in the listener methods rather than working around the issue.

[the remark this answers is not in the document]
sundargates: I think it might be a better idea to start the subscription handler in
```
private void setupSubscriptionStateHandler(ExecuteStageRequest executeStageRequest) {
```

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/SubscriptionStateHandlerImpl.java
```diff
@@ -53,7 +53,10 @@ class SubscriptionStateHandlerImpl extends AbstractScheduledService implements S
 
     @Override
     public void startUp() {
-        currentState.set(SubscriptionState.of(clock));
+        if (currentState.get() == null) {
```
sundargates, line 56: I would convert this into a Precondition instead.

mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/WorkerExecutionOperationsNetworkStage.java
```diff
@@ -478,6 +478,12 @@ private void setupSubscriptionStateHandler(ExecuteStageRequest executeStageReque
         };
 
         this.subscriptionStateHandler = subscriptionStateHandler;
+        try {
+            this.subscriptionStateHandler.startAsync().awaitRunning(Duration.of(5, ChronoUnit.SECONDS));
```
calvin681, line 482: This should be TimeUnit no?

Andyz26: TimeUnit is in java.util.concurrent and this Duration.of requires the temporal unit interface which is from java.time.

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/SubscriptionStateHandlerImpl.java
Andyz26: I don't think it has to fail if the service got started again. Let me remove the check here directly (atomic set with null should be sufficient.)

PR #569, opened 2023-10-18
Fix unsubsribe on mantisMasterClient schedulingChanges

Andyz26:
### Context

Using behavior subject breaks the subscription propagation back to the connection (mantis API layer) and can cause a connection leak on the API layer.
Revert the behavior subject change and apply "replay" instead of "share" on the client side observable stream to ensure last valid data item is still available while having the source connection stream unsub properly.

### Checklist


mantis-server/mantis-server-worker/src/main/java/io/mantisrx/server/worker/WorkerExecutionOperationsNetworkStage.java
```diff
@@ -279,7 +279,15 @@ public void executeStage(final ExecutionDetails setup) throws IOException {
         //Observable<JobSchedulingInfo> selfSchedulingInfo = mantisMasterApi.schedulingChanges(executionRequest.getJobId()).switchMap((e) -> Observable.just(e).repeatWhen(x -> x.delay(5 , TimeUnit.SECONDS))).subscribeOn(Schedulers.io()).share();
 
         // JobSchedulingInfo has metadata around which stage runs on which set of workers
-        Observable<JobSchedulingInfo> selfSchedulingInfo = mantisMasterApi.schedulingChanges(executionRequest.getJobId()).subscribeOn(Schedulers.io()).share();
+        Observable<JobSchedulingInfo> selfSchedulingInfo =
+            mantisMasterApi.schedulingChanges(executionRequest.getJobId())
+                .subscribeOn(Schedulers.io())
+                .replay(1)
+                .refCount()
+                .doOnSubscribe(() -> logger.info("mantisApi schedulingChanges subscribe"))
```
calvin681, line 287: Nice!

calvin681 (approved):

PR #576, opened 2023-11-02
Metrics for each of the RC Actor messages

sundargates:
### Context

### Checklist


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
Andyz26, line 1:
```
.match(GetClusterUsageRequest.class,
    req -> sender().tell(this.executorStateManager.getClusterUsage(req), self()))
.match(GetClusterIdleInstancesRequest.class,
    req -> sender().tell(onGetClusterIdleInstancesRequest(req), self()))
```
These two could be costly and worth checking too.

Andyz26 (approved):
Can you check/fix the test failure as well?

PR #579, opened 2023-11-06
Send registrations once per TE

sundargates:
### Context

### Checklist


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -550,16 +550,25 @@ private void onHeartbeat(TaskExecutorHeartbeat heartbeat) {
         try {
             final TaskExecutorID taskExecutorID = heartbeat.getTaskExecutorID();
             final TaskExecutorState state = this.executorStateManager.get(taskExecutorID);
-            boolean stateChange = state.onHeartbeat(heartbeat);
-            if (stateChange) {
-                if (state.isAvailable()) {
-                    this.executorStateManager.tryMarkAvailable(taskExecutorID);
+            if (state.getRegistration() == null) {
+                TaskExecutorRegistration registration = this.mantisJobStore.getTaskExecutor(heartbeat.getTaskExecutorID());
```
Andyz26, line 554: I suppose we should avoid blocking calls at this thread? move to future instead?

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/utils/DurableBooleanState.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public void setState(boolean state) throws IOException {
+        try (FileOutputStream fos = new FileOutputStream(fileName)) {
+            fos.write(state ? 1 : 0);
+            fos.flush();
+            log.info("The state has been stored successfully!");
+        }
+    }
+
+    public boolean getState() throws IOException {
```
Andyz26, line 40: To be safe from file corruption maybe catch the IO exception here and reset to default instead of leaving this error to the cxn to handle?

sundargates: Since this is a low-level utility, it is appropriate to propagate this error for semantic reasons.

fdc-ntflx (approved):

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/ResourceManagerGatewayCxn.java
```diff
@@ -114,10 +133,16 @@ protected Scheduler scheduler() {
...
+            if (!alreadyRegistered.getState()) {
+                log.info("Trying to register with resource manager {}", gateway);
+                registerTaskExecutorWithRetry();
+            } else {
+                log.info("Registered with resource manager {} already", gateway);
+            }
+
             registered = true;
+            alreadyRegistered.setState(true);
         } catch (Exception e) {
```
Andyz26, line 146: handle the IO error here? I guess we don't need to disconnect etc. if the alreadyRegistered file got corrupted.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
Andyz26, line 554: also what's the timeout here (max wait duration per req)?

sundargates: I don't think that's set here in the repo. In nfmantis, this is set to 10 seconds I think.

PR #580, opened 2023-11-07
Implement batch schedule request

fdc-ntflx:
### Context

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import javax.annotation.Nullable;
 import lombok.Value;
@@ -100,14 +101,13 @@ public interface ResourceCluster extends ResourceClusterGateway {
 
     /**
      * Can throw {@link NoResourceAvailableException} wrapped within the CompletableFuture in case there
-     * are no task executors.
+     * are no enough available task executors.
      *
-     * @param machineDefinition machine definition that's requested for the worker
-     * @param workerId          worker id of the task that's going to run on the node.
-     * @return task executor assigned for the particular task.
+     * @param allocationRequests set of machine definitions requested for 1 or more workers
+     * @return task executors assigned for the particular request
      */
-    CompletableFuture<TaskExecutorID> getTaskExecutorFor(
-        TaskExecutorAllocationRequest allocationRequest);
+    CompletableFuture<Map<TaskExecutorAllocationRequest, TaskExecutorID>> getTaskExecutorsFor(
+        Set<TaskExecutorAllocationRequest> allocationRequests);
 
     /**
      * Returns the Gateway instance to talk to the task executor. If unable to make connection with
@@ -187,6 +187,13 @@ public interface ResourceCluster extends ResourceClusterGateway {
      */
     CompletableFuture<Map<TaskExecutorID, WorkerId>> getTaskExecutorWorkerMapping(
         Map<String, String> attributes);
+
+    /**
+     * Signals that the given job id has been killed.
+     *
+     * @param jobId id of the job killed
+     */
+    CompletableFuture<Ack> unscheduleJob(String jobId);
 
     class NoResourceAvailableException extends Exception {
 
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
@@ -1317,6 +1317,7 @@ public class JobActor extends AbstractActorWithTimers implements IMantisJobManag
         private void initializeRunningWorkers() {
             // Scan for the list of all corrupted workers to be resubmitted.
             List<JobWorker> workersToResubmit = markCorruptedWorkers();
+            List<IMantisWorkerMetadata> workersToSubmit = new ArrayList<>();
 
             // publish a refresh before enqueuing tasks to the Scheduler, as there is a potential race between
             // WorkerRegistryV2 getting updated and isWorkerValid being called from SchedulingService loop
@@ -1357,7 +1358,14 @@ public class JobActor extends AbstractActorWithTimers implements IMantisJobManag
 
                         scheduler.initializeRunningWorker(scheduleRequest, wm.getSlave(), wm.getSlaveID());
                     } else if (wm.getState().equals(WorkerState.Accepted)) {
-                        queueTask(wm);
+
+                        // If the job is in accepted state, queue all its pending workers at once in a batch request.
+                        // This is important when before master failover there were pending batch requests
+                        if (JobState.isAcceptedState(mantisJobMetaData.getState())) {
+                            workersToSubmit.add(wm);
+                        } else {
+                            queueTask(wm);
+                        }
                     }
                 }
 
@@ -1365,6 +1373,10 @@ public class JobActor extends AbstractActorWithTimers implements IMantisJobManag
                     stageAssignments.put(stageMeta.getStageNum(), new WorkerAssignments(stageMeta.getStageNum(),
                             stageMeta.getNumWorkers(), workerHosts));
                 }
+            }
+
+            if (JobState.isAcceptedState(mantisJobMetaData.getState()) && !workersToSubmit.isEmpty()) {
+                queueTasks(workersToSubmit, empty());
             }
 
             // publish another update after queuing tasks to Fenzo (in case some workers were marked Started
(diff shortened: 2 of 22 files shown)


mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
```diff
@@ -188,6 +188,13 @@ CompletableFuture<PagedActiveJobOverview> getActiveJobOverview(
     CompletableFuture<Map<TaskExecutorID, WorkerId>> getTaskExecutorWorkerMapping(
         Map<String, String> attributes);
 
+    /**
+     * Signals that the given job id has been killed.
+     *
+     * @param jobId id of the job killed
+     */
+    CompletableFuture<Ack> unscheduleJob(String jobId);
```
sundargates, line 196: The interface seems to be leaking implementation details.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -227,10 +251,56 @@ public Optional<Entry<TaskExecutorID, TaskExecutorState>> findFirst(
...
+        for (Entry<MachineDefinition, List<TaskExecutorAllocationRequest>> entry : request.getGroupedByMachineDef().entrySet()) {
+            final MachineDefinition machineDefinition = entry.getKey();
+            final List<TaskExecutorAllocationRequest> allocationRequests = entry.getValue();
+
+            Optional<Map<TaskExecutorID, TaskExecutorState>> taskExecutors = findTaskExecutorsFor(request, machineDefinition, allocationRequests, isJobIdAlreadyPending);
+
+            // Mark noResourcesAvailable if we can't find enough TEs for a given machine def
+            if (!taskExecutors.isPresent() || noResourcesAvailable) {
+                noResourcesAvailable = true;
+                continue;
```
calvin681, line 279: nit: u can just break here

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
fdc-ntflx: As discussed in person, we can probably get rid of this and simply evict on cache TTL.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
fdc-ntflx: I don't think breaking is the right thing to do in here because we can potentially have more than 1 sku to add to the pending list so we want to loop through all the entries. Let me give you an example: job A has stage 1 with sku small and stage 2 with small medium. If both don't have enough resources available we want to insert both into the pending set which happens on `findTaskExecutorsFor`; however if we break after finding out that there are not enough resources available on sku small we would miss out on adding the medium sku into the pending set.

fdc-ntflx: you're right. this was there because of an older implementation of the cache.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java
```diff
@@ -120,6 +135,58 @@ public Receive createReceive() {
             .build();
     }
 
+    private void onBatchScheduleRequestEvent(BatchScheduleRequestEvent event) {
+        log.info("Received batch schedule request event: {}", event);
+
+        // If the size of the batch request is 1 we'll handle it as the "old" schedule request
+        if (event.getRequest().getScheduleRequests().size() == 1) {
```
Andyz26, line 142: any reason/upside to keeping the 1 node case special?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java
```diff
@@ -120,6 +135,58 @@ public Receive createReceive() {
...
+                    final AssignedScheduleRequestEvent assignedScheduleRequestEvent = new AssignedScheduleRequestEvent(scheduleRequestEvent, value);
+
+                    self().tell(assignedScheduleRequestEvent, self());
+                });
+    }
+
+    private void onFailedToBatchScheduleRequestEvent(FailedToBatchScheduleRequestEvent event) {
+        batchSchedulingFailures.increment();
+        if (event.getAttempt() >= this.maxScheduleRetries) {
+            log.error("Failed to submit the batch request {} because of ", event.getScheduleRequestEvent(), event.getThrowable());
```
Andyz26, line 177: in this case does it need to route some message/status back to the job actor?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareScheduler.java
```diff
@@ -37,8 +38,13 @@ public class ResourceClusterAwareScheduler implements MantisScheduler {
...
     @Override
-    public void scheduleWorker(ScheduleRequest scheduleRequest) {
-        schedulerActor.tell(ScheduleRequestEvent.of(scheduleRequest), null);
+    public void scheduleWorkers(BatchScheduleRequest scheduleRequest) {
+        schedulerActor.tell(BatchScheduleRequestEvent.of(scheduleRequest), null);
+    }
+
+    @Override
+    public void unscheduleJob(String jobId) {
+        schedulerActor.tell(CancelBatchRequestEvent.of(jobId),null);
```
Andyz26, line 47: CancelBatchRequestEvent seems to be only cancelling retrying on the scheduler side. I am wondering what happens if the rcActor is sending the assignment request while this is happening? Will the agent still get the cancel request?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
```diff
@@ -1357,7 +1358,14 @@ private void initializeRunningWorkers() {
 
                         scheduler.initializeRunningWorker(scheduleRequest, wm.getSlave(), wm.getSlaveID());
                     } else if (wm.getState().equals(WorkerState.Accepted)) {
```
Andyz26, line 1360: Is this part a potential state mismatch source that  a worker has been submitted to an agent but failed to update the worker info wm here during leader switch and it will get queued again (where the previous assigned agent is also running the same worker already)?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java
fdc-ntflx: Good question. If we're talking about actions like autoscaling or resubmitting a worker, scheduling requests will always going to involve one worker at the time. In these situations, we'd rather use the `ScheduleRequestEvent` lifecycle, not the `BatchScheduleRequestEvent`. The reason is that `ScheduleRequestEvent` lets us try again at the worker level. This means setting up timers for each worker (using their worker-id), instead of using timers for the entire job (with job-id) like we would in `BatchScheduleRequestEvent`, which will collide.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java
fdc-ntflx: I'm following the behavior we currently have during scheduling ops -> https://github.com/Netflix/mantis/blob/81bce548e92f5f4cc048f07a23acc76a26941717/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java#L206

fdc-ntflx: do you think we should change this behavior? If so, how?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareScheduler.java
fdc-ntflx: Not sure I understand the question. If you're asking what happens if the user decides to kill the job, there are 2 cases: 
1) job is in pending state because not enough resources are available. In this case we'll simply delete the timer so the scheduler won't retry. 
2) job is in accepted state and workers are already assigned. Those will still get cancelled as we're currently doing via `unscheduleAndTerminateWorker`.

mantis-control-plane/mantis-control-plane-server/src/test/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerTests.java
```diff
@@ -308,28 +309,28 @@ public void testGetBestFit_WithGenerationFromScaleGroup() {
...
                     CLUSTER_ID));
 
         assertTrue(bestFitO.isPresent());
-        assertNotEquals(te5Id, bestFitO.get().getLeft());
-        assertNotEquals(TASK_EXECUTOR_ID_1, bestFitO.get().getLeft());
+        assertNotEquals(te5Id, bestFitO.get().getBestFit().values().stream().findFirst().get().getLeft());
+        assertNotEquals(TASK_EXECUTOR_ID_1, bestFitO.get().getBestFit().values().stream().findFirst().get().getLeft());
         assertEquals(SCALE_GROUP_1,
-            Objects.requireNonNull(bestFitO.get().getRight().getRegistration())
+            Objects.requireNonNull(bestFitO.get().getBestFit().values().stream().findFirst().get().getRight().getRegistration())
```
Andyz26, line 333: it would be nice to have some more UTs around the pending worker set scenarios here.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java
Andyz26: i guess in this case there is no more retry/actions to this job so maybe we should mark the job as failed instead? So users won't feel that the job is still being handled.

Andyz26:
lgtm thanks a lot!

Andyz26 (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -317,12 +386,35 @@ public GetClusterUsageResponse getClusterUsage(GetClusterUsageRequest req) {
             } else {
                 usageByGroupKey.put(groupKey, kvState);
             }
+
+            if ((value.isAssigned() || value.isRunningTask()) && value.getWorkerId() != null) {
```
Andyz26, line 390: I think we need metrics to track the pending set size with tags.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -317,12 +386,35 @@ public GetClusterUsageResponse getClusterUsage(GetClusterUsageRequest req) {
...
+                log.info("Removing job {} from pending requests", jobId);
+                pendingJobRequests.invalidate(jobId);
+            }
         });
 
         GetClusterUsageResponseBuilder resBuilder = GetClusterUsageResponse.builder().clusterID(req.getClusterID());
         usageByGroupKey.forEach((key, value) -> resBuilder.usage(UsageByGroupKey.builder()
             .usageGroupKey(key)
-            .idleCount(value.getLeft())
+            .idleCount(value.getLeft() - pendingCountByGroupKey.get(key))
```
Andyz26, line 417: Currently, we disable the scaler during the upgrade. With this logic job scheduling might see more interruptions and do you think we need to do some handling between upgrade?

PR #582, opened 2023-11-10
Use a separate directory to store registration information

sundargates:
### Context

^^^

### Checklist


mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfigurationWritable.java
```diff
@@ -252,4 +252,9 @@ public double getNetworkBandwidthInMB() {
     public String taskExecutorAttributes() {
         return this.taskExecutorAttributesStr;
     }
+
+    @Override
+    public File getRegistrationStoreDir() {
+        return null;
+    }
```
sundargates, line 259: This will not and should not be used by the worker itself. Hence, it returns a null for the time being.

PR #584, opened 2023-11-10
SinkConnectionsStatus to also provide running workers count

fdc-ntflx:
### Context

The current issues with unreliable periods identified on go/realtimesps largely arise from scale-up operations on the PlayAPI source job. Whenever a scale-up occurs, additional workers are introduced. However, the SPS tracker perceives this as a lack of connection with all up-stream workers and consequently flags the signal as unreliable. This is largely due to the current operational logic. Specifically, each SPS worker assesses the total number of workers in the source against the number it’s connected to. Since the SPS worker cannot connect to workers in the ‘launched’ state, this results in a discrepancy exceeding 5% (since the source is always scaled up by 40 units, surpassing 5% of the total). The discussion yesterday highlighted two potential solutions: 1) the total connections should perhaps be compared with source workers that are in a connectable/running state, and 2) there might be a need to delay the process of flagging the signal as unreliable, rather than doing so immediately upon the first detection of a mismatch, which currently occurs soon after the workers are assigned post the scale-up operation.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/MasterClientWrapper.java
+++ b/mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/MasterClientWrapper.java
@@ -242,7 +242,7 @@ public class MasterClientWrapper {
                                                     .getHosts()
                                                     .values()
                                                     .stream()
-                                                    .filter(e -> MantisJobState.isRunningState(e.getState()))
+                                                    .filter(e -> MantisJobState.isOnStartedState(e.getState()))
                                                     .count();
                                                 numSinkWorkersSubject.onNext(new JobSinkNumWorkers(jobId, totalFromPartitions, runningWorkers));
                                                 if (usePartition(workerIndex, totalFromPartitions, forPartition, totalPartitions)) {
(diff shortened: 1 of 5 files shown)


Andyz26:
can you add short context to PR description?

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/MasterClientWrapper.java
```diff
@@ -238,7 +238,13 @@ public Observable<EndpointChange> getSinkLocations(final String jobId, final int
                                             for (WorkerHost host : workerAssignments.getHosts().values()) {
                                                 final int workerIndex = host.getWorkerIndex();
                                                 final int totalFromPartitions = workerAssignments.getNumWorkers();
-                                                numSinkWorkersSubject.onNext(new JobSinkNumWorkers(jobId, totalFromPartitions));
+                                                final int runningWorkers = (int) workerAssignments
+                                                    .getHosts()
+                                                    .values()
+                                                    .stream()
+                                                    .filter(e -> MantisJobState.isRunningState(e.getState()))
```
hmitnflx, line 245: isRunningState is true for launched, startInitiated, and started. I think we'd have the same problem as using totalWorkers during autoscaling in that case.

fdc-ntflx: good catch. I meant to use `isOnSlaveState` wdyt?

calvin681:
My biggest concern is that this changes the meaning of existing metrics. It may have unintended consequences that we are not aware of. Should we instead add additional metrics for STARTED source workers and then modify SPS tracker to use them.

calvin681 (approved):

PR #587, opened 2023-11-16
bump spectator library to a more recent version

hmitnflx:
### Context

### Checklist

### DIFF
--- a/mantis-publish/mantis-publish-core/build.gradle
+++ b/mantis-publish/mantis-publish-core/build.gradle
@@ -17,7 +17,7 @@ ext {
     archaiusVersion = 'latest.release'
     versionMockito = '2.18.+'
     mqlVersion = '3.2.2'
-    spectatorVersion = '0.134.0'
+    spectatorVersion = '1.6.10'
 }
 
 dependencies {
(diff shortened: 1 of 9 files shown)


mantis-publish/mantis-publish-core/build.gradle
```diff
@@ -17,7 +17,7 @@ ext {
     archaiusVersion = 'latest.release'
     versionMockito = '2.18.+'
     mqlVersion = '3.2.2'
-    spectatorVersion = '0.134.0'
```
sundargates, line 20: Can we get rid of the version here and instead use ext.libraries in the root build.gradle?

sundargates (approved):

mantis-publish/mantis-publish-core/build.gradle
```diff
@@ -17,7 +17,7 @@ ext {
...
+    spectatorVersion = '1.6.10'
```
Andyz26, line 20: just wondering if we need to actually declare spectator as "API" in this module or can it be compileOnly to let application resolve it.

calvin681: This would become a breaking change if we do that right? if an application doesn't have explicit dependency on spectator api.

Andyz26: that's true. I just assume all internal applications have a dependency on spectator already.

PR #594, opened 2023-11-29
Fix losing job on control plane restart

Andyz26:
### Context

Currently, any failure during the loading job from the persistency layer step causes the job to be dropped without proper cleanup.
Here the logic is updated as follows:
* If two duplicate workers on the same stage index are found, load the one with the newer worker number to use as active.
* The dupe one with the older worker number then gets terminated in the job actor when it sends the heartbeat and doesn't match with the active worker number.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
@@ -402,7 +402,7 @@ public class KeyValueBasedPersistenceProvider implements IMantisPersistenceProvi
                     // If there are duplicate workers on the same stage index, only attach the one with latest
                     // worker number. The stale workers will not present in the stage metadata thus gets terminated
                     // when it sends heartbeats to its JobActor.
-                    boolean addedWorker = jobMeta.addWorkerMedata(workerMeta.getStageNum(), workerMeta);
+                    boolean addedWorker = jobMeta.tryAddOrReplaceWorker(workerMeta.getStageNum(), workerMeta);
                     if (!addedWorker) {
                         staleWorkersFoundCounter.increment();
                     }
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
@@ -49,7 +49,6 @@ import io.mantisrx.shaded.com.fasterxml.jackson.databind.DeserializationFeature;
 import io.mantisrx.shaded.com.fasterxml.jackson.databind.ObjectMapper;
 import io.mantisrx.shaded.com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import io.mantisrx.shaded.com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
-import io.mantisrx.shaded.com.google.common.base.Preconditions;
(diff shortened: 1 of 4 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
```diff
@@ -394,18 +399,18 @@ public List<IMantisJobMetadata> loadAllJobs() throws IOException {
...
-                        jobMeta.addWorkerMedata(workerMeta.getStageNum(), workerMeta, null),
-                        "JobID=%s stage=%d workerIdx=%d has existing worker, existing=%s, new=%s",
-                        workerMeta.getJobId(),
-                        workerMeta.getStageNum(),
-                        workerMeta.getWorkerIndex(),
-                        jobMeta.getWorkerByIndex(workerMeta.getStageNum(), workerMeta.getWorkerIndex()).getWorkerId(),
-                        workerMeta.getWorkerId());
+                    // If there are duplicate workers on the same stage index, only attach the one with latest
+                    // worker number. The stale workers will not present in the stage metadata thus gets terminated
+                    // when it sends heartbeats to its JobActor.
```
sundargates, line 404: It looks like this logic is being decided at the MantisJobWorkerMetadataWritable level. Could we instead do this here? We already have two different building blocks today - one to add and another to replace. Let's use them instead. 

It's also not clear to me what the boolean represents in this new method call.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
```diff
@@ -394,18 +399,18 @@ public List<IMantisJobMetadata> loadAllJobs() throws IOException {
...
+                    boolean addedWorker = jobMeta.addWorkerMedata(workerMeta.getStageNum(), workerMeta);
+                    if (!addedWorker) {
+                        staleWorkersFoundCounter.increment();
+                    }
                 }
                 jobMetas.add(DataFormatAdapter.convertMantisJobWriteableToMantisJobMetadata(jobMeta, eventPublisher));
             } catch (Exception e) {
-                logger.warn("Exception loading job {}", jobId, e);
+                logger.error("Exception loading job {}", jobId, e);
```
sundargates, line 412: Should this be critical instead?

sundargates:
Can you add some tests to test this behavior?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
Andyz26: "one to add and another to replace,": i think it's just different names used for different components. I only find one path to add/replace workers in the code.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
Andyz26: there is no critical level at logger?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/KeyValueBasedPersistenceProvider.java
Andyz26: I can do "addWorkerMedata" -> "tryAddOrReplaceWorker"

Andyz26:
> Can you add some tests to test this behavior?

UT added in the 2nd commit

PR #599, opened 2023-12-11
Fix duplicate TE allocation to multiple workers with different resource spec matching the same SKU 

fdc-ntflx:
### Context

This PR addresses a problem encountered with TE allocation in scenarios where jobs have TEs with different resource specs that match the same SKU. For instance, if a job contains a TE with 1 core and another with 2 cores, and the smallest SKU available is a 2 core, both TEs will match the smallest SKU. This situation can lead to the same TE being incorrectly allocated to different workers, as currently the logic assumes that resource spec always matches the SKU size (which in a few cases it's not the case).

In order to address this issue, we now pass the list of already assigned TEs over to the matching logic, ensuring we aren't reusing them for other worker requests.

### Checklist


sundargates (approved):
lgtm!

Issue #605, opened 2023-12-13
Crashing containers following official docker-compose getting started guide


jpittis:
### Summary

Hey folks. I'm encountering crashing containers while following the official [docker-compose guide getting started guide](https://netflix.github.io/mantis/getting-started/tutorials/docker/):

```
$ mkdir mantis
$ cd mantis
$ wget https://raw.githubusercontent.com/Netflix/mantis/master/docker-compose.yml
$ docker-compose -f docker-compose.yml up
```

I can see logs showing that mantisagent, mantismaster, and mantisapi all exit with non-zero status codes (in that order):

```
mantis-compose-mantisagent-1 exited with code 2
mantis-compose-mantismaster-1 exited with code 1
mantis-compose-mantisapi-1 exited with code 1
```

I'm running this on OSX Sonoma (aarch64), but have been able to reproduce on Linux (amd64) as well.

### Details

Right before the crash of the agent and the master, I see the following log line from ZooKeeper:

```
mantis-compose-zookeeper-1         | 2023-12-13 01:40:04,582 [myid:] - WARN  [NIOServerCxn.Factory:0.0.0.0/0.0.0.0:2181:ZooKeeperServer@675] - Received packet at server of unknown type 19
```

Other potential culprit logs (there are too many to dump all of them here) include:

```
mantis-compose-mantisagent-1       | 2023-12-13 01:40:03 ERROR AgentV2Main:114 - Unexpected error: java.lang.IllegalStateException: Expected the service TaskExecutorStarter [FAILED] to be RUNNING, but the service has FAILED
```

```
mantis-compose-mantisagent-1       | Caused by: io.mantisrx.shaded.org.apache.zookeeper.KeeperException$UnimplementedException: KeeperErrorCode = Unimplemented for /mantis/master
```

```
mantis-compose-mantismaster-1      | 2023-12-13 01:40:04 ERROR MasterMain:269 - caught exception on Mantis Master initialization
mantis-compose-mantismaster-1      | java.lang.RuntimeException: java.lang.IllegalStateException: Expected the service ZookeeperMasterMonitor [FAILED] to be RUNNING, but the service has FAILED
```

```
mantis-compose-mantismaster-1      | Caused by: io.mantisrx.shaded.org.apache.zookeeper.KeeperException$UnimplementedException: KeeperErrorCode = Unimplemented for /mantis/master

jpittis:
[...]

Issue #604, opened 2023-12-13
Encountering errors building Mantis from source following official docs


jpittis:
## Summary

Hey folks. I'm having some difficulty building the complete Mantis project, as well as its examples from source.

I assume this is my environment being incorrectly configured rather than an issue with the master branch.

It would be awesome if we could figure out what's missing and update the docs to include those dependencies.

Thanks for taking a look!

## Reproduction

I get errors when following the build instructions in the README:

```
$ git clone git@github.com:Netflix/mantis.git
$ cd mantis
$ ./gradlew clean build
BUILD FAILED in 4s
```

I get similar errors when following the instructions from [the "Local" tutorial](https://netflix.github.io/mantis/getting-started/tutorials/local/):

```
$ git clone git@github.com:Netflix/mantis.git
$ cd mantis/mantis-examples/mantis-examples-synthetic-sourcejob
$ ../../gradlew execute
BUILD FAILED in 3s
```

I've demonstrated my environment is at least somewhat correctly setup by successfully building [mantis-mql](https://github.com/Netflix/mantis-mql) from source:

```
$ git clone https://github.com/Netflix/mantis-mql
$ ./gradlew clean build
BUILD SUCCESSFUL in 52s
```

Here's some metadata around what I believe is being used to build Mantis:

```
$ ./gradlew --version

------------------------------------------------------------
Gradle 7.5
------------------------------------------------------------

Build time:   2022-07-14 12:48:15 UTC
Revision:     c7db7b958189ad2b0c1472b6fe663e6d654a5103

Kotlin:       1.6.21
Groovy:       3.0.10
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          17.0.6 (Eclipse Adoptium 17.0.6+10)
OS:           Mac OS X 14.1.2 aarch64
```

## Details

It looks like at least two tasks are failing with quite loud stacktraces which I'm partially summarizing below:

```
> Task :mantis-common-serde:spotlessJava FAILED
Step 'removeUnusedImports' found problem in 'mantis-common-serde/src/main/java/io/mantisrx/common/JsonSerializer.java':
null
java.lang.reflect.InvocationTargetException
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
...
        at java.base/java.lang.Thread.run(Thread.java:833)
[...]

PR #606, opened 2023-12-13
Bump docker-compose ZooKeeper version to 3.5.5 to fix compatability bug.

jpittis:
### Context

https://github.com/Netflix/mantis/issues/605 describes a fatal issue with the existing docker-compose file that I resolved with the contents of this PR. Based on conversations in https://github.com/yahoo/CMAK/issues/748, it sounds like Curator is no longer compatible with ZooKeeper 3.4.

### How to reproduce and test

The following commands will reproduce the issue, which can then be resolved by using this commit instead:

```
$ mkdir mantis
$ cd mantis
$ wget https://raw.githubusercontent.com/Netflix/mantis/master/docker-compose.yml
$ docker-compose -f docker-compose.yml up
```

You'll likely need to prune your volumes when switching between versions to allow for the ZooKeeper upgrade.

### Checklist

    - I'm unable to build Mantis as described in https://github.com/Netflix/mantis/issues/604.
    - I'm unable to build Mantis as described in https://github.com/Netflix/mantis/issues/604.

### DIFF
--- a/docker-compose-local.yml
+++ b/docker-compose-local.yml
@@ -10,7 +10,7 @@ version: '3'
 services:
 
   zookeeper:
-    image: jplock/zookeeper:3.4.6
+    image: jplock/zookeeper:3.5.5
     ports:
       - "2181:2181"
     networks:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,7 @@ version: '3'
 services:
 
   zookeeper:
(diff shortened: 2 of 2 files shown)


jpittis:
Closing as per https://github.com/Netflix/mantis/issues/605. Helm should be used instead of docker-compose. We should update the docs to point to helm, and remove these outdated docker-compose files.

PR #607, opened 2023-12-13
Update getting started local tutorial to clarify required JDK version.

jpittis:
### Context

See https://github.com/Netflix/mantis/issues/604 for context. The Mantis project does not presently compile will all JDK versions larger or equal to 8.

Note that I haven't tested for at which point the incompatibility begins, just that JDK17 doesn't work.

I added a tip box for the tip about JDK selection tooling that @calvin681 gave me in https://github.com/Netflix/mantis/issues/604.

I also ran through the rest of the guide and confirmed it's functional (with a slight adjustment to the wording around the port number selection incase that trips anyone up).

### Checklist

### DIFF
--- a/docs/docs/getting-started/tutorials/local.md
+++ b/docs/docs/getting-started/tutorials/local.md
@@ -2,7 +2,10 @@
 
 ## Prerequisites
 
-JDK 8 or higher
+JDK 8 (newer versions may not be supported)
+
+!!! tip
+    Some users find https://sdkman.io/ helpful for managing multiple Java versions.
 
 ## Build and run the synthetic-sourcejob sample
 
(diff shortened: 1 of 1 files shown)


docs/docs/getting-started/tutorials/local.md
```diff
@@ -2,7 +2,10 @@
 
 ## Prerequisites
 
-JDK 8 or higher
+JDK 8 (newer versions may not be supported)
```
sundargates, line 5: It would be better to use the java toolchain support in gradle to define this directly and get rid of these instructions altogether. https://docs.gradle.org/current/userguide/toolchains.html

PR #622, opened 2024-01-25
[Runtime] Remove mesos deps and refactor layers

Andyz26:
### Context
[Warning] This is a breaking change and required builds from this commit to be marked as 3.x

* Remove mesos dependencies in the runtime package.
* Refactor the dependencies on user runtime layers as below.

RuntimeV2PR

### Checklist


mantis-examples/mantis-examples-sine-function/build.gradle
```diff
@@ -27,6 +27,6 @@ task execute(type:JavaExec) {
     classpath = sourceSets.main.runtimeClasspath
 }
 dependencies {
-    implementation project(':mantis-server:mantis-server-worker')
+    implementation project(':mantis-runtime-executor')
```
sundargates, line 30: I think we should remove this completely and have it part of runtime.

So, essentially, we will have a runtime loader, which has the API to load the runtime and then the runtime itself.

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/cgroups/CgroupsMetricsCollector.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.mantisrx.server.agent.metrics.cgroups;
+package io.mantisrx.runtime.loader.cgroups;
```
sundargates, line 17: Why is this part of the API module? I thought the loader is just the API between runtime and runtime-loader.

Andyz26: the Cgroups metrics collector is used on both the agent side and the executor side thus they need to be at a shared module (before the server-worker depends on the server-agent directly and I am removing that dependency)>

PR #623, opened 2024-01-29
Introducing Allocation Attributes for Container Scheduling.

fdc-ntflx:
### Introduction

The objective of this PR is to facilitate running multiple versions of Task Executors, which are uniformly sized but differ in their runtime. A practical example can be operating two sets of large containers - one on jdk8 and another on jdk17. Moreover, as part of Mantis cost-optimization efforts, it's important to support multiple container sizes - using equivalent cores but varying memory configurations.

### Details

This PR enables running multiple versions of similar-sized Task Executors by differentiating using factors like runtime versions or network configurations. The system should allow thoughtful scheduling of workers based on resources and differentiating factors. For instance, it could operate "large" containers in two environments, JDK 8 and JDK 17.

To describe varied Task Executors of the same size, the proposal suggests creating multiple SKUs and storing variables like runtime JDK in existing SKU metadata field. This approach leads to an increasing list of SKU-IDs to be shown to users during worker size selection. To keep the selection user-friendly, it's proposed to enrich SKUs with a size object which includes a 't-shirt size' label.

Worker scheduling logic has to change. Now, the system can define stage sizes in a scheduling request through the existing machine definition or the t-shirt size name in the SKU spec. An updated form of scheduling request will be introduced that encompasses resource definitions and assignment attributes.

Task Executors will continue to register in the system with full metadata to allow on-demand scheduling. But now registration will provide allocation attributes and size name to allow the new form of scheduling discussed above.

Regarding ASG provisioning, the challenge is to ensure that both scheduling constraints and new size definition are forwarded to the ASG provisioning workflow. No changes are required for interaction between the scaler actor and the resource cluster actor as it is already centered around SKU-ID.

Under job submission, the request needs to specify both resource and scheduling constraints. Resources can be represented traditionally using an explicit machine definition or through the new size names. The plan also facilitates streamlining the API by separating job submissions that use the machine definition from those that use size labels.

For job cluster creation and updates, labels will indicate allocation attributes, while scheduling information can express the resource definition for different stages. The system dynamically merges job cluster definition and artifact-based attributes, helping the scheduler to make informed scheduling decisions.

### Backward Compatibility

**This change is meant to be backward compatible.**

### Checklist


mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/SkuSizeID.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package io.mantisrx.server.master.resourcecluster;
+
+import lombok.RequiredArgsConstructor;
+import lombok.Value;
+
+@RequiredArgsConstructor(staticName="of")
+@Value
+public class SkuSizeID {
```
sundargates, line 24: nit: You might want to make the constructor access private if you expose the static constructor. 
```
access = AccessLevel.PRIVATE
```

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
```diff
@@ -113,20 +113,14 @@ public boolean containsAttributes(Map<String, String> requiredAttributes) {
     }
 
     @JsonIgnore
-    public Optional<ContainerSkuID> getTaskExecutorContainerDefinitionId() {
+    public ContainerSkuID getTaskExecutorContainerDefinitionId() {
```
sundargates, line 116: what happens if this is not present? Also, how would this work when we have on-demand TEs?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/SkuSizeSpec.java
```diff
@@ -0,0 +1,62 @@
...
+
+    String skuSizeName; // ie. small
+
+    int cpuCoreCount;
+
+    int memorySizeInMB;
+
+    int networkMbps;
+
+    int diskSizeInMB;
```
sundargates, line 41: nit: use MachineDefinition.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/SkuSizeSpec.java
```diff
@@ -0,0 +1,62 @@
...
+import lombok.Value;
+
+@Builder
+@Value
+@EqualsAndHashCode(onlyExplicitlyIncluded = true)
+public class SkuSizeSpec {
+    @EqualsAndHashCode.Include
+    SkuSizeID skuSizeID; // ie. small-v0, small-v1
+
+    String skuSizeName; // ie. small
```
sundargates, line 33: nit: since it's scoped inside the SKUSizeSpec class we can just call this name.

mantis-common/src/main/java/io/mantisrx/runtime/AllocationConstraints.java
```diff
@@ -0,0 +1,35 @@
...
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.ToString;
+
+@RequiredArgsConstructor
+@Getter
+@EqualsAndHashCode
+@AllArgsConstructor(staticName = "of")
+@ToString
+public class AllocationConstraints {
```
sundargates, line 31: nit: SchedulingConstraints feels more apt.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobDefinition.java
```diff
@@ -63,6 +66,8 @@ public class JobDefinition {
     private final int withNumberOfStages;
     private Map<String, Label> labels; // Map label->name to label instance.
 
+    private final static String ALLOCATION_CONSTRAINT_LABEL_REGEX = "_mantis\\.allocationConstraint\\.(.+)";
```
sundargates, line 69: nit: schedulingConstraint

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobDefinition.java
```diff
@@ -246,6 +251,21 @@ public Optional<ClusterID> getResourceCluster() {
             .map(l -> ClusterID.of(l.getValue()));
     }
 
+    public Map<String, String> getAssignmentAttributes() {
+        final Pattern pattern = Pattern.compile(ALLOCATION_CONSTRAINT_LABEL_REGEX);
```
sundargates, line 255: nit: make this static.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
```diff
@@ -1586,7 +1586,7 @@ private ScheduleRequest createSchedulingRequest(
                                 mantisJobMetaData.getMinRuntimeSecs()
                         ),
                         mantisJobMetaData.getSla().orElse(new JobSla.Builder().build()).getDurationType(),
-                        stageMetadata.getMachineDefinition(),
+                        AllocationConstraints.of(stageMetadata.getMachineDefinition(), mantisJobMetaData.getJobDefinition().getAssignmentAttributes()),
```
sundargates, line 1589: nit: stageMetadata.getSchedulingConstraints() would be cleaner.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -61,37 +62,53 @@
 @Slf4j
 class ExecutorStateManagerImpl implements ExecutorStateManager {
     private final Map<TaskExecutorID, TaskExecutorState> taskExecutorStateMap = new HashMap<>();
-    Cache<String, JobRequirements> pendingJobRequests = CacheBuilder.newBuilder()
+
+    /**
+     * Cache the available executors ready to accept assignments. Note these executors' state are not strongly
+     * synchronized and requires state level check when matching.
+     */
```
sundargates, line 69: Why is strong synchronization needed if it's only used by the actor?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -61,37 +62,53 @@
...
+    private final ConcurrentHashMap<ContainerSkuID, NavigableSet<TaskExecutorHolder>> executorsBySkuID = new ConcurrentHashMap<>();
```
sundargates, line 70: It seems like using a ConcurrentHashMap may not be necessary in this case, as it is meant to manage actor state.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
fdc-ntflx: TEs won't be able to register anymore without skuId as part of the attributes. Is that a problem? 

how would this work when we have on-demand TEs? I guess it won't work for now. Are we interested in making things work for on-demand as well?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorAllocationHelper.java
```diff
@@ -0,0 +1,152 @@
...
+        return bestFitSkuID == null ? Optional.empty() : Optional.of(bestFitSkuID);
+    }
+
+    public void addSkuIdIfAbsent(ContainerSkuID skuID) {
+        if (!skus.containsKey(skuID)) {
+            refreshMetadata(skuID);
+        }
+    }
+
+    public void removeSkuIdIfPresent(ContainerSkuID skuID) {
```
Andyz26, line 74: is remove a legit scenario?

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfiguration.java
```diff
@@ -159,7 +159,7 @@ default Time getHeartbeatInterval() {
     @Default(value = "128.0")
     double getNetworkBandwidthInMB();
 
-    @Config("mantis.taskexecutor.attributes")
+    @Config({"mantis.taskexecutor.attributes", "MANTIS_TASKEXECUTOR_ATTRIBUTES"})
```
sundargates, line 162: nit: why do we need to change this?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorAllocationHelper.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+
+public class TaskExecutorAllocationHelper {
+    private final ClusterID clusterID;
+    private final Map<String, String> allocationAttributesWithDefaults;
+    private final IMantisPersistenceProvider storageProvider;
+
+    private Map<ContainerSkuID, SkuTypeSpec> skus = new ConcurrentHashMap<>();
+    private Map<SkuSizeID, SkuSizeSpec> skuSizes = new ConcurrentHashMap<>();
```
sundargates, line 41: nit: Assuming this represents the Actor state, we don't need explicit synchronization.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorAllocationHelper.java
```diff
@@ -0,0 +1,152 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+
+public class TaskExecutorAllocationHelper {
```
sundargates, line 35: nit: package private.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorAllocationHelper.java
```diff
@@ -0,0 +1,152 @@
...
+        final SkuSizeSpec size1 = skuSizes.get(getSkuSizeIdFrom(skus.get(sku1)));
+        if (size0.getCpuCoreCount() < size1.getCpuCoreCount() || (
+            size0.getCpuCoreCount() == size1.getCpuCoreCount() && size0.getMemorySizeInMB() < size1.getMemorySizeInMB()
+        )) {
+            return sku0;
+        }
+        return sku1;
+    }
+
+    private void refreshMetadata(ContainerSkuID skuID) {
```
sundargates, line 111: Hitting the database to refresh state is not recommended with actors, which are expected to manage the state. Can we manage this state either within the ResourceClusterActor or another actor that you think is more appropriate for this?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorAllocationHelper.java
```diff
@@ -0,0 +1,152 @@
...
+
+        for (SkuTypeSpec sku : skus.values()) {
+            if (canFit(sku.getSkuId(), allocationConstraints)) {
+                if (bestFitSkuID != null) {
+                    bestFitSkuID = findBestFit(bestFitSkuID, sku.getSkuId());
+                } else {
+                    bestFitSkuID = sku.getSkuId();
+                }
+            }
+        }
```
sundargates, line 64: nit: If a fitness method could give you a score between 0.0 and 1.0, where 0.0 means unfit and 1.0 means a perfect fit, it would simplify things and eliminate the need for multiple methods.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorAllocationHelper.java
```diff
@@ -0,0 +1,152 @@
...
+
+    private ContainerSkuID findBestFit(ContainerSkuID sku0, ContainerSkuID sku1) {
+        final SkuSizeSpec size0 = skuSizes.get(getSkuSizeIdFrom(skus.get(sku0)));
+        final SkuSizeSpec size1 = skuSizes.get(getSkuSizeIdFrom(skus.get(sku1)));
+        if (size0.getCpuCoreCount() < size1.getCpuCoreCount() || (
+            size0.getCpuCoreCount() == size1.getCpuCoreCount() && size0.getMemorySizeInMB() < size1.getMemorySizeInMB()
+        )) {
+            return sku0;
+        }
+        return sku1;
```
sundargates, line 108: nit: It seems appropriate to make "fitting another SkuSizeSpec" a property of SkuSizeSpec. Additionally, if we can utilize the existing MachineDefinition data structure within the SkuSizeSpec, we could simply use the "canFit" method of MachineDefinition.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorAllocationHelper.java
```diff
@@ -0,0 +1,152 @@
...
+                    }
+                });
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private SkuSizeID getSkuSizeIdFrom(SkuTypeSpec skuTypeSpec) {
+        return skuTypeSpec.getSizeId() == null ? SkuSizeID.of("skuTypeSpec-" + skuTypeSpec.getSkuId().getResourceID()): skuTypeSpec.getSizeId();
+    }
```
sundargates, line 151: This logic could be moved to SkuTypeSpec to provide a standard way of handling it, instead of being deeply baked into assumptions.

mantis-common/src/main/java/io/mantisrx/runtime/AllocationConstraints.java
```diff
@@ -0,0 +1,35 @@
...
+package io.mantisrx.runtime;
+
+import java.util.Map;
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.ToString;
+
+@RequiredArgsConstructor
```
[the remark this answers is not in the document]
fdc-ntflx: doesn't need to. let me move it.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
Andyz26: @fdc-ntflx can you confirm (maybe add to the PR context) about the decision of whether or not going fwd we still want to support any on-demand/bring-your-own-container kind of scenarios?

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
```diff
@@ -113,20 +113,14 @@ public boolean containsAttributes(Map<String, String> requiredAttributes) {
...
         // handle back compat on key case insensitivity.
         String containerDefIdLower = WorkerConstants.WORKER_CONTAINER_DEFINITION_ID.toLowerCase();
         if (this.taskExecutorAttributes.containsKey(containerDefIdLower)) {
-            return Optional.ofNullable(ContainerSkuID.of(this.getTaskExecutorAttributes().get(containerDefIdLower)));
+            return ContainerSkuID.of(this.getTaskExecutorAttributes().get(containerDefIdLower));
```
Andyz26, line 120: what's the release plan to prod master/agents for this breaking change?

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/SkuSizeID.java
Andyz26, line 24: Add doc about it's expected usage and format.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/ResourceClusterRouteHandler.java
```diff
@@ -54,4 +56,9 @@ CompletionStage<GetResourceClusterScaleRulesResponse> createAllScaleRule(
 
     CompletionStage<GetResourceClusterScaleRulesResponse> getClusterScaleRules(
         GetResourceClusterScaleRulesRequest request);
+
+    CompletionStage<GetResourceClusterSkuSizesResponse> createClusterSkuSize(
```
Andyz26, line 60: how is this being used? isn't adding a new skuSize meaning updating the RC spec?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/SkuSizeSpec.java
fdc-ntflx: the machine definition includes the ports number which wouldn't make sense to set at the sku level.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobDefinition.java
fdc-ntflx: why?

sundargates: Why is the pattern compiled every time the assignment attributes are requested when the regex itself is static?

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
fdc-ntflx: I've been thinking about this a lot and decided to leave the semantics as is today. This is the reasoning. 

For the scheduler to be capable of making informed scheduling decisions, it is crucial that TEs are registered in the system with both the machine definition and scheduling constraints. Two possible strategies have been evaluated:

1. Requiring TEs to provide the SKU-ID they are associated with at the time of registration
2. Permitting TEs to communicate their scheduling constraints during registration

Each of these options carries its own set of benefits and drawbacks. The first strategy simplifies the system through a SKU-ID-centric approach, yet it also represents a major constraint, limiting TEs to register only when a predefined resource cluster is present. This restriction leads to significant modifications in mantis-helm as well and necessitates the deployment of a specialized resource cluster provider in the oss. 

The second strategy, though it may not entirely resolve the ambiguity regarding SKU-ID usage, does have the advantage of preserving existing system characteristics. This includes detaching TE registration from resource cluster definition, enabling TEs to register whenever needed, and scheduling workers "on-demand". These features eliminate the need for additional modifications in mantis-helm or introducing new providers in the oss. However, a downside of this approach is that it mandates a redeployment of agents each time new scheduling constraints are introduced or existing ones need tweaking, ensuring they match the revised scheduling constraints.

Option 2 is the one we're going with.

fdc-ntflx: @Andyz26 I also updated the internal design doc with the same section. Let me know if that's clear now.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobDefinition.java
fdc-ntflx: Sorry I thought you were referring to the entire method. +1

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
fdc-ntflx: good feedback but I'm trying to avoid changing too many things at once.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
fdc-ntflx: this was a prev comment. let me update it.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
fdc-ntflx: +1

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorAllocationHelper.java
fdc-ntflx: I removed that logic all together.

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfiguration.java
fdc-ntflx: no need.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorAllocationHelper.java
fdc-ntflx: agreed

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorAllocationHelper.java
fdc-ntflx: +1 on using Machine definition for this purpose.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/persistence/IMantisPersistenceProvider.java
```diff
@@ -209,4 +210,8 @@ ResourceClusterScaleRulesWritable registerResourceClusterScaleRule(
         ResourceClusterScaleRulesWritable ruleSpec) throws IOException;
 
     ResourceClusterScaleRulesWritable registerResourceClusterScaleRule(ResourceClusterScaleSpec rule) throws IOException;
+
+    List<SkuSizeSpec> registerResourceClusterSkuSize(SkuSizeSpec skuSizeSpec) throws IOException;
+
+    List<SkuSizeSpec> getResourceClusterSkuSizes() throws IOException;
```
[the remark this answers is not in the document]
fdc-ntflx: updated the model, so we don't need to store sku sizes separately anymore.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
fdc-ntflx: this is not changing anymore, so there is no problem during rollout.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/SkuSizeID.java
fdc-ntflx: class was dropped.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/ResourceClusterRouteHandler.java
fdc-ntflx: yeah. no need for this anymore.

Andyz26:
since "http://go/mantis-scheduling-constraints" is internal, it could be nice to capture the main behavior/requirement change bullet points here too.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/scheduler/SchedulingConstraints.java
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @param constraints - SchedulingConstraints object, carries machine definition and assignment attributes.
+     * @param assignmentAttributesAndDefaults - Map containing assignment attributes to be verified and their default values.
+     *
+     * @return - The fitness value ranging between 0 and 1. Returns 0 if the provided assignment attributes
+     * do not match those of this instance or if the machine definition doesn't fit into the current instance's
+     * machine definition. Conversely, it returns a value close to 1 indicating higher similarity between
+     * the machine definitions.
+     */
+    public double fitness(SchedulingConstraints constraints, Map<String, String> assignmentAttributesAndDefaults) {
```
Andyz26, line 55: maybe separate the different fit check?

Andyz26: (for better logging too)

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
```diff
@@ -1598,6 +1604,27 @@ private ScheduleRequest createSchedulingRequest(
...
+        /**
+         * Merges attributes assignment between job and artifact definitions. It does it by first fetching
+         * the associated JobArtifact tags using the artifact ID from the job, and then merging them with the assignment
+         * attributes from the job definition itself. The keys from the job definition take precedence over the
+         * keys from the artifact's tags.
+         *
+         * @param artifactID the artifact used by the job whose attributes are to be fetched
+         * @return A merged map of assignment attributes. The precedence of keys follows: job definition > artifact's tags.
+         */
+        private Map<String, String> mergeJobDefAndArtifactAssigmentAttributes(ArtifactID artifactID) {
```
Andyz26, line 1616: should this happen and persist at the job cluster level so the user can get the result of the final attributes?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -61,37 +61,55 @@
 @Slf4j
 class ExecutorStateManagerImpl implements ExecutorStateManager {
     private final Map<TaskExecutorID, TaskExecutorState> taskExecutorStateMap = new HashMap<>();
-    Cache<String, JobRequirements> pendingJobRequests = CacheBuilder.newBuilder()
+
+    /**
+     * Cache the available executors ready to accept assignments
+     */
+    private final Map<SchedulingConstraints, NavigableSet<TaskExecutorHolder>> executorsBySchedulingConstraints = new HashMap<>();
```
Andyz26, line 68: SchedulingConstraints as key here imply the machine def needs to match the schedule request exactly which is error prone and might not be fully back compact on existing jobs not exactly matching on network/disk?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/MantisResourceClusterSpec.java
```diff
@@ -98,8 +101,12 @@ public static class SkuTypeSpec {
         int diskSizeInMB;
 
         @Singular
-        Map<String, String> skuMetadataFields;
+        Map<String, String> skuMetadataFields; // ie. jdk17, sbn3
 
+        @Nullable
```
Andyz26, line 106: nullable?

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/scheduler/SchedulingConstraints.java
fdc-ntflx: can you elaborate?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
fdc-ntflx: It could be done but I don't know how worth it is given that the final attributes are compute at job submission time anyway since some attributes can still be provided at that time.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
fdc-ntflx: I don't see the problem with changing network/disk and have a separate key for newly registered TEs. Assignment will be done in either the old or new set of TEs. am I missing anything?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/MantisResourceClusterSpec.java
fdc-ntflx: yeah for backward compatibility.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
Andyz26: I think I am more referring to that SchedulingConstraints is a contract/data class used across APIs while its key (toString) here is also used as the tracking logic inside the internal of the executorManager. Thus any change on this class can impact the tracking mechanism in executorManager, which is a coupling that might cause unintentional behavior to any changes around it.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
Andyz26: I feel like we want to display some result back to users so it's visible to them what is actually being used. Is there any info can be used to show such info on UI?

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/scheduler/SchedulingConstraints.java
Andyz26: see the other comments regarding the usage of this class.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
```diff
@@ -141,4 +144,30 @@ public Optional<String> getAttributeByKey(String attributeKey) {
...
+     * The keys in allocationAttributesWithDefault denote the necessary attributes for scheduling, while the values act as
+     * the fallback options when a specific attribute value isn't located in an external attribute store.
+     *
+     * @param allocationAttributesWithDefault The keys are allocations attributes that will be checked during scheduling and
+     *                                        the values are possible defaults in case the keys are not present in the external
+     *                                        attribute store.
+     * @return The SchedulingConstraints which is based on the machineDefinition and the processed allocation attributes.
+     */
+    @JsonIgnore
+    public SchedulingConstraints getSchedulingConstraints(Map<String, String> allocationAttributesWithDefault) {
```
Andyz26, line 168: let's have a convention on what the attribute look like to be compatible with gradle/envVar.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
```diff
@@ -141,4 +147,35 @@ public Optional<String> getAttributeByKey(String attributeKey) {
...
+     * its corresponding value is used. If a key is not found, it uses the associated
+     * default value from `allocationAttributesWithDefault`.
+     *
+     * @param allocationAttributesWithDefault A Map where keys are resource allocation attributes required for scheduling,
+     *                                        and values are default options used when specific attributes are not present in the registration's attributes.
+     *
+     * @return A `SchedulingConstraints` object, built from the `SizeDefinition` and the processed `allocationAttributes`.
+     */
+    @JsonIgnore
+    public SchedulingConstraints getSchedulingConstraints(Map<String, String> allocationAttributesWithDefault) {
```
sundargates, line 168: I'm wondering why a Task Executor (TE) would have a method that returns its scheduling constraints. Wouldn't it be better to have a Fitness Calculator that takes into account both the Scheduling Constraints of the task that needs to be scheduled and the Task Executor metadata (TE Registration in this case) and then tries to find how good the match is?

fdc-ntflx:
Closing in favor of referenced PRs above where we split the entire logic into smaller chunks.

PR #624, opened 2024-01-30
Disable mesos job creation on control plane

Andyz26:
### Context

* Set default cluster id from config during job cluster creation if rc info is missing.
* Remove Mesos services and scheduler.

### Checklist


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/proto/JobClusterProtoAdapter.java
```diff
@@ -77,7 +79,7 @@ public static final CreateJobClusterRequest toCreateJobClusterRequest(final Name
                 jd.getMigrationConfig(),
                 jd.getIsReadyForJobMaster(),
                 jd.getParameters(),
-                processLabels(jd)
+                processLabels(jd, defaultResourceClusterId)
```
sundargates, line 82: Can we change the NamedJobDefinition to include the resourceClusterId? It would make the code look cleaner.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/proto/JobClusterProtoAdapter.java
```diff
@@ -264,6 +268,10 @@ public static JobClusterInfo toJobClusterInfo(MantisJobClusterMetadataView jobCl
     }
 
     protected static List<Label> processLabels(MantisJobDefinition jd) {
+        return processLabels(jd, null);
+    }
+
+    protected static List<Label> processLabels(MantisJobDefinition jd, final String defaultResourceClusterId) {
```
sundargates, line 274: same as above for MantisJobDefinition.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/config/MasterConfiguration.java
```diff
@@ -51,6 +51,10 @@ public interface MasterConfiguration extends CoreConfiguration {
     @Config("mantis.master.resourceClusterProvider")
     String getResourceClusterProvider();
 
+    @Config("mantis.master.resourceClusterDefaultId")
+    @Default("mantisagent")
+    String getDefaultResourceClusterId();
+
```
sundargates, line 57: hmm, as mentioned elsewhere, why not make resourceClusterId a first class citizen of the JobDefinition / JobClusterDefinition?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/MantisSchedulerFactoryImpl.java
```diff
@@ -39,13 +37,12 @@ public class MantisSchedulerFactoryImpl implements MantisSchedulerFactory {
...
     private final JobMessageRouter jobMessageRouter;
-    private final SchedulingService mesosSchedulingService;
     private final MasterConfiguration masterConfiguration;
     private final MetricsRegistry metricsRegistry;
     private final Map<ClusterID, MantisScheduler> actorRefMap = new ConcurrentHashMap<>();
 
     @Override
-    public MantisScheduler forClusterID(@Nullable ClusterID clusterID) {
+    public MantisScheduler forClusterID(ClusterID clusterID) {
         Optional<ClusterID> clusterIDOptional = Optional.ofNullable(clusterID);
```
sundargates, line 46: If a field is not marked @nullable, it should assumed to be not null. We can get rid of this check here.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/proto/JobClusterProtoAdapter.java
Andyz26: this change is to override with a default to be back compact with current API behavior so anyone using existing API will not fall back to mesos stack. Since we are already having a internal "_mantis.resourceCluster" label being used I guess what you propose is more like a v2 API for job cluster APIs we can follow up with?

sundargates: I don't think it's necessary to change the way data is stored within a JobDefinition to access this property. Instead, I suggest adding a getResourceCluster() function to the JobDefinition class that always returns a valid resource cluster ID. This way, we avoid duplicating the logic in multiple locations. Eventually, we can update the data structure to store the property in a different manner, as you suggested.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/MantisSchedulerFactoryImpl.java
```diff
@@ -39,17 +36,15 @@ public class MantisSchedulerFactoryImpl implements MantisSchedulerFactory {
...
     private final Map<ClusterID, MantisScheduler> actorRefMap = new ConcurrentHashMap<>();
 
     @Override
-    public MantisScheduler forClusterID(@Nullable ClusterID clusterID) {
-        Optional<ClusterID> clusterIDOptional = Optional.ofNullable(clusterID);
-        if (clusterIDOptional.isPresent()) {
-            if (Strings.isNullOrEmpty(clusterIDOptional.get().getResourceID())) {
-                log.error("Received empty resource id: {}", clusterIDOptional.get());
+    public MantisScheduler forClusterID(ClusterID clusterID) {
+        if (clusterID != null) {
```
sundargates, line 45: when will this be null given that the parameter is not marked nullable?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/config/MasterConfiguration.java
sundargates, line 57: Do we still need this?

sundargates (approved):
This is a great change! Thank you for wrapping this up!

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/MantisSchedulerFactoryImpl.java
Andyz26: never. (so null case is just log and throw).

PR #626, opened 2024-02-07
Allow failures setting nextWorkerNumberToUse

hmitnflx:
Issue:
Sometimes, the underlying store isn't available for the extensive writes made to update nextWorkerNumberToUse. An exception is thrown cascaded upstream preventing further writes and updates to worker from JobActor. To handle it, we tolerate some (consecutive) write failures for this call before propagating the exception upstream.

Invariant:
This is fine from a constraints perspective. Here's why: On new mantis master leader election, job and worker state is read from the DB. If the worker count was synced correctly no issue. If not and there were extra workers than workerMax stored in DB, those extra workers will be killed when TEs send heartbeat. If not and workerMax in DB is higher than actual workers, we'll schedule those workers on leader election.

### Context

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
@@ -1175,6 +1175,8 @@ public class JobActor extends AbstractActorWithTimers implements IMantisJobManag
     static class WorkerNumberGenerator {
 
         private static final Logger LOGGER = LoggerFactory.getLogger(WorkerNumberGenerator.class);
+        private static final int MAX_ATTEMPTS = 10;
+        private static final long SLEEP_DURATION_MS = Duration.ofSeconds(2).toMillis();
         private static final int DEFAULT_INCREMENT_STEP = 10;
         private final int incrementStep;
         private int lastUsed;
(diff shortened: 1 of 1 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
```diff
@@ -1202,21 +1205,28 @@ static class WorkerNumberGenerator {
...
 
         private void advance(MantisJobMetadataImpl mantisJobMetaData, MantisJobStore jobStore) {
+            currLimit += incrementStep;
             try {
-                currLimit += incrementStep;
-
                 mantisJobMetaData.setNextWorkerNumberToUse(currLimit, jobStore);
+                pendingFlushCnt.set(0);
             } catch (Exception e) {
+                if (pendingFlushCnt.incrementAndGet() > MAX_PENDING_FLUSH_CNT) {
```
Andyz26, line 1213: the write failure will still throw the runtime exception?

hmitnflx: oops, missed an early return here during a refactoring I did. Thanks!

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
```diff
@@ -1203,28 +1210,47 @@ static class WorkerNumberGenerator {
...
+                LOGGER.error("Exception setting next Worker number to use, " +
+                    "{} consecutive failed attempts", MAX_PENDING_FLUSH_CNT, e);
                 throw new RuntimeException("Unexpected error setting next worker number to use", e);
             }
         }
 
+        private void setNextWorkerNumberWithRetries(MantisJobMetadataImpl mantisJobMetaData, MantisJobStore jobStore, int value) throws Exception {
+            int attempts = 0;
+            Exception exception = null;
+            while (attempts < MAX_ATTEMPTS) {
```
sundargates, line 1228: can we retry infinitely?

sundargates (approved):
The change looks good! Ship it!

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
hmitnflx: I don't prefer an infinite retry for now, we can customize it in the future. For now, I increased it from 3 to 10

PR #629, opened 2024-02-12
[PR/623 - pt 1] Introduce scheduling constraints

fdc-ntflx:
### Context

This PR is the initial effort in breaking down the [larger PR](https://github.com/Netflix/mantis/pull/623) into more digestible, manageable pieces.

The first concept we're introducing involves scheduling constraints, which is a combined set of machine specifications and assignment attributes, such as differentiating between jdk17 and jdk8. The scheduling constraints are now integrated into task executor allocation requests, however, aren't used yet, keeping this PR as simple and streamlined as possible.

The scheduling constraints are created by the job cluster actor and then inserted into schedule requests. These constraints, intended for each worker that requires scheduling, are compiled using the requested machine definition from the worker and extracting assignment attributes. The extraction of these attributes is achieved through merging the tags of job artifacts, that are now stored during the build process, and job definition attributes that are set either during job cluster creation or job submission.

As mentioned earlier, the main scheduling algorithm continues functioning at the core-level and does not consider assignment attributes at this moment to restrict the amount of alterations introduced in each PR. This core scheduling logic will incorporate these attributes in a future PR.

### Checklist

- [ ] `./gradlew build` compiles code correctly
- [ ] Added new tests where applicable
- [ ] `./gradlew test` passes all tests
- [ ] Extended README or added javadocs where applicable

PR #632, opened 2024-02-13
[PR/623 - pt 2] Use scheduling constraints to find best fit TEs

fdc-ntflx:
### Context
This PR is the second one of an effort to split a [larger PR](https://github.com/Netflix/mantis/pull/623) into manageable parts.

In the [preceding PR](https://github.com/Netflix/mantis/pull/629), we introduced "scheduling constraints," composed of machine and assignment attributes, to guide scheduling decisions.

The focus of this PR is to employ these constraints optimally to identify the best Task Executors.

Task Executors will continue to register in the system, carrying essential metadata for facilitating on-demand scheduling. However, they now also carry "allocation attributes" to support the newly introduced scheduling logic. While initially the registered TEs were indexed solely by core count, we now allow multiple "versions" of TEs for each core count. For instance:

- 2 cores, 16GB | jdk:8
- 2 cores, 32GB | jdk:8
- 2 cores, 16GB | jdk:17

With this system, there might be multiple SKUs available for a given core count. Efficient and safe scheduling requires us to search for a single SKU that fits a request the best. To achieve this, we will change our indexing strategy for Task Executors: instead of purely using cores count, we will use a group key that encapsulates both machine definition and allocation attributes.

The scheduling search process now involves two steps. First, we need to ensure that the allocation attributes specified in the request can be satisfied by the TEs. Second, the selected TEs should be the "best fit" for the request in terms of CPU and memory. The first step is achieved by passing the required allocation attributes (and potential default values) along with the request (via config) to the RC actor. The second step is accomplished through a new interface introduced to calculate the fitness score of two machine definitions -- the one from the request and the one from the registered TE. An implementation of this interface is provided, which computes the fitness score based on a CPU/memory ratio of 18/1.
### Important note
To maintain focus on scheduling, this PR intentionally neglects the pending scheduling cache logic, which is currently flawed as it groups by cores, while it should use all scheduling constraints. This will be resolved in a future PR, the third in this series.

Unresolved items, marked with TODO in the PR, have been left deliberately to reduce complexity. An example would be the passage of the fitness function through the config; this and other items will be taken care of in subsequent PRs.

### Tests

* Added more unit tests
* verified mantis-helm works

### Checklist

### DIFF
--- a/mantis-common/src/main/java/io/mantisrx/common/WorkerConstants.java
+++ b/mantis-common/src/main/java/io/mantisrx/common/WorkerConstants.java
@@ -16,10 +16,13 @@
 
 package io.mantisrx.common;
 
+import java.util.regex.Pattern;
+
 public class WorkerConstants {
     public static final String WORKER_CONTAINER_DEFINITION_ID = "MANTIS_WORKER_CONTAINER_DEFINITION_ID";
     public static final String WORKER_TASK_ATTRIBUTE_ENV_KEY = "MANTIS_WORKER_CONTAINER_ATTRIBUTE";
     // TODO(fdichiara): make this configurable.
     public static final String AUTO_SCALE_GROUP_KEY = "NETFLIX_AUTO_SCALE_GROUP";
     public static final String MANTIS_WORKER_CONTAINER_GENERATION = "MANTIS_WORKER_CONTAINER_GENERATION";
+    public static final Pattern SCHEDULING_CONSTRAINT_PATTERN = Pattern.compile("^SCHEDULING_CONSTRAINT_(.*)$");
 }
--- a/mantis-common/src/main/java/io/mantisrx/common/WorkerConstants.java
+++ b/mantis-common/src/main/java/io/mantisrx/common/WorkerConstants.java
@@ -24,5 +24,5 @@ public class WorkerConstants {
     // TODO(fdichiara): make this configurable.
     public static final String AUTO_SCALE_GROUP_KEY = "NETFLIX_AUTO_SCALE_GROUP";
     public static final String MANTIS_WORKER_CONTAINER_GENERATION = "MANTIS_WORKER_CONTAINER_GENERATION";
-    public static final Pattern SCHEDULING_CONSTRAINT_PATTERN = Pattern.compile("^SCHEDULING_CONSTRAINT_(.*)$");
+    public static final Pattern MANTIS_SCHEDULING_ATTRIBUTE_PATTERN = Pattern.compile("^MANTIS_SCHEDULING_ATTRIBUTE_(.*)$");
 }
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
@@ -22,8 +22,12 @@ import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonCreator;
 import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonIgnore;
 import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonProperty;
 import io.mantisrx.shaded.com.google.common.collect.ImmutableMap;
+import java.util.AbstractMap;
 import java.util.Map;
 import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import lombok.AccessLevel;
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
@@ -140,5 +144,18 @@ public class TaskExecutorRegistration {
         }
 
         return Optional.empty();
+    }
+
+    @JsonIgnore
+    public Map<String, String> getAllocationAttributes() {
+        return taskExecutorAttributes.entrySet().stream()
+            .flatMap(entry -> {
+                Matcher matcher = WorkerConstants.SCHEDULING_CONSTRAINT_PATTERN.matcher(entry.getKey());
+                return matcher.matches() ? Stream.of(new AbstractMap.SimpleEntry<>(matcher.group(1).toLowerCase(), entry.getValue())) : Stream.empty();
+            })
+            .collect(Collectors.toMap(
+                Map.Entry::getKey,
+                entry -> getAttributeByKey(entry.getKey()).orElse(entry.getValue())
+            ));
     }
 }
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
@@ -29,11 +29,13 @@ import java.util.regex.Matcher;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.NonNull;
 import lombok.ToString;
+import lombok.Value;
 import lombok.experimental.FieldDefaults;
 
 /**
@@ -147,15 +149,27 @@ public class TaskExecutorRegistration {
     }
 
     @JsonIgnore
-    public Map<String, String> getAllocationAttributes() {
+    public Map<String, String> getSchedulingAttributes() {
         return taskExecutorAttributes.entrySet().stream()
             .flatMap(entry -> {
-                Matcher matcher = WorkerConstants.SCHEDULING_CONSTRAINT_PATTERN.matcher(entry.getKey());
+                Matcher matcher = WorkerConstants.MANTIS_SCHEDULING_ATTRIBUTE_PATTERN.matcher(entry.getKey());
                 return matcher.matches() ? Stream.of(new AbstractMap.SimpleEntry<>(matcher.group(1).toLowerCase(), entry.getValue())) : Stream.empty();
             })
             .collect(Collectors.toMap(
                 Map.Entry::getKey,
                 entry -> getAttributeByKey(entry.getKey()).orElse(entry.getValue())
             ));
+    }
+
+    @JsonIgnore
+    public TaskExecutorGroupKey getTaskExecutorGroupKey() {
+        return new TaskExecutorGroupKey(machineDefinition, getSchedulingAttributes());
+    }
+
+    @Value
+    @AllArgsConstructor
+    public static class TaskExecutorGroupKey {
+        MachineDefinition machineDefinition;
+        Map<String, String> schedulingAttributes;
     }
 }
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
@@ -35,11 +35,11 @@ import io.mantisrx.server.master.resourcecluster.ResourceCluster.ResourceOvervie
 import io.mantisrx.server.master.resourcecluster.TaskExecutorAllocationRequest;
 import io.mantisrx.server.master.resourcecluster.TaskExecutorID;
 import io.mantisrx.server.master.resourcecluster.TaskExecutorRegistration;
-import io.mantisrx.shaded.com.google.common.base.Splitter;
+import io.mantisrx.server.master.resourcecluster.TaskExecutorRegistration.TaskExecutorGroupKey;
 import io.mantisrx.shaded.com.google.common.cache.Cache;
 import io.mantisrx.shaded.com.google.common.cache.CacheBuilder;
 import io.mantisrx.shaded.com.google.common.cache.RemovalListener;
-import io.mantisrx.shaded.com.google.common.collect.ImmutableMap;
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashMap;
@@ -50,9 +50,7 @@ import java.util.NavigableSet;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.SortedMap;
 import java.util.TreeSet;
-import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -85,15 +83,14 @@ class ExecutorStateManagerImpl implements ExecutorStateManager {
     }
 
     /**
-     * Cache the available executors ready to accept assignments. Note these executors' state are not strongly
-     * synchronized and requires state level check when matching.
+     * Cache the available executors ready to accept assignments.
      */
-    private final SortedMap<Double, NavigableSet<TaskExecutorHolder>> executorByCores = new ConcurrentSkipListMap<>();
+    private final Map<TaskExecutorGroupKey, NavigableSet<TaskExecutorHolder>> executorsByGroup = new HashMap<>();
 
     // TODO(fdichiara): make this configurable
     private final FitnessCalculator fitnessCalculator = new CpuWeightedFitnessCalculator();
 
-    private final Map<String, String> assignmentAttributesAndDefaults;
+    private final Map<String, String> schedulingAttributes;
 
     private final Cache<TaskExecutorID, TaskExecutorState> archivedState = CacheBuilder.newBuilder()
         .maximumSize(10000)
@@ -102,8 +99,8 @@ class ExecutorStateManagerImpl implements ExecutorStateManager {
             log.info("Archived TaskExecutor: {} removed due to: {}", notification.getKey(), notification.getCause()))
         .build();
 
-    ExecutorStateManagerImpl(String assignmentAttributesAndDefaults) {
-        this.assignmentAttributesAndDefaults = assignmentAttributesAndDefaults.isEmpty() ? ImmutableMap.of() : Splitter.on(",").withKeyValueSeparator(':').split(assignmentAttributesAndDefaults);
+    ExecutorStateManagerImpl(Map<String, String> schedulingAttributes) {
+        this.schedulingAttributes = schedulingAttributes;
     }
 
     @Override
@@ -128,16 +125,16 @@ class ExecutorStateManagerImpl implements ExecutorStateManager {
         if (state.isAvailable() && state.getRegistration() != null) {
             TaskExecutorHolder teHolder = TaskExecutorHolder.of(taskExecutorID, state.getRegistration());
             log.debug("Marking executor {} as available for matching.", teHolder);
-            double cpuCores = state.getRegistration().getMachineDefinition().getCpuCores();
-            if (!this.executorByCores.containsKey(cpuCores)) {
-                log.info("[executorByCores] adding {} from TE: {}", cpuCores, teHolder);
-                this.executorByCores.putIfAbsent(
-                    cpuCores,
+            TaskExecutorGroupKey taskExecutorGroupKey = state.getRegistration().getTaskExecutorGroupKey();
+            if (!this.executorsByGroup.containsKey(taskExecutorGroupKey)) {
+                log.info("[executorsByGroup] adding {} from TE: {}", taskExecutorGroupKey, teHolder);
+                this.executorsByGroup.putIfAbsent(
+                    taskExecutorGroupKey,
                     new TreeSet<>(TaskExecutorHolder.generationFirstComparator));
             }
 
             log.info("Assign {} to available.", teHolder.getId());
-            return this.executorByCores.get(cpuCores).add(teHolder);
+            return this.executorsByGroup.get(taskExecutorGroupKey).add(teHolder);
         }
         else {
             log.debug("Ignore unavailable TE: {}", taskExecutorID);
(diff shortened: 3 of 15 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -318,29 +332,71 @@ private Optional<Map<TaskExecutorID, TaskExecutorState>> findBestFitFor(TaskExec
...
+     *
+     * @param schedulingConstraints The requested scheduling constraints to evaluate each TaskExecutor.
+     * @param targetCoreCount The target core count to select TaskExecutors from.
+     * @param currentBestFit Previously selected best fit TaskExecutors to exclude from consideration.
+     * @param numWorkers The desired number of workers in the final selection.
+     *
+     * @return A map of available TaskExecutorIDs to their TaskExecutorState that best fit the requested scheduling constraints.
+     * The number of entries is limited to `numWorkers`. If no TaskExecutor satisfies the scheduling constraints, this would be an empty map.
+     */
+    private Map<TaskExecutorID, TaskExecutorState> findAvailableBestFitTaskExecutors(SchedulingConstraints schedulingConstraints, Double targetCoreCount, BestFit currentBestFit, Integer numWorkers) {
```
Andyz26, line 351: this seems to violate the assumption that all TEs assigned to a stage will always be the same SKU variant.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -318,29 +332,71 @@ private Optional<Map<TaskExecutorID, TaskExecutorState>> findBestFitFor(TaskExec
...
+        double bestFitnessScore = 0;
+        List<TaskExecutorHolder> bestFittingTEs = new ArrayList<>();
+
+        for (TaskExecutorHolder teHolder: this.executorByCores.get(targetCoreCount).descendingSet()) {
```
Andyz26, line 355: it seems rather in-efficient to track all TEs under same core count then need to re-categorize based on attributes. I feel like the previous approach of tracking via a SKU variant ID is more efficient/safer.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
fdc-ntflx: I've updated the logic to index by a group key so we shouldn't have this problem.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
fdc-ntflx: fair point. I've updated the logic to index by a group key so we shouldn't have this problem.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -284,63 +297,75 @@ public Optional<BestFit> findBestFit(TaskExecutorBatchAssignmentRequest request)
...
-        if (Math.abs(targetCoreCount - requestedCoreCount) > 1E-10) {
-            // this mismatch should not happen in production and indicates TE registration/spec problem.
-            log.warn("Requested core count mismatched. requested: {}, found: {} for {}", requestedCoreCount,
-                targetCoreCount,
-                request);
-        }
 
-        if (this.executorByCores.get(targetCoreCount).isEmpty()) {
-            log.warn("No available TE found for core count: {}, request: {}", targetCoreCount, request);
+        log.debug("Applying assignment request: {} to constraints {}.", request, bestFitTeGroupKey);
```
Andyz26, line 318: nit: this can be info i think. Should be only there with every submission.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -284,63 +297,75 @@ public Optional<BestFit> findBestFit(TaskExecutorBatchAssignmentRequest request)
...
+        if (this.executorByGroupKey.getOrDefault(bestFitTeGroupKey.get(), Collections.emptyNavigableSet()).isEmpty()) {
```
Andyz26, line 319: nit: contains() is enough?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/scheduler/CpuWeightedFitnessCalculator.java
```diff
@@ -0,0 +1,38 @@
...
+
+import io.mantisrx.runtime.MachineDefinition;
+
+/**
+ * Implementation of FitnessCalculator. Uses cpu cores with higher weight than memory.
+ */
+public class CpuWeightedFitnessCalculator implements FitnessCalculator {
+    @Override
+    public double calculate(MachineDefinition requested, MachineDefinition available) {
+        if (!available.canFit(requested)) {
```
Andyz26, line 27: GIven we don't want to handle anything for disk/bandwidth/ports maybe let's remove the old check? (also I am thinking the case where we might want to adjust network bw more dynamically).

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -462,6 +488,31 @@ private Optional<Map<TaskExecutorID, TaskExecutorState>> findTaskExecutorsFor(Ta
...
+     * @return An Optional wrapping the Task Executor Group Key with the best fit constraints. If no fitting constraints
+     * are found, it returns an empty Optional.
+     */
+    private Optional<TaskExecutorGroupKey> findBestFitTeGroupKey(SchedulingConstraints requestedConstraints) {
+        return executorByGroupKey.keySet()
+            .stream()
+            // filter in the keys that have satisfied allocation constraints
+            .filter(taskExecutorGroupKey -> areAllocationConstraintsSatisfied(requestedConstraints, taskExecutorGroupKey.getAllocationAttributes()))
+            // filter in the keys with non-zero fitness score
+            .filter(taskExecutorGroupKey -> fitnessCalculator.calculate(requestedConstraints.getMachineDefinition(), taskExecutorGroupKey.getMachineDefinition()) > 0)
```
Andyz26, line 509: nit: .map().max() to avoid calculate twice

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/scheduler/CpuWeightedFitnessCalculator.java
fdc-ntflx: "we don't want to handle anything for disk/bandwidth/ports" I don't think that's totally accurate, or at least this wasn't my intent. My intent was to still verify requested disk/bw/port fit in available containers and that's the reason I kept the `canFit` call. I'm not sure we want to change that.

mantis-common/src/main/java/io/mantisrx/common/WorkerConstants.java
```diff
@@ -16,10 +16,13 @@
...
 
+import java.util.regex.Pattern;
+
 public class WorkerConstants {
     public static final String WORKER_CONTAINER_DEFINITION_ID = "MANTIS_WORKER_CONTAINER_DEFINITION_ID";
     public static final String WORKER_TASK_ATTRIBUTE_ENV_KEY = "MANTIS_WORKER_CONTAINER_ATTRIBUTE";
     // TODO(fdichiara): make this configurable.
     public static final String AUTO_SCALE_GROUP_KEY = "NETFLIX_AUTO_SCALE_GROUP";
     public static final String MANTIS_WORKER_CONTAINER_GENERATION = "MANTIS_WORKER_CONTAINER_GENERATION";
+    public static final Pattern SCHEDULING_CONSTRAINT_PATTERN = Pattern.compile("^SCHEDULING_CONSTRAINT_(.*)$");
```
Andyz26, line 27: mantis prefix?

Andyz26 (approved):

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
```diff
@@ -141,4 +145,17 @@ public Optional<String> getAttributeByKey(String attributeKey) {
 
         return Optional.empty();
     }
+
+    @JsonIgnore
+    public Map<String, String> getAllocationAttributes() {
```
sundargates, line 150: nit: getSchedulingAttributes

mantis-common/src/main/java/io/mantisrx/common/WorkerConstants.java
sundargates: nit: SCHEDULING_ATTRIBUTE

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -116,8 +116,8 @@ class ResourceClusterActor extends AbstractActorWithTimers {
 
     private final boolean isJobArtifactCachingEnabled;
 
-    static Props props(final ClusterID clusterID, final Duration heartbeatTimeout, Duration assignmentTimeout, Duration disabledTaskExecutorsCheckInterval, Clock clock, RpcService rpcService, MantisJobStore mantisJobStore, JobMessageRouter jobMessageRouter, int maxJobArtifactsToCache, String jobClustersWithArtifactCachingEnabled, boolean isJobArtifactCachingEnabled) {
-        return Props.create(ResourceClusterActor.class, clusterID, heartbeatTimeout, assignmentTimeout, disabledTaskExecutorsCheckInterval, clock, rpcService, mantisJobStore, jobMessageRouter, maxJobArtifactsToCache, jobClustersWithArtifactCachingEnabled, isJobArtifactCachingEnabled)
+    static Props props(final ClusterID clusterID, final Duration heartbeatTimeout, Duration assignmentTimeout, Duration disabledTaskExecutorsCheckInterval, Clock clock, RpcService rpcService, MantisJobStore mantisJobStore, JobMessageRouter jobMessageRouter, int maxJobArtifactsToCache, String jobClustersWithArtifactCachingEnabled, boolean isJobArtifactCachingEnabled, String assignmentAttributesAndDefaults) {
```
sundargates, line 119: nit: This class requires too many parameters. We should create a ResourceClusterActorSettings class to wrap all the necessary configurations.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -148,7 +149,7 @@ static Props props(final ClusterID clusterID, final Duration heartbeatTimeout, D
         this.maxJobArtifactsToCache = maxJobArtifactsToCache;
         this.jobClustersWithArtifactCachingEnabled = jobClustersWithArtifactCachingEnabled;
 
-        this.executorStateManager = new ExecutorStateManagerImpl();
+        this.executorStateManager = new ExecutorStateManagerImpl(assignmentAttributesAndDefaults);
```
sundargates, line 152: nit: schedulingAttributes.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -114,16 +127,16 @@ private boolean tryMarkAvailable(TaskExecutorID taskExecutorID, TaskExecutorStat
         if (state.isAvailable() && state.getRegistration() != null) {
             TaskExecutorHolder teHolder = TaskExecutorHolder.of(taskExecutorID, state.getRegistration());
             log.debug("Marking executor {} as available for matching.", teHolder);
-            double cpuCores = state.getRegistration().getMachineDefinition().getCpuCores();
-            if (!this.executorByCores.containsKey(cpuCores)) {
-                log.info("[executorByCores] adding {} from TE: {}", cpuCores, teHolder);
-                this.executorByCores.putIfAbsent(
-                    cpuCores,
+            TaskExecutorGroupKey taskExecutorGroupKey = TaskExecutorGroupKey.from(state.getRegistration());
```
sundargates, line 130: Shouldn't this be a property of TaskExecutorRegistration if it's purely a function of it?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -80,10 +85,14 @@ public int getTotalWorkers() {
...
-     * synchronized and requires state level check when matching.
+     * Cache the available executors ready to accept assignments.
      */
-    private final SortedMap<Double, NavigableSet<TaskExecutorHolder>> executorByCores = new ConcurrentSkipListMap<>();
+    private final Map<TaskExecutorGroupKey, NavigableSet<TaskExecutorHolder>> executorByGroupKey = new HashMap<>();
+
+    // TODO(fdichiara): make this configurable
+    private final FitnessCalculator fitnessCalculator = new CpuWeightedFitnessCalculator();
+
+    private final Map<String, String> assignmentAttributesAndDefaults;
```
sundargates, line 95: I feel like we should be consistent with the naming. It feels like we have three different names for the same concept: scheduling.*, assignment.*, allocation.*.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -92,6 +101,10 @@ public int getTotalWorkers() {
             log.info("Archived TaskExecutor: {} removed due to: {}", notification.getKey(), notification.getCause()))
         .build();
 
+    ExecutorStateManagerImpl(String assignmentAttributesAndDefaults) {
+        this.assignmentAttributesAndDefaults = assignmentAttributesAndDefaults.isEmpty() ? ImmutableMap.of() : Splitter.on(",").withKeyValueSeparator(':').split(assignmentAttributesAndDefaults);
```
sundargates, line 105: Can we pass a Map here instead? what's the need for using primitive types here?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
fdc-ntflx: lol. you're right. let me make things a bit more consistent.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
fdc-ntflx: agreed.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
fdc-ntflx: Can do at the end.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/config/MasterConfiguration.java
```diff
@@ -351,6 +351,11 @@ default Duration getSchedulerIntervalBetweenRetries() {
     @Default("false")
     boolean isBatchSchedulingEnabled();
 
+    // Example: "jdk:17"
+    @Config("mantis.scheduler.schedulingConstraints")
+    @Default("")
+    String getSchedulingConstraints();
+
```
sundargates, line 358: Is it possible to perform the transformation at this level instead of the leaf level (i.e., the actual usage)? This will allow us to set the expectation that this should be the return type of the config value in case we switch from the config-magic library to a different one in the future. For example, take a look at the getHeartbeatInterval example below.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/scheduler/CpuWeightedFitnessCalculator.java
```diff
@@ -0,0 +1,38 @@
...
+            return 0.0;
+        }
+
+        double cpuScore = 1 - (available.getCpuCores() - requested.getCpuCores()) / available.getCpuCores();
+        double memoryScore = 1 - (available.getMemoryMB() - requested.getMemoryMB()) / available.getMemoryMB();
+
+        // The weight for cpuScore is 18 and for memoryScore is 1. Hence, the total weight is 19.
+        // TODO: make these weights configurable.
+        return ((18 * cpuScore) + memoryScore) / 19;
```
sundargates, line 36: wow, these numbers will seem like magic without context!

mantis-control-plane/mantis-control-plane-server/src/test/java/io/mantisrx/master/resourcecluster/ResourceClusterActorTest.java
```diff
@@ -200,7 +200,7 @@ public void setupActor() {
                 jobMessageRouter,
                 0,
                 "",
-                false);
+                false, "");
```
sundargates, line 203: nit: next line.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
```diff
@@ -141,4 +147,29 @@ public Optional<String> getAttributeByKey(String attributeKey) {
...
+            ));
+    }
+
+    @JsonIgnore
+    public TaskExecutorGroupKey getTaskExecutorGroupKey() {
+        return TaskExecutorGroupKey.of(machineDefinition, getSchedulingAttributes());
+    }
+
+    @Value
+    @AllArgsConstructor(staticName = "of")
```
sundargates, line 170: nit: just curious, what's the need for this static constructor here?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -462,6 +488,31 @@ private Optional<Map<TaskExecutorID, TaskExecutorState>> findTaskExecutorsFor(Ta
...
+     *
+     * It filters the keys such that they satisfy the `requestedConstraints` and have a positive fitness score,
+     * then decides the best fit based on the highest fitness score.
+     *
+     * @param requestedConstraints Constraints of the scheduling request, serving as the reference for determining the best fit.
+     *
+     * @return An Optional wrapping the Task Executor Group Key with the best fit constraints. If no fitting constraints
+     * are found, it returns an empty Optional.
+     */
+    private Optional<TaskExecutorGroupKey> findBestFitTeGroupKey(SchedulingConstraints requestedConstraints) {
```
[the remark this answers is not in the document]
sundargates: findBestGroup maybe.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -343,6 +342,30 @@ private Optional<Map<TaskExecutorID, TaskExecutorState>> findBestFitFor(TaskExec
...
+     *
+     * Hence, the function ensures that the TaskExecutor scheduling attributes match or satisfy the constraints required. If either the TaskExecutor registration or the scheduling request lacks an attribute, it uses the provided defaults.
+     *
+     * @param constraints The schedule request constraints to be satisfied.
+     * @param teAssignmentAttributes The scheduling attributes of a Task Executor that needs to satisfy scheduling constraints.
+     *
+     * @return true if all allocation constraints are satisfied, false otherwise.
+     */
+    public boolean areSchedulingAttributeConstraintsSatisfied(SchedulingConstraints constraints, Map<String, String> teAssignmentAttributes) {
+        return schedulingAttributes.entrySet()
```
sundargates, line 361: Why do we need to define this as part of the control plane? The job definition already provides the scheduling constraints, including the hard constraints. All we need to do is match the TE attributes against these hard constraints. So, why do we need to duplicate this information in the control plane?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -284,57 +295,45 @@ public Optional<BestFit> findBestFit(TaskExecutorBatchAssignmentRequest request)
...
-
                     if (currentBestFit.contains(teHolder.getId())) {
                         return false;
                     }
-
                     TaskExecutorState st = this.taskExecutorStateMap.get(teHolder.getId());
                     return st.isAvailable() &&
                         st.getRegistration() != null &&
-                        st.getRegistration().getMachineDefinition().canFit(machineDefinition);
+                        st.getRegistration().getMachineDefinition().canFit(schedulingConstraints.getMachineDefinition());
```
sundargates, line 336: Do we need this check since the find best group already checks for the fitness? Can we just return the first TE in a group that's not assigned?

sundargates (approved):
LGTM! Thank you for this change!

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
fdc-ntflx: that's fair, I think this check is now redundant.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/scheduler/CpuWeightedFitnessCalculator.java
fdc-ntflx: will have more context.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
fdc-ntflx: we don't anymore

fdc-ntflx:
closing in favor of https://github.com/Netflix/mantis/pull/638

PR #635, opened 2024-02-15
[PR/623 - pt 4] Introduce sku size spec

fdc-ntflx:
### Context

This PR is the fourth one of an effort to split a [larger PR](https://github.com/Netflix/mantis/pull/623) into manageable parts.

This PR introduces a new class `SkuSizeSpec` and updates the existing class `SkuTypeSpec` to include a size attribute.

To describe varied Task Executors of the same size, the proposal suggests creating multiple SKUs and storing variables like runtime JDK in existing SKU metadata field. This approach leads to an increasing list of SKU-IDs to be shown to users during worker size selection. To keep the selection user-friendly, we will instead enrich SKUs with a size attribute which includes a 't-shirt size' label.

Key changes include:

* Introducing SkuSizeSpec Class: This new class includes the 't-shirt size' label/name that is to be displayed to the user, along with the machine resources definition.

* Updating SkuTypeSpec Class: The existing SkuTypeSpec class has been updated to include a size attribute of the new SkuSizeSpec class.

* Marking Existing Fields as Deprecated: The existing machine definition fields, cpuCoreCount, memorySizeInMB, networkMbps, and diskSizeInMB, have been marked as deprecated. These fields will be removed in a future PR after the upgrade of the existing skus.

These changes provide the basis for moving forward with the proposed scheduling constraints updates. This PR only introduces the new class and updates the SkuTypeSpec to include the size, but does not change any scheduling logic. 

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/SkuSizeSpec.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/SkuSizeSpec.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2022 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.master.resourcecluster.proto;
+
+import io.mantisrx.runtime.MachineDefinition;
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonCreator;
+import io.mantisrx.shaded.com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Builder;
+import lombok.Value;
+import org.apache.commons.lang3.Validate;
+
+/**
+ * Encapsulates the specifications of a SKU size in terms of the machine's resources.
+ * Each SKU size has a name (similar to the label of t-shirt sizes), and a dedicated machine definition that
+ * consists of specifications for CPU, disk, memory, and network.
+ *
+ * The name is a convenient way to refer to a particular configuration of machine resources while the machine definition represents the actual values.
+ */
+@Builder
+@Value
+public class SkuSizeSpec {
+    // Includes the specifications of a machine: CPU cores, disk in MB, memory in MB, and network bandwidth in Mbps
+    MachineDefinition machineDefinition;
+
+
+    // The name for the SKU size. It is a short, descriptive term typically representing the size of the machine resource being specified (e.g., "small", "medium", "large")
+    String name;
+
+    @JsonCreator
+    public SkuSizeSpec(
+        @JsonProperty("machineDefinition") final MachineDefinition machineDefinition,
+        @JsonProperty("name") final String name) {
+
+        validateMachineDefinition(machineDefinition);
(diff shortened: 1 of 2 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/SkuSizeSpec.java
```diff
@@ -0,0 +1,51 @@
...
+@Builder
+@Value
+@EqualsAndHashCode(onlyExplicitlyIncluded = true)
+public class SkuSizeSpec {
+    // Includes the specifications of a machine: CPU cores, disk in MB, memory in MB, and network bandwidth in Mbps
+    MachineDefinition machineDefinition;
+
+
+    // The label for the SKU size. It is a short, descriptive term typically representing the size of the machine resource being specified (e.g., "small", "medium", "large")
+    String label;
```
Andyz26, line 42: if this is specifically for sku description maybe make it more explicit e.g. sizeName.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/SkuSizeSpec.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Encapsulates the specifications of a SKU size in terms of the machine's resources.
+ * Each SKU size has a label (similar to the label of t-shirt sizes), and a dedicated machine definition that
+ * consists of specifications for CPU, disk, memory, and network.
+ *
+ * The label is a convenient way to refer to a particular configuration of machine resources while the machine definition represents the actual values.
+ */
+@Builder
+@Value
+@EqualsAndHashCode(onlyExplicitlyIncluded = true)
```
Andyz26, line 35: onlyExplicitlyIncluded = true but I don't see any fields marked?

Andyz26 (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClustersHostManagerActor.java
```diff
@@ -417,4 +417,29 @@ private static Optional<String> validateClusterSpec(ProvisionResourceClusterRequ
...
+                && sku.getNetworkMbps() >= 1;
+        }
+
+        // If size has been defined, validate its components
+        MachineDefinition machineDefinition = sku.getSize().getMachineDefinition();
+        return machineDefinition.getCpuCores() >= 1
+            && machineDefinition.getDiskMB() >= 1
+            && machineDefinition.getMemoryMB() >= 1
+            && machineDefinition.getNetworkMbps() >= 1;
+    }
```
sundargates, line 444: Can we do this validation when the SkuTypeSpec object is being created? It feels like we can avoid doing this check everywhere.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/SkuSizeSpec.java
```diff
@@ -0,0 +1,49 @@
...
+
+    // The name for the SKU size. It is a short, descriptive term typically representing the size of the machine resource being specified (e.g., "small", "medium", "large")
+    String name;
+
+    @JsonCreator
+    public SkuSizeSpec(
+        @JsonProperty("machineDefinition") final MachineDefinition machineDefinition,
+        @JsonProperty("name") final String name) {
+        this.name = name;
+        this.machineDefinition = machineDefinition;
```
sundargates, line 47: do the validation here.

sundargates (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/proto/SkuSizeSpec.java
fdc-ntflx: done

PR #636, opened 2024-02-16
[PR/623 - pt 5] Use sku size spec as scheduling constraint

fdc-ntflx:
### Context

This PR is the fourth one of an effort to split a [larger PR](https://github.com/Netflix/mantis/pull/623) into manageable parts.

This PR builds upon the current scheduling mechanism to integrate the newly incorporated size identifier from the sku spec. The primary objective is to leverage the t-shirt size label for augmented scheduling decisions, reverting to the explicit machine definition only when the t-shirt size label isn't available.

Under the updated scheduling approach, Task Executors can now be registered within the system, inclusive of the t-shirt size category they are associated with, thereby facilitating this new scheduling methodology.

Scheduling algo will now leverage these 3 functions:

* `findBestGroup`: This method appropriately uses a priority-based system to find the best group. It first attempts to find a match based on size names, and if that falls through, it tries to calculate the best fit using a fitness calculator.

* `findBestFitGroupSizeNameMatch`: This method correctly filters by checking if the size names for a group and the requested constraints are both present and the same. If these conditions are met and if the scheduling attribute constraints are satisfied, the group is selected. The use of findFirst ensures that the stream stops processing as soon as a match is found which is very efficient.

* `findBestGroupUsingFitnessCalculator`: It starts by filtering out the keys where size names exist and are different. This approach allows the fitness calculator to only be used on candidates that have already passed this check, which maximizes the efficiency of the algorithm. Finally, the use of the max function in conjunction with the fitness score ensures that the best fit group key is returned

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/scheduler/SchedulingConstraints.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/scheduler/SchedulingConstraints.java
@@ -20,11 +20,15 @@ import io.mantisrx.runtime.MachineDefinition;
 import io.mantisrx.shaded.com.google.common.annotations.VisibleForTesting;
 import io.mantisrx.shaded.com.google.common.collect.ImmutableMap;
 import java.util.Map;
+import java.util.Optional;
 import lombok.AllArgsConstructor;
 import lombok.Value;
 
 /**
- * A class that represents scheduling constraints. These constraints include the resource constraints, and a map of scheduling attributes (e.g. jdkVersion:17).
+ * A class that represents scheduling constraints.
+ * These constraints include the resource constraints for scheduling,
+ * the size name that could match the size label of a Task Executor,
+ * and a map of scheduling attributes (e.g., jdkVersion:17).
  */
 @AllArgsConstructor(staticName = "of")
 @Value
@@ -32,11 +36,20 @@ public class SchedulingConstraints {
     // Defines the resource constraints for scheduling
     MachineDefinition machineDefinition;
 
+    // Optional field to set predefined size name. When present, the scheduling system tries
+    // to match this field with the size name of a Task Executor.
+    Optional<String> sizeName;
+
     // Additional attributes for scheduling (ie. jdkVersion:17)
     Map<String, String> schedulingAttributes;
 
     @VisibleForTesting
     public static SchedulingConstraints of(MachineDefinition machineDefinition) {
-        return SchedulingConstraints.of(machineDefinition, ImmutableMap.of());
+        return SchedulingConstraints.of(machineDefinition, Optional.empty(), ImmutableMap.of());
+    }
+
+    @VisibleForTesting
+    public static SchedulingConstraints of(MachineDefinition machineDefinition, Map<String, String> schedulingAttributes) {
+        return SchedulingConstraints.of(machineDefinition, Optional.empty(), schedulingAttributes);
     }
 }
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/scheduler/SchedulingConstraints.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/scheduler/SchedulingConstraints.java
@@ -33,14 +33,21 @@ import lombok.Value;
 @AllArgsConstructor(staticName = "of")
 @Value
 public class SchedulingConstraints {
-    // Defines the resource constraints for scheduling
+    /**
+     * Defines the resource constraints for scheduling
+     */
     MachineDefinition machineDefinition;
 
-    // Optional field to set predefined size name. When present, the scheduling system tries
-    // to match this field with the size name of a Task Executor.
+    /**
+     * Optional field to set a predefined size name. When this field is present, the scheduling system prioritizes
+     * matching this field with the size name of a Task Executor Group during the worker allocation process. If no match is found,
+     * the function falls back to a fitness calculation on machine definition.
+     */
     Optional<String> sizeName;
 
-    // Additional attributes for scheduling (ie. jdkVersion:17)
+    /**
+     * Additional attributes for scheduling (ie. jdkVersion:17)
+     */
     Map<String, String> schedulingAttributes;
 
     @VisibleForTesting
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
@@ -15,6 +15,8 @@
  */
 package io.mantisrx.server.master.resourcecluster;
 
+import static io.mantisrx.common.WorkerConstants.MANTIS_CONTAINER_SIZE_NAME_KEY;
+
 import io.mantisrx.common.WorkerConstants;
 import io.mantisrx.common.WorkerPorts;
 import io.mantisrx.runtime.MachineDefinition;
@@ -163,13 +165,14 @@ public class TaskExecutorRegistration {
 
     @JsonIgnore
     public TaskExecutorGroupKey getTaskExecutorGroupKey() {
-        return new TaskExecutorGroupKey(machineDefinition, getSchedulingAttributes());
+        return new TaskExecutorGroupKey(machineDefinition, getAttributeByKey(MANTIS_CONTAINER_SIZE_NAME_KEY), getSchedulingAttributes());
     }
 
     @Value
     @AllArgsConstructor
     public static class TaskExecutorGroupKey {
         MachineDefinition machineDefinition;
+        Optional<String> sizeName;
         Map<String, String> schedulingAttributes;
     }
 }
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/StringConstants.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/StringConstants.java
@@ -19,4 +19,5 @@ package io.mantisrx.master;
 public class StringConstants {
     // User for actions performed by Mantis master
     public static final String MANTIS_MASTER_USER = "MantisMaster";
+    public static final String MANTIS_STAGE_CONTAINER_SIZE_NAME_KEY = "_mantis.stageContainerSizeName";
 }
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/StringConstants.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/StringConstants.java
@@ -19,5 +19,6 @@ package io.mantisrx.master;
 public class StringConstants {
     // User for actions performed by Mantis master
     public static final String MANTIS_MASTER_USER = "MantisMaster";
+    // Key to define the size name within a stage container's attributes map
     public static final String MANTIS_STAGE_CONTAINER_SIZE_NAME_KEY = "_mantis.stageContainerSizeName";
 }
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/IMantisJobMetadata.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/IMantisJobMetadata.java
@@ -194,6 +194,4 @@ public interface IMantisJobMetadata {
      * is past timeout secs
      */
     long getWorkerTimeoutSecs();
-
-    Optional<String> getStageAttribute(int stageNum, String sizeName);
 }
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/IMantisJobMetadata.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/IMantisJobMetadata.java
@@ -194,4 +194,6 @@ public interface IMantisJobMetadata {
      * is past timeout secs
      */
     long getWorkerTimeoutSecs();
+
+    Optional<String> getStageAttribute(int stageNum, String sizeName);
 }
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
@@ -1594,7 +1594,8 @@ public class JobActor extends AbstractActorWithTimers implements IMantisJobManag
                         // TODO(fdichiara): make this a property of JobStageMetadata. https://github.com/Netflix/mantis/pull/629/files#r1487043262
                         SchedulingConstraints.of(
                             stageMetadata.getMachineDefinition(),
-                            mantisJobMetaData.getStageAttribute(stageMetadata.getStageNum(), MANTIS_STAGE_CONTAINER_SIZE_NAME_KEY),
+                            // Fetch the 'sizeName' for the given stage among its container attributes
+                            stageMetadata.getSizeAttribute(),
                             mergeJobDefAndArtifactAssigmentAttributes(jobMetadata.getJobArtifact())),
                         hardConstraints,
                         softConstraints,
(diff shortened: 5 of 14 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/IMantisJobMetadata.java
```diff
@@ -194,4 +194,6 @@ public interface IMantisJobMetadata {
      * is past timeout secs
      */
     long getWorkerTimeoutSecs();
+
+    Optional<String> getStageAttribute(int stageNum, String sizeName);
```
Andyz26, line 198: not sure i understand what this function does with stageNum + sizeName. Can you add an example to doc?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/MantisJobMetadataImpl.java
```diff
@@ -503,4 +503,10 @@ public int hashCode() {
     public Costs getJobCosts() {
         return jobCosts;
     }
+
+    @Override
+    public Optional<String> getStageAttribute(int stageNum, String sizeName) {
```
Andyz26, line 508: do you mean sizeNameKey?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -511,19 +510,59 @@ private Optional<Map<TaskExecutorID, TaskExecutorState>> findTaskExecutorsFor(Ta
...
+     * serving as the reference for determining the best fit.
      *
-     * @return An Optional wrapping the Task Executor Group Key with the best fit constraints. If no fitting constraints
-     * are found, it returns an empty Optional.
+     * @return An Optional wrapping the Task Executor Group Key with the best fit constraints.
+     * If no fitting constraints are found, it returns an empty Optional.
      */
     private Optional<TaskExecutorGroupKey> findBestGroup(SchedulingConstraints requestedConstraints) {
-        return executorsByGroup.keySet()
+        // Partition the keys into two sets: those that match the sizeName and those that don't
```
Andyz26, line 524: suggestion: simplify this logic without storing an extra intermediate map of partition results. e.g.:
1. check for exact match on sizeName if present and return if found.
2. if not found scan with fitness score.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/scheduler/SchedulingConstraints.java
```diff
@@ -20,23 +20,36 @@
...
     // Additional attributes for scheduling (ie. jdkVersion:17)
     Map<String, String> schedulingAttributes;
 
     @VisibleForTesting
     public static SchedulingConstraints of(MachineDefinition machineDefinition) {
-        return SchedulingConstraints.of(machineDefinition, ImmutableMap.of());
+        return SchedulingConstraints.of(machineDefinition, Optional.empty(), ImmutableMap.of());
+    }
+
+    @VisibleForTesting
```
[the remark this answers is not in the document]
fdc-ntflx: because we are using this only for tests.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/IMantisJobMetadata.java
fdc-ntflx: my bad this function is supposed to return the value of a given key of stage container attributes (which may be null or the key may not exist). let me rename `sizeName` with something more appropriate and rename the method to reflect this. 

The idea here is to extract the size name from the stage info containerAttributes map.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/MantisJobMetadataImpl.java
fdc-ntflx: I will rename this as `getStageAttributeByKey`

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/scheduler/SchedulingConstraints.java
```diff
@@ -20,23 +20,36 @@
...
  */
 @AllArgsConstructor(staticName = "of")
 @Value
 public class SchedulingConstraints {
     // Defines the resource constraints for scheduling
     MachineDefinition machineDefinition;
 
+    // Optional field to set predefined size name. When present, the scheduling system tries
+    // to match this field with the size name of a Task Executor.
+    Optional<String> sizeName;
```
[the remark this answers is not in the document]
fdc-ntflx: Yeah the idea is to be able to support both scheduling with size name & machine definition but for now I'm assuming size name to be optional while machine definition mandatory. this might change in the future though.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
fdc-ntflx: I agree. I polished this function a bit. I hope it's more clear now.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/scheduler/SchedulingConstraints.java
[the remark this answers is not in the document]
fdc-ntflx: I improved the docs. let me know if that's still not enough.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -508,32 +506,78 @@ private Optional<Map<TaskExecutorID, TaskExecutorState>> findTaskExecutorsFor(Ta
...
+     * Finds the best fit Task Executor Group by using a fitness calculator on machine definitions.
+     *
+     * Groups that match requestedConstraints and have a fitness score greater than 0 are considered.
+     * Among these, the key with the highest score is returned.
+     *
+     * @param requestedConstraints The constraints for the scheduling request.
+     * @return An Optional of the best fit Task Executor Group Key according to a fitness calculator.
+     *         If no suitable key is found, it returns an empty Optional.
+     */
+    private Optional<TaskExecutorGroupKey> findBestGroupByFitnessCalculator(SchedulingConstraints requestedConstraints) {
```
Andyz26, line 558: add log?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -508,32 +506,78 @@ private Optional<Map<TaskExecutorID, TaskExecutorState>> findTaskExecutorsFor(Ta
...
         return executorsByGroup.keySet()
             .stream()
-            // filter in the keys that have satisfied allocation constraints
-            .filter(taskExecutorGroupKey -> areSchedulingAttributeConstraintsSatisfied(requestedConstraints, taskExecutorGroupKey.getSchedulingAttributes()))
-            // Map each TaskExecutorGroupKey to a Pair containing the key and its corresponding fitness score
+            // Filter out if both sizeName exist and are different (ie. small vs large)
+            .filter(taskExecutorGroupKey -> {
+                Optional<String> teGroupSizeName = taskExecutorGroupKey.getSizeName();
```
Andyz26, line 563: do we still need to care size names on the fallback case?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
Andyz26: or metric on fallback count.

mantis-control-plane/mantis-control-plane-server/src/test/java/io/mantisrx/master/resourcecluster/ResourceClusterActorClusterUsageAkkaTest.java
```diff
@@ -285,4 +285,6 @@ private static void assertIdleAndTotalCount(GetClusterUsageResponse usageRes, St
         assertEquals(idleCount, usage.getIdleCount());
         assertEquals(totalCount, usage.getTotalCount());
     }
+
+    // TODO: add test using sizeName....
```
Andyz26, line 289: is this still valid?

Andyz26 (approved):

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/scheduler/SchedulingConstraints.java
```diff
@@ -20,23 +20,43 @@
...
+     */
     MachineDefinition machineDefinition;
 
-    // Additional attributes for scheduling (ie. jdkVersion:17)
+    /**
+     * Optional field to set a predefined size name. When this field is present, the scheduling system prioritizes
+     * matching this field with the size name of a Task Executor Group during the worker allocation process. If no match is found,
+     * the function falls back to a fitness calculation on machine definition.
+     */
+    Optional<String> sizeName;
```
sundargates, line 46: why have this as an explicit property? why not pass this via attribute?

This is the semantics I am thinking: I have this resource requirements but with these attributes and the size can be part of this.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/TaskExecutorRegistration.java
```diff
@@ -163,13 +165,14 @@ public Map<String, String> getSchedulingAttributes() {
...
     public TaskExecutorGroupKey getTaskExecutorGroupKey() {
-        return new TaskExecutorGroupKey(machineDefinition, getSchedulingAttributes());
+        return new TaskExecutorGroupKey(machineDefinition, getAttributeByKey(MANTIS_CONTAINER_SIZE_NAME_KEY), getSchedulingAttributes());
     }
 
     @Value
     @AllArgsConstructor
     public static class TaskExecutorGroupKey {
         MachineDefinition machineDefinition;
+        Optional<String> sizeName;
```
sundargates, line 175: Same as above.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
```diff
@@ -1593,6 +1594,8 @@ private ScheduleRequest createSchedulingRequest(
                         // TODO(fdichiara): make this a property of JobStageMetadata. https://github.com/Netflix/mantis/pull/629/files#r1487043262
                         SchedulingConstraints.of(
                             stageMetadata.getMachineDefinition(),
+                            // Fetch the 'sizeName' for the given stage among its container attributes
+                            mantisJobMetaData.getStageAttributeByKey(stageMetadata.getStageNum(), MANTIS_STAGE_CONTAINER_SIZE_NAME_KEY),
                             mergeJobDefAndArtifactAssigmentAttributes(jobMetadata.getJobArtifact())),
```
sundargates, line 1599: stageMetadata.getSizeAttribute()

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/scheduler/SchedulingConstraints.java
fdc-ntflx: I did indeed consider this approach of passing the size as an attribute to preserve the existing structure of the scheduling constraints. Nonetheless, I had a few concerns that led me to opt for a different path.

Firstly, regarding readability: Retaining the size as an explicit property could potentially enhance the clarity and readability of the code. Integrating the size into the attributes might lead to confusion, as it's not immediately obvious that it's a "size" attribute among other "required" scheduling characteristics.

Secondly, I have concerns about "hard requirements" vs size: Scheduling attributes - like the JDK version - are sort of "hard requirements," whereas size is not. Mixing these diverse types of parameters might disrupt the original definition and purpose of the attributes.

Thirdly, in terms of handling key prefixes: To distinguish the size from other attributes, we may have to implement a prefix mechanism for the size. This wouldn't only add unnecessary complexity, but also divert from the simplicity and straightforwardness of keeping all attribute keys consistent.

Ultimately, it's about finding a convention that works, so if you have strong feelings about this, I'm open to making the changes you suggest.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/StringConstants.java
```diff
@@ -19,4 +19,5 @@
 public class StringConstants {
     // User for actions performed by Mantis master
     public static final String MANTIS_MASTER_USER = "MantisMaster";
+    public static final String MANTIS_STAGE_CONTAINER_SIZE_NAME_KEY = "_mantis.stageContainerSizeName";
```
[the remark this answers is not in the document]
fdc-ntflx: This line is a key used to define the size name within a stage container's attributes map. I'll add a comment in the code to make this clearer. Thanks for pointing it out!

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/IMantisJobMetadata.java
```diff
@@ -194,4 +194,16 @@ public interface IMantisJobMetadata {
...
+     * Retrieves the value of a specific container attribute (if it exists) for a given stage number.
+     *
+     * @param stageNum The stage number from which to retrieve the container attribute.
+     * @param attributeKey The key of the attribute to retrieve.
+     *
+     * @return An Optional that contains the attribute value if it exists. Returns an empty Optional if
+     *         container attributes is null for a given stage, or if the specific attribute key
+     *         is not found among container attributes.
+     */
+    Optional<String> getStageAttributeByKey(int stageNum, String attributeKey);
```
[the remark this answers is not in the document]
fdc-ntflx: agreed

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
fdc-ntflx: agreed

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
fdc-ntflx: My approach is as follows: if the scheduling request includes a size name, we should exclude any TEs that have a set size that differs from the request. We should only consider TEs that either have no size set at all or ones that also have the same size as specified in the scheduling request. This holds true even when we fallback to using the machine definition for matching. We want to maintain the integrity of the size label match regardless of the matching strategy used. Does it make sense?

mantis-control-plane/mantis-control-plane-server/src/test/java/io/mantisrx/master/resourcecluster/ResourceClusterActorClusterUsageAkkaTest.java
fdc-ntflx: yeah, I just added a bit more tests on the getUsage path when TEs register with size name.

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/scheduler/SchedulingConstraints.java
sundargates: Good point. I think eventually, it would be better to modify the JobDefinition to include hard and soft constraints instead of representing them as job tags.

sundargates (approved):
The change looks good to me.

fdc-ntflx:
Closing in favor of https://github.com/Netflix/mantis/pull/638

PR #640, opened 2024-02-27
For memory-based autoscaling, make memory and JVMMemory interchangeable.

sundargates:
### Context

Memory-based metrics are not useful for Java applications because the Java Virtual Machine (JVM) usually allocates memory aggressively and does not return it to the operating system. This means that even if the memory usage appears high, the actual size of live data could be very small. Therefore, this pull request intends to address this issue by making JVMMemory and Memory-based metrics interchangeable.

### Checklist


mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/WorkerMetricHandler.java
```diff
@@ -19,6 +19,7 @@
 import static io.mantisrx.server.core.stats.MetricStringConstants.*;
 
 import io.mantisrx.runtime.descriptor.StageScalingPolicy;
+import io.mantisrx.runtime.descriptor.StageScalingPolicy.ScalingReason;
```
calvin681, line 22: Remove this and keep the code consistent.

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/Util.java
```diff
@@ -32,7 +32,7 @@ public static double getEffectiveValue(StageSchedulingInfo stageSchedulingInfo,
         case Memory:
```
calvin681, line 32: Should we annotate ScalingReason.Memory as deprecated?

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/WorkerMetricHandler.java
sundargates: Sure. It looks like it was automatically imported by Intellij.

calvin681:
This change can have a big blast radius. I don't know how many existing jobs has Memory as scaling policy. This will change their scaling pattern. We should be careful when resubmitting those jobs.

sundargates:
> This change can have a big blast radius. I don't know how many existing jobs has Memory as scaling policy. This will change their scaling pattern. We should be careful when resubmitting those jobs.

I agree. The goal of this PR is precisely that though. All these jobs today are getting autoscaled to the maximum number of nodes because they think their memory utilization is high when it's not the case.

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/Util.java
sundargates: done

calvin681 (approved):

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/WorkerMetricHandler.java
```diff
@@ -311,15 +311,16 @@ public void call() {
...
-                                jobAutoScaleObserver.onNext(
-                                        new JobAutoScaler.Event(StageScalingPolicy.ScalingReason.Memory, stage,
-                                                allWorkerAggregates.get(RESOURCE_USAGE_METRIC_GROUP).getGauges().get(MetricStringConstants.TOT_MEM_USAGE_CURR), numWorkers, ""));
                                 jobAutoScaleObserver.onNext(
                                         new JobAutoScaler.Event(StageScalingPolicy.ScalingReason.Network, stage,
                                                 allWorkerAggregates.get(RESOURCE_USAGE_METRIC_GROUP).getGauges().get(MetricStringConstants.NW_BYTES_USAGE_CURR), numWorkers, ""));
                                 jobAutoScaleObserver.onNext(
-                                        new JobAutoScaler.Event(StageScalingPolicy.ScalingReason.JVMMemory, stage,
-                                                allWorkerAggregates.get(RESOURCE_USAGE_METRIC_GROUP).getGauges().get("jvmMemoryUsedBytes"), numWorkers, "")
+                                        new JobAutoScaler.Event(StageScalingPolicy.ScalingReason.Memory, stage,
```
fdc-ntflx, line 318: can you add a comment in here?

fdc-ntflx (approved):
lgtm! thanks for fixing the autoscaler <3

Andyz26 (approved):

PR #641, opened 2024-02-28
Allow disabled job to load if it contains invalid state

Andyz26:
### Context

Allow job cluster loading on disabled job cluster definitions if the cluster definition misses required parameters to maintain proper back compat.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobClusterDefinitionImpl.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobClusterDefinitionImpl.java
@@ -59,7 +59,8 @@ public class JobClusterDefinitionImpl implements IJobClusterDefinition {
                                     @JsonProperty("migrationConfig") WorkerMigrationConfig migrationConfig,
                                     @JsonProperty("isReadyForJobMaster") boolean isReadyForJobMaster,
                                     @JsonProperty("parameters") List<Parameter> parameters,
-                                    @JsonProperty("labels") List<Label> labels
+                                    @JsonProperty("labels") List<Label> labels,
+                                    @JsonProperty("isDisabled") boolean isDisabled
     ) {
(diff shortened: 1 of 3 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobClusterDefinitionImpl.java
```diff
@@ -75,14 +76,27 @@ public JobClusterDefinitionImpl(@JsonProperty("name") String name,
...
+            Preconditions.checkNotNull(labels, "labels cannot be empty.");
+            Preconditions.checkArgument(
+                labels.stream()
+                    .anyMatch(l ->
+                        l.getName().equalsIgnoreCase(SystemLabels.MANTIS_RESOURCE_CLUSTER_NAME_LABEL.label)),
+                "Missing required label: " + SystemLabels.MANTIS_RESOURCE_CLUSTER_NAME_LABEL.label);
+        }
     }
 
+    public JobClusterDefinitionImpl(String name,
```
sundargates, line 89: nit: @VisibleForTesting? Also, does this need to public or is package-private good enough?

sundargates: Actually it looks like this is only called from the builder. so, maybe private is good enough.

sundargates (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobClusterDefinitionImpl.java
Andyz26: it's also used in other path where the create request etc. doesn't have include data about isDisabled thus I gave it a default to maintain the current behavior.

fdc-ntflx (approved):
lgtm!

PR #643, opened 2024-03-07
Prioritize latest generation TEs in scheduling during SKU upgrades

fdc-ntflx:
### Context

During the process of adding size metadata to an existing SKU and initiating corresponding ASG updates, TEs from both new and existing ASGs are now grouped separately in the resource cluster actor, i.e., one with size and one without. Due to this, issues may arise during task migrations as it's not predictable which TEs will be chosen by the scheduler. While, instead, we want to always use TEs from the latest ASGs.

This PR enhances our scheduling logic to solve this problem. Specifically, the `findBestGroupBySizeNameMatch` and `findBestGroupByFitnessCalculator` functions will now prefer TE groups with "younger" generation (aka higher ASG name). This effectively ensures that the newer ASG (with a higher generation value), resulting from a SKU upgrade, will be preferably used over the older ASGs.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
@@ -81,7 +81,11 @@ class ExecutorStateManagerImpl implements ExecutorStateManager {
             this.groupToTaskExecutorCount = constraintsToTaskAllocationRequests
                 .entrySet()
                 .stream()
-                .collect(Collectors.toMap(entry -> findBestFitGroupOrDefault(entry.getKey()), entry -> entry.getValue().size()));
+                .collect(Collectors.toMap(
+                    entry -> findBestFitGroupOrDefault(entry.getKey()),
+                    entry -> entry.getValue().size(),
+                    Integer::sum
+                ));
         }
 
         public int getTotalWorkers() {
@@ -553,7 +557,15 @@ class ExecutorStateManagerImpl implements ExecutorStateManager {
             // Verify scheduling attribute constraints
             .filter(taskExecutorGroupKey -> areSchedulingAttributeConstraintsSatisfied(requestedConstraints,
                 taskExecutorGroupKey.getSchedulingAttributes()))
-            .findFirst();
+            // Get highest generation group
+            .max(Comparator.comparing(taskExecutorGroupKey -> {
+                NavigableSet<TaskExecutorHolder> holders = executorsByGroup.get(taskExecutorGroupKey);
+                if (holders.isEmpty()) {
+                    return null;
+                } else {
+                    return holders.last().getGeneration();
+                }
+            }, Comparator.nullsLast(Comparator.reverseOrder())));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -542,18 +546,32 @@ private Optional<TaskExecutorGroupKey> findBestGroup(SchedulingConstraints reque
...
+                // Filter to retain groups where the requested sizeName is also present
+                .filter(group -> requestedConstraints.getSizeName().isPresent())
+                // Filter to retain groups where sizeNames of group and requested constraints are equal
+                .filter(group -> group.getSizeName().get().equalsIgnoreCase(requestedConstraints.getSizeName().get()))
+                // Verify scheduling attribute constraints
+                .filter(taskExecutorGroupKey -> areSchedulingAttributeConstraintsSatisfied(requestedConstraints,
+                        taskExecutorGroupKey.getSchedulingAttributes()))
+                // Get highest generation group
+                .max(Comparator.comparing(taskExecutorGroupKey -> {
+                    NavigableSet<TaskExecutorHolder> holders = executorsByGroup.get(taskExecutorGroupKey);
```
Andyz26, line 568: is it possible to save these as sorted to avoid re-sort every time?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -542,18 +546,32 @@ private Optional<TaskExecutorGroupKey> findBestGroup(SchedulingConstraints reque
      *         If no suitable key is found, it returns an empty Optional.
      */
     private Optional<TaskExecutorGroupKey> findBestGroupBySizeNameMatch(SchedulingConstraints requestedConstraints) {
+        // Filter to retain groups where sizeName is present
+        // Filter to retain groups where the requested sizeName is also present
+        // Filter to retain groups where sizeNames of group and requested constraints are equal
+        // Verify scheduling attribute constraints
+        // Sorted descending by generation
+        // Get highest generation group
         return executorsByGroup.keySet()
```
Andyz26, line 555: I am a little lost here: are same SKU id under different generations sharing the same group key?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -589,9 +601,21 @@ private Optional<TaskExecutorGroupKey> findBestGroupByFitnessCalculator(Scheduli
...
             .filter(entry -> entry.getValue() > 0)
-            // Find the entry with the highest fitness score (if it exists)
-            .max(Entry.comparingByValue())
-            .map(AbstractMap.SimpleEntry::getKey);
+            // Get highest generation group
+            .sorted((entry1, entry2) -> {
+                int fitnessComparison = entry2.getValue().compareTo(entry1.getValue());
+                if (fitnessComparison != 0) {
+                    return fitnessComparison;
+                } else {
```
Andyz26, line 609: add comments.

Andyz26 (approved):

PR #648, opened 2024-03-29
Use injected usage supplier in RuntimeTaskImpl.

arnavdugar:
### Summary
Use injected usage supplier in `RuntimeTaskImpl`.

### Context
Currently `CgroupsMetricsCollector` is hard-coded in `RuntimeTaskImpl.java`, so if it is overridden using the `
mantis.taskexecutor.metrics.collector` field in the Mantis agent properties, that isn't respected. This change uses the dynamically loaded implementation of `MetricsCollector`.

### Checklist

### DIFF
--- a/mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/WorkerExecutionOperationsNetworkStage.java
+++ b/mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/WorkerExecutionOperationsNetworkStage.java
@@ -53,6 +53,7 @@ import io.mantisrx.server.core.StatusPayloads;
 import io.mantisrx.server.core.WorkerAssignments;
 import io.mantisrx.server.core.WorkerHost;
 import io.mantisrx.server.master.client.MantisMasterGateway;
+import io.mantisrx.server.worker.client.SseWorkerConnection;
 import io.mantisrx.server.worker.client.WorkerMetricsClient;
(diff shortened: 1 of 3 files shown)


mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/RuntimeTaskImpl.java
```diff
@@ -99,8 +99,9 @@ public void initialize(
             this.wrappedExecuteStageRequest =
                 new WrappedExecuteStageRequest(PublishSubject.create(), executeStageRequest);
 
-            log.info("Picking Cgroups metrics collector.");
-            configWritable.setMetricsCollector(CgroupsMetricsCollector.valueOf(System.getProperties()));
+            MetricsCollector metricsCollector = config.getUsageSupplier();
```
Andyz26, line 102: hi @arnavdugar we are trying to move away from passing objects via the config class and this config class you use here is generated from the sterilized version (L94 -96) so the supplier is always empty.
If you need to support another metrics collector I would suggest adding a config value as string and then construct the metrics collector type instance based the string value instead (basically treat the current interface at initialize as entry and do not pass external instances from there). We need this assumption to support our internal wrapper to these entries.

Andyz26 (changes requested):

arnavdugar:
I accidentally closed this PR, so I created https://github.com/Netflix/mantis/pull/674.

PR #654, opened 2024-04-10
Fix Resource cluster API timeout

Andyz26:
### Context

Fix slow APIs (update/upgrade) gateway timeout.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterScalerActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterScalerActor.java
@@ -279,7 +279,7 @@ public class ResourceClusterScalerActor extends AbstractActorWithTimers {
     }
 
     private void onQueueClusterRuleRefreshRequest(QueueClusterRuleRefreshRequest req) {
-        log.info("{}: Queue a request to refresh cluster rules", this.clusterId);
+        log.debug("{}: Queue a request to refresh cluster rules", this.clusterId);
         self().tell(new TriggerClusterRuleRefreshRequest(this.clusterId), self());
         getSender().tell(Ack.getInstance(), self());
     }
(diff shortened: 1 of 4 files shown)


sundargates (commented):
This does not fix the problem IMO. Imagine you lose the event because the node restarted after you have acked the incoming request. 

I think the right way to solve this would be increase the timeout in these cases. This way when you have acked the request, there's no case of potential loss of event.

Andyz26:
> This does not fix the problem IMO. Imagine you lose the event because the node restarted after you have acked the incoming request.
> 
> I think the right way to solve this would be increase the timeout in these cases. This way when you have acked the request, there's no case of potential loss of event.

Synced offline.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterScalerActor.java
```diff
@@ -277,6 +278,12 @@ private void onTriggerClusterRuleRefreshRequest(TriggerClusterRuleRefreshRequest
         this.fetchRuleSet();
     }
 
+    private void onQueueClusterRuleRefreshRequest(QueueClusterRuleRefreshRequest req) {
+        log.info("{}: Queue a request to refresh cluster rules", this.clusterId);
+        self().tell(new TriggerClusterRuleRefreshRequest(this.clusterId), self());
+        getSender().tell(Ack.getInstance(), self());
+    }
+
```
sundargates, line 286: What's the need for a separate message? How is this different from sending an ack back as the first thing in TriggerClusterRuleRefreshRequest message handler?

sundargates (approved):
LGTM, but we can avoid the extra message because it only adds to the complexity of these actors.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterScalerActor.java
Andyz26: will switch to debug.

PR #661, opened 2024-04-24
Fix leaked schedulingChanges connections

Andyz26:
### Context

We are seeing a high frequency of unexpected calls to the "assignmentResults" API to closed jobs in a prod instance, which turns out to be connections stuck in an infinite retry policy on the observable created by mantisClient.

The fix here added the error handling to both server API response (return 404 on closed jobs) and client-side (escape the infinite retry if getting 404 response).

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/MantisMasterClientApi.java
+++ b/mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/MantisMasterClientApi.java
@@ -699,16 +699,24 @@ public class MantisMasterClientApi implements MantisMasterGateway {
      * @return
      */
     public Observable<JobSchedulingInfo> schedulingChanges(final String jobId) {
+        final ConditionalRetry retryObject = new ConditionalRetry(null, "assignmentresults_" + jobId);
         return masterMonitor.getMasterObservable()
                 .filter(masterDescription -> masterDescription != null)
                 .retryWhen(retryLogic)
                 .switchMap((Func1<MasterDescription,
                         Observable<JobSchedulingInfo>>) masterDescription -> getRxnettySseClient(
                                 masterDescription.getHostname(), masterDescription.getSchedInfoPort())
-                        .submit(HttpClientRequest.createGet("/assignmentresults/" + jobId + "?sendHB=true"))
+                        .submit(
+                            HttpClientRequest.createGet("/assignmentresults/" + jobId + "?sendHB=true"))
                         .flatMap((Func1<HttpClientResponse<ServerSentEvent>,
                                 Observable<JobSchedulingInfo>>) response -> {
-                            if (!HttpResponseStatus.OK.equals(response.getStatus())) {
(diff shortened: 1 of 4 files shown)


mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/MantisMasterClientApi.java
```diff
@@ -699,16 +699,24 @@ public Observable<String> getJobStatusObservable(final String jobId) {
      * @return
      */
     public Observable<JobSchedulingInfo> schedulingChanges(final String jobId) {
+        final ConditionalRetry retryObject = new ConditionalRetry(null, "assignmentresults_" + jobId);
         return masterMonitor.getMasterObservable()
                 .filter(masterDescription -> masterDescription != null)
                 .retryWhen(retryLogic)
```
calvin681, line 705: There are multiple calls to `retryWhen`, do you need to change all of them?

I want to understand this scenario more. In a typical case, when a jobId is gone, it's because it is replaced by a new job (SLA 1/1). In that case, the old connection is properly closed, and we should unsubscribe to this observable normally. From your investigation, when are we fetching assignments indefinitely?

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/MantisMasterClientApi.java
```diff
@@ -727,7 +736,7 @@ public Observable<JobSchedulingInfo> schedulingChanges(final String jobId) {
                                     ;
                         }))
                 .repeatWhen(repeatLogic)
-                .retryWhen(retryLogic)
+                .retryWhen(retryObject.getRetryLogic())
```
sundargates, line 739: This feels like a convoluted way of saying not to retry after a "not found" exception.

Andyz26: exactly that's how this ConditionalRetry class is designed i think?

mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/MantisMasterClientApi.java
Andyz26: (there isn't enough logs available to prove it so this is only my theory): the job IDs being queried seem to belong to the jobs that went through bad deployments (e.g. SourceJobA-3 to SourceJobA-4, where 4 was in a bad state and then the source job was updated to SourceJobA-5). In this case, the observable chain to SourceJobA-4 wasn't closed/unsub properly so the query part was kept alive in the upper switchMap and stuck in its infinite retry. 

This can explain why only a small subset of source job IDs are being constantly queried in control plane logs (the ones not getting closed correctly).
Thus I think i only need to handle the one retry policy responsible for this infinite retry (the added test was stuck before the change and now can close the observable correctly).

PR #672, opened 2024-05-15
enable consumers to implement MasterMain and rename LeadershipManager to remove ZK

mcowgill-stripe:
### Context

In order for teams to re-implement the MasterMain functionality of the Mantis Server the ServiceLifecycle functions need to be available to teams. I've changed those to be public, but I'm happy to find another solution or extract an interface here.

UPDATE: I am reverting this part of the change.

Second the LeadershipManagerZKImpl to the best of my knowledge is a ZK independent leadership manager. We are implementing a leader elector based on DynamoDB and are able to use this LeadershipManager without modifying it. This name change is proposing to reflect the fact other teams can reuse the Leadership Manager when implementing their own leader election.

In order to change this name it appears I'll need to deprecate as some tests are failing. I'll consider this in a future PR to expand leader election options in Mantis.

### Checklist


sundargates (approved):

crioux-stripe:
Closing as this has been addressed in https://github.com/Netflix/mantis/pull/682 which was an extension of @mcowgill-stripe 's work.

PR #673, opened 2024-05-21
Update autoscaler strategies to utilize JobAutoScalerManager

hmitnflx:
### Context

Use newly introduced JobAutoScalerManager in autoscaler strategies. Use is controlled with a new knob that disables the manager by default. This autoscaler manager is used to control scale up and scale downs at runtime.

An example is during a failover event in a multi-region setup, all traffic is moved away from the failed over region—`evacuee`—to other region(s)—`savior(s)`. After extended period of evacuated state for the region, the jobs in that region would have been scaled down to lowest values. When the traffic is restored in that region, the job will see a big surge in incoming requests overwhelming it.

### Mitigation
Use the new `JobAutoScalerManager` that allows runtime control to accomplish this.

### Checklist


mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/WorkerExecutionOperationsNetworkStage.java
```diff
@@ -110,17 +111,20 @@ public class WorkerExecutionOperationsNetworkStage implements WorkerExecutionOpe
     private final List<Closeable> closeables = new ArrayList<>();
     private final ScheduledExecutorService scheduledExecutorService;
     private final ClassLoader classLoader;
+    private final FailoverStatusClient failoverStatusClient;
```
Andyz26, line 114: same as the question in runtimeTaskImpl.

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/WorkerMetricHandler.java
```diff
@@ -276,8 +279,8 @@ public void call() {
                             final int numWorkers = numStageWorkersFn.call(stage);
                             // get the aggregate metric values by metric group for all workers in stage
                             Map<String, GaugeData> allWorkerAggregates = getAggregates(listofAggregates);
-                            logger.info("Job stage " + stage + " avgResUsage from " +
-                                    workersMap.size() + " workers: " + allWorkerAggregates.toString());
+                            logger.info("Job stage {} avgResUsage from {} workers: {}", stage, workersMap.size(), allWorkerAggregates.toString());
+                            // TODO(hmitnflx): Add a failover mode based event into jobAutoScaleObserver
```
Andyz26, line 283: is this still valid?

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/RuntimeTaskImpl.java
```diff
@@ -71,9 +72,15 @@ public class RuntimeTaskImpl extends AbstractIdleService implements RuntimeTask
...
+    private FailoverStatusClient failoverStatusClient;
 
     public RuntimeTaskImpl() {
+        this(FailoverStatusClient.DEFAULT);
+    }
+
+    public RuntimeTaskImpl(FailoverStatusClient failoverStatusClient) {
         this.tasksStatusSubject = PublishSubject.create();
+        this.failoverStatusClient = failoverStatusClient;
     }
```
Andyz26, line 84: I think this will work fine on internal nfmantis wrapper but on OSS the instance is built from SingleTaskOnlyFactory so there is no path yet to inject this client to OSS yet?

Andyz26: I think it's ok for now but probably better to document this somewhere explicitly.

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/WorkerMetricHandler.java
hmitnflx: will remove

mantis-common/src/main/java/io/mantisrx/runtime/descriptor/StageScalingPolicy.java
```diff
@@ -162,7 +144,8 @@ public enum ScalingReason {
         ClutchRps,
         RPS,
         JVMMemory,
-        SourceJobDrop
+        SourceJobDrop,
+        FailoverAware
```
sundargates, line 148: what does ScalingReason::FailoverAware mean? I thought these were sent as events from the metric handler to the autoscaling policy.

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoScaler.java
```diff
@@ -357,15 +358,19 @@ public void scaleUpStage(final int numCurrentWorkers, final int desiredWorkers,
...
-          logger.warn("Job " + jobId + " stage " + stage + " is not scalable, can't decrement #workers by " + decrement);
+        final StageScalingPolicy scalingPolicy = stageSchedulingInfo.getScalingPolicy();
+        if (!scalingPolicy.isEnabled()) {
+          logger.warn("Job {} stage {} is not scalable, can't decrement #workers by {}", jobId, stage, decrement);
           return numCurrentWorkers;
         }
         if (numCurrentWorkers < 0 || decrement < 1) {
           logger.error("current number of workers({}) not known or decrement({}) < 1, will not scale down", numCurrentWorkers, decrement);
           return numCurrentWorkers;
+        } else if (!scalingPolicy.isAllowScaleDownDuringEvacuated() && failoverStatusClient.getFailoverStatus() == FailoverStatusClient.FailoverStatus.REGION_EVACUEE) {
```
sundargates, line 369: I think it would be better to separate out the AutoscalingPolicy similar to sources and sinks in mantis and have a Netflix-specific implementation in nfmantis. These changes seem to be specific to Netflix. However, it's not clear how much of a change that is and whether its feasible in the short term.

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/RuntimeTask.java
```diff
@@ -30,7 +31,8 @@ public interface RuntimeTask extends Service {
     void initialize(
         String executeStageRequestString,
         String workerConfigurationString,
-        UserCodeClassLoader userCodeClassLoader);
+        UserCodeClassLoader userCodeClassLoader,
+        FailoverStatusClient failoverStatusClient);
```
sundargates, line 35: It might be helpful to check if we can avoid leaking autoscaling-specific changes to the runtime task interface. For example, why can it not be initialized within nfmantis-runtime?

mantis-common/src/main/java/io/mantisrx/runtime/descriptor/StageScalingPolicy.java
hmitnflx: these are indeed sent as events but then we also need a failover autoscaling strategy similar to a `CPU` or `UserDefined` autoscaling strategy.

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/RuntimeTaskImpl.java
```diff
@@ -71,9 +72,15 @@ public class RuntimeTaskImpl extends AbstractIdleService implements RuntimeTask
     private Optional<Job> mantisJob = Optional.empty();
 
     private ExecuteStageRequest executeStageRequest;
+    private FailoverStatusClient failoverStatusClient;
```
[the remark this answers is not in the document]
hmitnflx: I'm sure I am missing something basic but here's my confusion — How could this be created directly in the scaler and provide a custom implementation in netflix fork?

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoScaler.java
hmitnflx: I think it's a good idea to refactor autoscaling policies for easy management.
That said, I'd argue that this isn't specific to Netflix since multi-region deployments of infra are quite common and region failover is a common mitigation strategy. 
With that in mind, it's a matter of time when quite a few jobs would require autoscaling during region failoves.

docs/docs/operate/autoscalingstrategies.md
```diff
@@ -3,25 +3,39 @@ many workers to scale up/down a stage. They fall into 2 main categories. Rule ba
...
 
-## Rule Based Strategy
+A stage's autoscaling policy is composed of following configurations -
+
+| Name                       | Type                    | Description                                                                                                        |
+|----------------------------|-------------------------|--------------------------------------------------------------------------------------------------------------------|
+| `stage`                    | `int`                   | Index of the stage (`1`-based since `0` is reserved for job master).                                               |
+| `min`                      | `int`                   | Minimum number of workers in the stage.                                                                            |
+| `max`                      | `int`                   | Maximum number of workers in the stage.                                                                            |
+| `increment`                | `int`                   | Number of workers to add when scaling up.                                                                          |
```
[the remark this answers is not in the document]
hmitnflx: fair point, will add a note about it.

mantis-common/src/main/java/io/mantisrx/runtime/descriptor/StageScalingPolicy.java
```diff
@@ -162,7 +85,8 @@ public enum ScalingReason {
...
```
Andyz26, line 89: should this match the other PR?

Andyz26 (approved):

mantis-common/src/main/java/io/mantisrx/runtime/descriptor/StageScalingPolicy.java
hmitnflx: yep, updated.

PR #679, opened 2024-06-11
Add handling to cancelled task on agent + cron spec update fix

Andyz26:
### Context
* Cron spec update is broken due to resetting the cron trigger ID early.
* There are cases where workers on agents can "leak" due to a shutdown event failing to get processed on the agent (the agent was in an unhealthy/busy/stuck) state while its resource gateway connection can recover. 

The new handling for the leaked worker's issue does the following:
* When the control plane retries the failed request to shut down the worker, the worker id is marked as canceled on the agent's state in the resource cluster actor.
* When the next heartbeat from the agent still shows the agent is occupied by the same worker ID, an error will be returned by the control plane to the heartbeat request, containing the body of the error with the worker ID.
* The agent's heartbeat handling in the resource gateway connection will handle the new error response with status code (not accepted) and invoke cancelTask(workerID) on the task executor to force another worker shutdown.
 

### Checklist


fdc-ntflx (approved):
I think the new semantics are clear and well defined. lgtm!

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
```diff
@@ -97,6 +97,8 @@ default CompletableFuture<List<TaskExecutorID>> getUnregisteredTaskExecutors() {
 
     CompletableFuture<Ack> addNewJobArtifactsToCache(ClusterID clusterID, List<ArtifactID> artifacts);
 
+    CompletableFuture<Ack> markExecutorTaskCancelled(WorkerId workerId);
+
```
sundargates, line 101: Why do you need an explicit method here? Can we somehow listen in on the event that's sent to the TaskExecutorGateway to know that the TE has been asked to shutdown.

sundargates: It looks like this method is only being called during scheduling failures. It should also be called whenever a worker is sending heartbeats that the JobActor / JobClusterActor is not acknowledging.

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/ResourceManagerGatewayCxn.java
```diff
@@ -218,19 +221,30 @@ protected void runIteration() throws Exception {
...
-            registered = false;
+    private void handleHeartbeatFailure(Throwable ex)
+        throws ExecutionException, InterruptedException, TimeoutException {
+        log.error("Failed to send heartbeat to gateway", ex);
+
+        if (ex instanceof TaskExecutorTaskCancelledException) {
+            log.error("[Fatal][LeakedTask] TaskExecutorTaskCancelledException received, killing current task: ", ex);
+            CompletableFuture<Ack> cancelFuture =
+                this.taskExecutor.cancelTask(((TaskExecutorTaskCancelledException) ex).getWorkerId());
+            cancelFuture.get(5, TimeUnit.SECONDS);
```
sundargates, line 239: nit: make this configurable?

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/master/resourcecluster/ResourceCluster.java
```diff
@@ -97,6 +97,8 @@ default CompletableFuture<List<TaskExecutorID>> getUnregisteredTaskExecutors() {
 
     CompletableFuture<Ack> addNewJobArtifactsToCache(ClusterID clusterID, List<ArtifactID> artifacts);
 
+    CompletableFuture<Ack> markExecutorTaskCancelled(WorkerId workerId);
```
sundargates, line 100: rename Task to Worker to be consistent

sundargates (approved):
LGTM!

PR #680, opened 2024-06-14
Introduce job autoscaler manager

hmitnflx:
### Context

This will allow runtime control over autoscaler, toggling on/off decisions to scale up and scale down a mantis job stage.
A use-case for this could be disabling autoscaler during a failover period to disable scale down.

### Checklist

### DIFF
--- a/mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoscalerManager.java
+++ b/mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoscalerManager.java
@@ -19,21 +19,24 @@ package io.mantisrx.server.worker.jobmaster;
 import java.util.Properties;
 
 /**
- * Failover status client interface to get the failover status of current region
+ * A manager to control autoscaling of a job. This allows a runtime control to autoscaling behavior
+ * without necessitating a job redeployment.
+ * An example use is disabling scale up or scale down during a maintenance window.
+ * Yet another example is during region failovers when a failed over region shouldn't scale down
  */
 public interface JobAutoscalerManager {
 
     JobAutoscalerManager DEFAULT = new NoopJobAutoscalerManager();
 
     /**
-     * Knobs to toggle autoscaling scale ups
+     * Knobs to toggle autoscaler scale ups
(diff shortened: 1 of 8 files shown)


calvin681 (approved):

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoscalerManager.java
```diff
@@ -0,0 +1,56 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.server.worker.jobmaster;
+
+import java.util.Properties;
+
+/**
+ * Failover status client interface to get the failover status of current region
```
Andyz26, line 22: nit: comment out of date

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoscalerManager.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+public interface JobAutoscalerManager {
```
Andyz26, line 24: what's the relationship between JobAutoscalerManager and the existing auto-scaling strategies? Does it apply on top of all the strategies?

Andyz26 (approved):

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoscalerManager.java
hmitnflx: thanks, addressed!

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoscalerManager.java
hmitnflx: yep, there's a way to enable the autoscaler manager for a job or jobcluster. If enabled, yes it'll apply on top of all autoscaling strategies.

PR #681, opened 2024-06-15
Set default parallelism for stage operators to the available number of processors

sundargates:
### Context

For stages that can be executed in parallel, set the default parallelism to the available number of cores.

### Checklist

### DIFF
--- a/mantis-runtime/src/main/java/io/mantisrx/runtime/executor/StageExecutors.java
+++ b/mantis-runtime/src/main/java/io/mantisrx/runtime/executor/StageExecutors.java
@@ -276,8 +276,13 @@ public class StageExecutors {
      * @return untyped to support multiple callers return types
      */
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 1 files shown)


mantis-runtime/src/main/java/io/mantisrx/runtime/executor/StageExecutors.java
```diff
@@ -276,41 +276,39 @@ private static <T, R> Observable<Observable<R>> executeInners(Observable<Observa
...
+        final long timeout,
+        final int concurrency) {
         logger.info("initializing {}", computation.getClass().getCanonicalName());
         computation.init(context);
 
         // from groups to observable
         final Func2<Context, Observable<T>, Observable<R>> c
                 = (Func2<Context, Observable<T>, Observable<R>>) computation;
 
-        if (concurrency == StageConfig.DEFAULT_STAGE_CONCURRENCY) {
```
calvin681, line 288: In this original code, if the concurrency is 1, then you don't need to do a groupBy and flatMap. Which may be more efficient. I feel we should keep that.

PR #689, opened 2024-07-10
expose jobVersion, jobJarUrl as part of RuntimeTask interface

sarahwada-stripe:
### Context
Expose `jobVersion`, `jobJarUrl` as part of RuntimeTask interface.

This allows us to add version and sha labels to our job metrics. We'd like these metric labels so we can monitor our automated job deploys.

### Checklist

### DIFF
--- a/mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/RuntimeTaskImpl.java
+++ b/mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/RuntimeTaskImpl.java
@@ -239,5 +239,4 @@ public class RuntimeTaskImpl extends AbstractIdleService implements RuntimeTask 
     public String getWorkerId() {
         return executeStageRequest.getWorkerId().getId();
     }
-
 }
--- a/mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/RuntimeTaskImpl.java
+++ b/mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/RuntimeTaskImpl.java
@@ -240,11 +240,4 @@ public class RuntimeTaskImpl extends AbstractIdleService implements RuntimeTask 
         return executeStageRequest.getWorkerId().getId();
     }
 
-    public String getJobVersion() {
-        return executeStageRequest.getJobVersion();
-    }
-
-    public String getJobJarURL() {
-        return executeStageRequest.getJobJarUrl().toString();
-    }
 }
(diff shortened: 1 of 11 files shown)


mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/RuntimeTask.java
```diff
@@ -33,4 +33,9 @@ void initialize(
         UserCodeClassLoader userCodeClassLoader);
 
     String getWorkerId();
+
```
Andyz26, line 36: I would prefer we do not add specific metrics/metadata into these entry-interfaces. can you elaborate on why it's on the runtimeTask for metrics instead of using regular metrics collector tagging? I feel like there are existing metrics path to do this without adding into this interface.

Andyz26 (changes requested):
See comments above.

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/RuntimeTask.java
sarahwada-stripe: @Andyz26 thanks for the review! 

> can you elaborate on why it's on the runtimeTask for metrics instead of using regular metrics collector tagging? 
The job metadata is only available as part of the job submission, not the actual running job itself. From my understanding, we need to hook into the task/job submission to get information like version. Right now, we have written our own AgentMain class, and we can hook into the TaskExecutor `onTaskStarting` to add custom job labels, like the scaffolding that exists here: https://github.com/Netflix/mantis/blob/c1c7f34a1cdf2fea6a84575eb6f6b94eb3dc71a7/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/AgentV2Main.java#L59 . 

Given the context above, do you have a recommendation for a better place to make this change? Happy to move things around, it's possible I've missed a better metrics path when going through the code.

Andyz26: imo since agent to inner task is not always 1 to 1 it would be more consistent to construct such metrics within the worker runtime where you have access to things like MetricsRegistry (I don't know how your current metrics integration is done into with this one).

sarahwada-stripe: got it, thanks for the feedback! are you opposed to adding `jobVersion` to the `ExecuteStageRequest`? I think we'll need it to capture the `jobVersion` on the worker

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/RuntimeTaskImpl.java
```diff
@@ -239,4 +239,5 @@ protected Observable<Status> getStatus() {
     public String getWorkerId() {
         return executeStageRequest.getWorkerId().getId();
     }
+
```
Andyz26, line 242: nit: remove blank line

Andyz26 (approved):

PR #691, opened 2024-07-15
DynamoDB Leader Monitor: On Next Should Always Provide MasterDescription

kmg-stripe:
It appears that specifying null to the observer results in a weird terminal condition where all values after null are not observed.

ToDo(kmg): Need to verify this and create or add to the unit tests.

### Context

### Checklist


crioux-stripe:
Closing in favor of https://github.com/Netflix/mantis/pull/692

PR #693, opened 2024-07-18
Fix maven missing deps: upgrade nebula + gradle

Andyz26:
### Context

To fix the removed maven dependencies, the Nebula oss version + gradle version needs to be updated.
* New gradle (8.x) is also causing error on the extra copyLibs task in control-plane-core which now requires explicit declaration on downstream consumers.
* Baseline plugin new version requires jdk17. Disable for now.

### Checklist

### DIFF
--- a/baseline.gradle
+++ b/baseline.gradle
@@ -14,15 +14,15 @@
  * limitations under the License.
  */
 
-allprojects {
-    apply plugin: 'com.palantir.baseline-idea'
-}
+// allprojects {
+//     apply plugin: 'com.palantir.baseline-idea'
+// }
 
-subprojects {
-    // Currently, if any subproject applies the blanket Baseline plugin, it forces the Baseline plugin
-    // to be applied to ALL projects. And we are not prepared to address all of the build errors that
-    // occur as a result at this time.
+// subprojects {
+//     // Currently, if any subproject applies the blanket Baseline plugin, it forces the Baseline plugin
+//     // to be applied to ALL projects. And we are not prepared to address all of the build errors that
+//     // occur as a result at this time.
 
-    apply plugin: 'com.palantir.baseline-exact-dependencies'
-    apply plugin: 'com.palantir.baseline-format'
-}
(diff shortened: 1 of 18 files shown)


build.gradle
```diff
@@ -104,9 +105,8 @@ project.snapshot.configure { finalizedBy printAllReleasedArtifacts }
 subprojects {
     apply plugin: 'java-library'
```
hmitnflx, line 106: How about we add explicit dep on jar task for control-plane project here in `subprojects` block instead of adding to individual projects.

baseline.gradle
```diff
@@ -14,15 +14,15 @@
  * limitations under the License.
  */
 
-allprojects {
-    apply plugin: 'com.palantir.baseline-idea'
-}
+// allprojects {
```
hmitnflx, line 17: delete altogether instead of commenting?

hmitnflx (approved):

build.gradle
Andyz26: This is a specific handling between control-plane-core etc. and it feels a bit too generic to add these rules for all sub-proj?

build.gradle
```diff
@@ -27,7 +27,7 @@ buildscript {
     dependencies {
         classpath 'com.netflix.nebula:gradle-netflixoss-project-plugin:11.5.0'
         classpath 'com.netflix.nebula:nebula-dependency-recommender:11.+'
-        classpath 'io.mantisrx:mantis-gradle-plugin:1.2.6-rc.3'
+        classpath 'io.mantisrx:mantis-gradle-plugin:1.2.6-rc.4'
```
hmitnflx, line 30: nit: Pl consider using a release version instead of a candidate.

hmitnflx (approved):

build.gradle
Andyz26: yeha will switch once this is fully fixed. (it's an open pr on the other side).

Andyz26: (currently blocked due to gradle plugin repo publish failure).

PR #698, opened 2024-07-30
Client Should Not Retry 400

kmg-stripe:
### Context

We noticed that workers will retry `400` from the master.  This should fail fast with an error.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/BadRequestException.java
+++ b/mantis-control-plane/mantis-control-plane-client/src/main/java/io/mantisrx/server/master/client/BadRequestException.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2024 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.server.master.client;
+
+public class BadRequestException extends Exception {
+    public BadRequestException(String msg) {
+        super(String.format("Server returned bad request: %s", msg));
+    }
+}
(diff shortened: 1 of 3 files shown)


kmg-stripe:
> @kmg-stripe, in which case do you see the control plane throwing bad request results for this call?

@Andyz26 - full disclosure, this arose while one of our clusters was in a really bad state due to a few dynamodb provider bugs (they are now fixed).  

the issue was that leader would discover all jobs, but could not discover most job workers on start-up, which would lead to excessive re-submissions.  since a new leader is elected every time we deploy to masters, this was happening pretty often.  our theory was that old agents would ask the leader about assignments, and the leader was returning `400` because of the state inconsistency.  the `400` itself wasn't a huge issue, but it was spamming the logs, and would would expect the agents to fail fast instead of retrying.

kmg-stripe:
> > > @kmg-stripe, in which case do you see the control plane throwing bad request results for this call?
> > 
> > 
> > @Andyz26 - full disclosure, this arose while one of our clusters was in a really bad state due to a few dynamodb provider bugs (they are now fixed).
> > the issue was that leader would discover all jobs, but could not discover most job workers on start-up, which would lead to excessive re-submissions. since a new leader is elected every time we deploy to masters, this was happening pretty often. our theory was that old agents would ask the leader about assignments, and the leader was returning `400` because of the state inconsistency. the `400` itself wasn't a huge issue, but it was spamming the logs, and would would expect the agents to fail fast instead of retrying.
> 
> my main concern is that since this is a API used in many places, early termination on unexpected control plane result might cause workers to exit/restart when the control plane is in a transient bad state and disrupt critical jobs which currently can survive such problems. In this case would it be cleaner to have control plane emit not_found if it meant to kill the worker in such case?

Ah, understood!  Let me find the offending `400` and we can change it to a `404`, instead of handling the `400` in the client.

kmg-stripe:
@Andyz26 Given the state our clusters were in and your comment about transient issues with master, I can close this.

For completeness, here is what I think happened:

- Due to the weird state of our clusters, we were hitting "Failed to get Sched info stream info":  
https://github.com/Netflix/mantis/blob/a2934008e3976209673c5d1239fb2d5ea4924f91/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/handlers/JobDiscoveryRouteHandlerAkkaImpl.java#L134

- It looks like this results in `CLIENT_ERROR_NOT_FOUND`:

https://github.com/Netflix/mantis/blob/a2934008e3976209673c5d1239fb2d5ea4924f91/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/JobClustersManagerActor.java#L736

- Then, I think `CLIENT_ERROR_NOT_FOUND` is transformed into `BAD_REQUEST`:

https://github.com/Netflix/mantis/blob/a2934008e3976209673c5d1239fb2d5ea4924f91/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v0/BaseRoute.java#L58

PR #699, opened 2024-07-30
Fix WorkersByJobId: Filtered Workers Not Empty Check

kmg-stripe:
### Context

I forgot to update the previous PR with this fix.  We have enough cruft lying around from all of the churn, that we hit cases where we get a NPE when we cannot parse the worker payload in `getAllWorkersByJobId`.  This checks to ensure `workers` is not empty before referencing the first element of `workers`.

### Checklist


kmg-stripe:
@hmitnflx - If you have a sec, could you have a look at this one?  Should be quick.

crioux-stripe:
@Andyz26 Hey! One more for our production readiness if you don't mind.

crioux-stripe (approved):

crioux-stripe:
@Andyz26 Sorry about self-approving on this one. @kmg-stripe and I wanted to make sure that we made our target of releasing today.

PR #703, opened 2024-08-14
Fixes to Source Job Connector and Master Monitor Logging

kmg-stripe:
1. All source job connectors have ZK config hard-coded due to parameter-less
`MantisSourceJobConnectorFactory`.  Configuring a `MantisClient` in this context
doesn't make sense, since we can reference the configured `HighAvailabilityServices`

2. `DynamoDBMasterMonitor` was erroneously logging parse errors when reading
leadership info from the DB.

3. `HighAvailabilityServicesImpl` was setting the leader state to NULL, which lead
to `ResourceClusterGatewayClient` constructing an invalid URI.  Updating to not
update the `ResourceClusterGatewayClient` when master is NULL.

### Context

### Checklist

### DIFF
--- a/mantis-client/src/main/java/io/mantisrx/client/MantisSSEJob.java
+++ b/mantis-client/src/main/java/io/mantisrx/client/MantisSSEJob.java
@@ -23,6 +23,7 @@ import io.mantisrx.runtime.descriptor.SchedulingInfo;
 import io.mantisrx.runtime.parameter.Parameter;
 import io.mantisrx.runtime.parameter.SinkParameters;
 import io.mantisrx.server.master.client.ConditionalRetry;
+import io.mantisrx.server.master.client.HighAvailabilityServices;
 import io.mantisrx.server.master.client.NoSuchJobException;
 import io.reactivx.mantis.operators.DropOperator;
 import java.io.Closeable;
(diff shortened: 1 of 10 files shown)


mantis-client/src/main/java/io/mantisrx/client/MantisSSEJob.java
```diff
@@ -179,10 +180,6 @@ public static class Builder {
         private Observer<SinkConnectionsStatus> sinkConnectionsStatusObserver = null;
         private long dataRecvTimeoutSecs = 5;
 
-        public Builder(Properties properties) {
-            this(new MantisClient(properties));
-        }
-
         public Builder() {
```
kmg-stripe, line 183: We were seeing log spam, where a JobSource triggers this code path.  It doesn't appear to break anything, but seems a bit weird and leads to log spam (`HighAvailabilityServicesUtil` logs a few warnings, because this path attempts to "create" the HA services with no properties).

My understanding is that the HA services should generally be configured if we are here, so should just try to get the reference to HA services.

kmg-stripe:
cc: @codyrioux

kmg-stripe:
@Andyz26 - Just a ping...  I was also hoping to cut a release after this one.

crioux-stripe (approved):
LGTM.

PR #706, opened 2024-08-28
use jobJarUrl instead of artifactName to avoid stripping path prefixes

sarahwada-stripe:
### Context
I wanted to open a PR to kick off a discussion about the jobJarURL format that the mantis-master expects. It looks like the mantis-master expects the jobJarURL in the JobConfig to have a single path component, e.g. `https://my-job-jar.zip`, and drops path information when there are multiple path elements, e.g. `https://git-sha-prefix/my-job-jar.zip`.

A reproduction of the behavior that I've observed is:
1. Write a jobConfig with `jobJarFileLocation` like `https://git-sha-prefix/my-job-jar.zip`
2. Master submits a job to agent with the jobJarUrl `https://git-sha-prefix/my-job-jar.zip`
3. On master restart, the master strips the path prefixes, and persists the jobJarUrl `https://my-job-jar.zip` to the KV persistence store.
4. If a job is resubmitted to an agent, it's submitted with the jobJarUrl `https://my-job-jar.zip`.

I've narrowed down this behavior to a few places:
1. When a job is read from the KV store, various configurations (like `JobClusterConfig`) only store the **artifactName**, and do not store the full jobJarUrl.
2. When a job is sent to the agent or persisted, the master uses `generateURL(artifactName)` to generate the jobJarUrl, losing the path prefix information.

I'm curious if there is a deeper mantis or Netflix reason for this behavior, and if changing this behavior will break any other assumptions within Mantis. Also, FWIW we can't use a chained `PrefixedBlobStore` because these prefixes are dynamic.

@Andyz26 would appreciate any thoughts on this change, especially as it changes the JobClusterConfig interface!

cc @kmg-stripe (I'm going to be out of office and Kevin may take over this PR)

### Checklist

Tested E2E on local mantis cluster (master + DynamoDB + agent)

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
@@ -1511,6 +1511,7 @@ public class JobClusterActor extends AbstractActorWithTimers implements IJobClus
         return
             new JobDefinition.Builder()
                 .withJobSla(new JobSla.Builder().build())
+                .withJobJarUrl(clusterConfig.getJobJarUrl())
                 .withArtifactName(clusterConfig.getArtifactName())
                 .withVersion(clusterConfig.getVersion())
                 .withLabels(clusterDefinition.getLabels())
@@ -1657,6 +1658,9 @@ public class JobClusterActor extends AbstractActorWithTimers implements IJobClus
     private void validateJobDefinition(JobDefinition definition) throws InvalidJobRequestException {
         if (definition == null) {
             throw new InvalidJobRequestException("MantisJobDefinition cannot be null");
+        }
+        if (definition.getJobJarUrl() == null) {
+            throw new InvalidJobRequestException("MantisJobDefinition job jobJarUrl attribute cannot be null");
         }
         if (definition.getArtifactName() == null) {
             throw new InvalidJobRequestException("MantisJobDefinition job artifactName attribute cannot be null");
@@ -2174,6 +2178,7 @@ public class JobClusterActor extends AbstractActorWithTimers implements IJobClus
                 return;
             }
             JobClusterConfig newConfig = new JobClusterConfig.Builder().from(jobClusterMetadata.getJobClusterDefinition().getJobClusterConfig())
+                // TODO(swada): do we need to add jobJarURL to UpdateJobClusterArtifactRequest as well?
                     .withArtifactName(artifactReq.getArtifactName())
                     .withVersion(artifactReq.getVersion())
                     .withUploadedAt(System.currentTimeMillis())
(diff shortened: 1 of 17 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/DataFormatAdapter.java
```diff
@@ -151,11 +151,10 @@ public static List<JobClusterConfig> convertJarsToJobClusterConfigs(List<NamedJo
 
     public static NamedJob.Jar convertJobClusterConfigToJar(JobClusterConfig jConfig) throws MalformedURLException {
         SchedulingInfo sInfo = jConfig.getSchedulingInfo();
-        String name = jConfig.getArtifactName();
         long uploadedAt = jConfig.getUploadedAt();
         String version = jConfig.getVersion();
 
-        return new NamedJob.Jar(generateURL(name), uploadedAt, version, sInfo);
+        return new NamedJob.Jar(new URL(jConfig.getJobJarUrl()), uploadedAt, version, sInfo);
```
sarahwada-stripe, line 157: This is the usage of the `generateURL` method referenced in the PR body

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -2174,6 +2178,7 @@ public void onJobClusterUpdateArtifact(UpdateJobClusterArtifactRequest artifactR
                 return;
             }
             JobClusterConfig newConfig = new JobClusterConfig.Builder().from(jobClusterMetadata.getJobClusterDefinition().getJobClusterConfig())
+                // TODO(swada): do we need to add jobJarURL to UpdateJobClusterArtifactRequest as well?
```
sarahwada-stripe, line 2181: Open question: need to look into this

Andyz26:
@sundargates do you have any context on why the jar path is treated this way? I assume it has something todo with how we override the path internally?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/domain/JobClusterConfig.java
```diff
@@ -99,6 +108,13 @@ public static class Builder {
 
         public Builder() {}
 
+        public Builder withJobJarUrl(String jobJarUrl) {
+            Preconditions.checkNotNull(jobJarUrl, "artifactName cannot be null");
```
Andyz26, line 112: nit: artifactName/jobJarUrl

Andyz26 (commented):
overall lgtm. Can you add some back compact tests too?

sundargates:
> @sundargates do you have any context on why the jar path is treated this way? I assume it has something todo with how we override the path internally?

It was done this way because we wanted the paths to be compatible with the old mesos based mantisagent download scripts. I don't think there's any requirement that it has to be kept this way in the current world.

crioux-stripe:
Closing in favor of https://github.com/Netflix/mantis/pull/708.

PR #707, opened 2024-09-05
Support fallback blob store

Andyz26:
### Context
Add support to a fallback blob store which will be used if the default storage cannot fetch the target file. This is used for blob store migration.

### Checklist

### DIFF
--- a/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/BlobStore.java
+++ b/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/BlobStore.java
@@ -23,6 +23,7 @@ import java.net.URI;
 import java.util.concurrent.locks.Lock;
 import lombok.AccessLevel;
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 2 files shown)


mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/BlobStore.java
```diff
@@ -67,6 +72,23 @@ static BlobStore forHadoopFileSystem(URI clusterStoragePath, File localStoreDir)
                 .withThreadSafeBlobStore();
     }
 
+    static BlobStore forHadoopFileSystem(URI clusterStoragePath, URI fallbackStoragePath, File localStoreDir)
```
sundargates, line 75: nit: why custom API for this?

sundargates (approved):
LGTM!

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/BlobStore.java
Andyz26: what do you mean by custom API? This is overloading the same API above with extra param.

PR #708, opened 2024-09-06
 use jobJarUrl instead of artifactName to avoid stripping path prefixes

kmg-stripe:
### Context
This is a fork of @sarahwada-stripe's PR (#706).  I'll take this over, since she is OOO.

@Andyz26 - I addressed the few comments in the original PR.  I'll get back to this next week and will add the requested backwards compatible tests.

I wanted to open a PR to kick off a discussion about the jobJarURL format that the mantis-master expects. It looks like the mantis-master expects the jobJarURL in the JobConfig to have a single path component, e.g. `https://my-job-jar.zip`, and drops path information when there are multiple path elements, e.g. `https://git-sha-prefix/my-job-jar.zip`.

A reproduction of the behavior that I've observed is:
1. Write a jobConfig with `jobJarFileLocation` like `https://git-sha-prefix/my-job-jar.zip`
2. Master submits a job to agent with the jobJarUrl `https://git-sha-prefix/my-job-jar.zip`
3. On master restart, the master strips the path prefixes, and persists the jobJarUrl `https://my-job-jar.zip` to the KV persistence store.
4. If a job is resubmitted to an agent, it's submitted with the jobJarUrl `https://my-job-jar.zip`.

I've narrowed down this behavior to a few places:
1. When a job is read from the KV store, various configurations (like `JobClusterConfig`) only store the **artifactName**, and do not store the full jobJarUrl.
2. When a job is sent to the agent or persisted, the master uses `generateURL(artifactName)` to generate the jobJarUrl, losing the path prefix information.

I'm curious if there is a deeper mantis or Netflix reason for this behavior, and if changing this behavior will break any other assumptions within Mantis. Also, FWIW we can't use a chained `PrefixedBlobStore` because these prefixes are dynamic.

### Checklist

Tested E2E on local mantis cluster (master + DynamoDB + agent)

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/proto/JobClusterProtoAdapter.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/proto/JobClusterProtoAdapter.java
@@ -60,7 +60,9 @@ public class JobClusterProtoAdapter {
 
         final CreateJobClusterRequest request = new CreateJobClusterRequest(new JobClusterDefinitionImpl(
                 jd.getName(),
-                Arrays.asList(new JobClusterConfig(jd.getJobJarFileLocation().toString(),
+                Arrays.asList(new JobClusterConfig(
+                        jd.getJobJarFileLocation().toString(),
+                        jd.getJobJarFileLocation().toString(),
                         System.currentTimeMillis(),
                         jd.getVersion(),
                         jd.getSchedulingInfo()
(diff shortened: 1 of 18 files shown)


Andyz26 (approved):

crioux-stripe:
Merging this one in. Thanks Andy!

kmg-stripe:
@codyrioux Just tested this branch with the internal integration PR, which is working.  I'll add the additional tests today,

crioux-stripe:
@kmg-stripe Thank you, and no rush. Once you get those in I'll see about getting this and #713 merged and getting an RC cut.

PR #710, opened 2024-09-09
Fix blobstore fallback error type

Andyz26:
### Context

The underlying blob store use runtime exceptions for auth-related errors.
Switching catch to use base exception type.

### Checklist

### DIFF
--- a/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/BlobStore.java
+++ b/mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/BlobStore.java
@@ -118,7 +118,7 @@ public interface BlobStore extends Closeable {
             {
                 return blobStore.get(blobUrl);
             }
-            catch (IOException e) {
+            catch (Exception e) {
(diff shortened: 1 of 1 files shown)


mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/BlobStore.java
```diff
@@ -118,7 +118,7 @@ public File get(URI blobUrl) throws IOException {
             {
                 return blobStore.get(blobUrl);
```
sundargates, line 119: Shouldn't the blobstore only be throwing IOException?

sundargates (approved):
I would maybe avoid this check if possible given that the BlobStore is only expected to throw IOException

mantis-server/mantis-server-agent/src/main/java/io/mantisrx/server/agent/BlobStore.java
Andyz26: tested this one and the actual error thrown on access deny is runtime exception.

PR #714, opened 2024-09-19
Add a connectToAllJobs option to MantisClient getSinkClient

tcai-stripe:
### Context

Currently `JobSource` attempts to fetch the scheduling information for the `latest` target job. This can be painful if a new job is spun up but has not fully Launched. Most of the time we're looking to communicate with latest and only latest but there are cases where ALL would be more useful.

exposes an option to configure whether the MantisClient get the latest SinkClient or all SinkClients.

### Checklist


tcai-stripe:
cc @crioux-stripe

crioux-stripe (changes requested):
Hey @tcai-stripe!

This is a great start to this. Two changes I'd like to see are:

* Add tests for `mergeSinkClients` just to make sure it is doing what we expect.
* Plumb a [parameter](https://github.com/Netflix/mantis/blob/a59f5f659ad105e0c5aea948dadd748655c09911/mantis-connectors/mantis-connector-job-source/src/main/java/io/mantisrx/connector/job/source/JobSource.java#L74-L85) into `JobSource` that correctly passes this value down.

crioux-stripe:
I think we may want to close this in favor of https://github.com/Netflix/mantis/pull/717 which achieves what we want but in a slightly different fashion.

PR #715, opened 2024-10-02
ConsistentHashingRouter: Use XXH3 instead of 4 byte ketama for consis…

crioux-stripe:
Eliminates a repeatable hash collision with a 40 worker stage 2. We can either go with extending the bits in `ketama` or using `xxh3`. I'm recommending `xxh3`. 

### Context
We ran into hash collisions when hashing slots on a 40 worker stage 2. This was a bit unexpected but very repeatable in the unit tests. If either of the unit tests added are run with the `ketama` hash function used in Mantis by default the tests will fail. One test will compute 40,000 hashes per call to `route`. The other will compute 39999 unique hashes from 40000 ring buffer entries. The values used are real values generated by a 40 worker stage 2 on a `GroupToScalar` stage.

**Note for Netflix**: I'd check logs from stage 1 of any of your jobs that have over 28 or more indices on the second stage. Its very likely that the difference in resource cluster types (you're using higher CPU to Network ratio boxes as well as Intel vs our ARM) is hiding the fact that some jobs might be doing ring recomputation on every chunk send. You should see the log line `Recomputing ring due to difference in number of connections versus cache` being spammed very frequently on impacted jobs.

Our investigation shows:

```
stage_2_index_28_partition_1-241 collided with: 
stage_2_index_2_partition_2-213
stage_2_index_6_partition_1-126 collided with: 
stage_2_index_34_partition_2-876
```

### Checklist

### DIFF
--- a/mantis-network/build.gradle
+++ b/mantis-network/build.gradle
@@ -22,7 +22,7 @@ ext {
 dependencies {
     api "io.netty:netty-handler:$nettyVersion"
     implementation "io.mantisrx:mql-jvm:$mqlVersion"
-    implementation "net.openhft:zero-allocation-hashing:0.16"
+    implementation "net.openhft:zero-allocation-hashing:0.+"
     api project(':mantis-common')
     compileOnly libraries.spectatorApi
     testImplementation libraries.spectatorApi
--- a/mantis-network/build.gradle
+++ b/mantis-network/build.gradle
@@ -22,10 +22,11 @@ ext {
 dependencies {
     api "io.netty:netty-handler:$nettyVersion"
     implementation "io.mantisrx:mql-jvm:$mqlVersion"
+    implementation "net.openhft:zero-allocation-hashing:0.16"
     api project(':mantis-common')
     compileOnly libraries.spectatorApi
     testImplementation libraries.spectatorApi
-    
+
     testImplementation libraries.junitJupiter
     testImplementation libraries.mockitoCore
     testImplementation libraries.slf4jLog4j12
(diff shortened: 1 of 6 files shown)


crioux-stripe:
I've added a test showing even a 500 worker `GroupToScalar` stage with 4 partitions (double the default) will not cause a collision under the new scheme.

Andyz26:
thanks a lot @crioux-stripe! I searched around, and I don't see impacted jobs on our side yet (my guess is that the multi-stage jobs we have tend to have very small stage 2+, so it's less likely to hit the collision).

mantis-network/build.gradle
```diff
@@ -22,10 +22,11 @@ ext {
 dependencies {
     api "io.netty:netty-handler:$nettyVersion"
     implementation "io.mantisrx:mql-jvm:$mqlVersion"
+    implementation "net.openhft:zero-allocation-hashing:0.16"
```
Andyz26, line 25: any reason to pin to this version?

mantis-network/src/main/java/io/reactivex/mantis/network/push/HashFunctions.java
```diff
@@ -37,6 +38,26 @@ public long computeHash(byte[] keyBytes) {
         };
     }
 
+    public static HashFunction ketamaExtended() {
```
Andyz26, line 41: do we need to keep this if the default impl switched to use xxh3?

Andyz26 (approved):

mantis-network/src/main/java/io/reactivex/mantis/network/push/HashFunctions.java
crioux-stripe: We do not. I'll make the change.

mantis-network/build.gradle
crioux-stripe: Will unpin!

PR #717, opened 2024-10-04
Change Job discovery to use most recent launched job

Andyz26:
### Context
I would like to revive this [PR](https://github.com/Netflix/mantis/pull/549) from @hmitnflx to use the last launched job state instead of the last submitted to avoid downstream consumer jobs being stuck with a new submission at an error state or too long time to allocate resources.

* Replace discovery subject source from last submitted to last launched.
* Fixed and added UT.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
@@ -136,6 +136,7 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


Andyz26:
hi @crioux-stripe this is a behavior change on job discovery, so please lmk if you have any concerns.

hmitnflx (commented):
Is there any scenario where semantics of `jobSubmittedSubject` are of importance? like for SLA min/max!
It might be safer to expose this in the endpoint only possibly conditional on a requestParam...

Andyz26:
> Is there any scenario where semantics of `jobSubmittedSubject` are of importance? like for SLA min/max! It might be safer to expose this in the endpoint only possibly conditional on a requestParam...

I searched around the usages, and I don't think I see anything that fits better with last-submitted (since this is impacting clients, I think it shouldn't hurt to switch the full behavior to only start connecting after the new job instance becomes connectable).

crioux-stripe:
Amazing!! We have wanted this change but figured it might be painful for such a big behavior shift at Netflix where there are thousands of jobs.

We're enthusiastically in support of this.

hmitnflx (approved):

PR #719, opened 2024-10-05
Prevent per-node scheduling fallback during batch scheduling retry 

Andyz26:
### Context
Currently, when a new job gets submitted, all the workers get scheduled in batch to have an all-or-nothing manner. However, the job actor heartbeat check will also try to re-schedule a worker if it's "stuck" in the allocation phase for too long (based on the worker-heartbeat-timeout config). Thus the batch scheduling gets invalidated after some timeout (which could be problematic for larger jobs when we need more time to get the requested resource allocated from the cluster auto scaler).

Behavior changes here:
* Batch scheduling failure will retry without attempt limit. (We will rely on the cancel request message from its job actor to interrupt).
* JobActor heartbeat routine will not act on unscheduled workers.
* fixed the akka-tests and added these back to the CI build.

### Checklist

### DIFF
--- a/.github/workflows/nebula-ci.yml
+++ b/.github/workflows/nebula-ci.yml
@@ -40,7 +40,7 @@ jobs:
           restore-keys: |
             - ${{ runner.os }}-gradlewrapper-
       - name: Build with Gradle
-        run: ./gradlew --info --stacktrace build --warning-mode=all
+        run: ./gradlew --info --stacktrace build akkatest --warning-mode=all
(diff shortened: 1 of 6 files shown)


Andyz26:
@fdc-ntflx ptal

fdc91 (approved):
LGTM!

fdc-ntflx (approved):
Sorry I reviewed it with the wrong account. LGTM

PR #720, opened 2024-10-07
Allow user specified concurrency for Scalar to Group and Group to Scalar

crioux-stripe:
### Context
The default concurrency for `ScalarToGroup` and `GroupToScalar` stages is `-1` and thus relies on the number of inner observables when executing the stage. There may be cases such as CPU bound stages where the user wants fine grained control over how many threads are executing the stage logic.

Some discussion in #681 indicates a desire for this, but I agree with @calvin681 that changing defaults shouldn't be done lightly as users have come to expect certain threading behavior with the stages. I think this represents a compromise where unaware users can continue with the default and power users can begin to control threading.

### Checklist

### DIFF
--- a/mantis-runtime/src/main/java/io/mantisrx/runtime/KeyValueStageConfig.java
+++ b/mantis-runtime/src/main/java/io/mantisrx/runtime/KeyValueStageConfig.java
@@ -33,7 +33,12 @@ public abstract class KeyValueStageConfig<T, K, R> extends StageConfig<T, R> {
     private final Codec<K> keyCodec;
 
     public KeyValueStageConfig(String description, Codec<?> inputKeyCodec, Codec<T> inputCodec, Codec<K> outputKeyCodec, Codec<R> outputCodec, INPUT_STRATEGY inputStrategy, List<ParameterDefinition<?>> params) {
-        super(description, inputKeyCodec, inputCodec, outputCodec, inputStrategy, params);
+        super(description, inputKeyCodec, inputCodec, outputCodec, inputStrategy, params, DEFAULT_STAGE_CONCURRENCY);
+        this.keyCodec = outputKeyCodec;
+    }
+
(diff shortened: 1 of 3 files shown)


mantis-runtime/src/main/java/io/mantisrx/runtime/KeyValueStageConfig.java
```diff
@@ -33,7 +33,12 @@ public abstract class KeyValueStageConfig<T, K, R> extends StageConfig<T, R> {
     private final Codec<K> keyCodec;
 
     public KeyValueStageConfig(String description, Codec<?> inputKeyCodec, Codec<T> inputCodec, Codec<K> outputKeyCodec, Codec<R> outputCodec, INPUT_STRATEGY inputStrategy, List<ParameterDefinition<?>> params) {
-        super(description, inputKeyCodec, inputCodec, outputCodec, inputStrategy, params);
+        super(description, inputKeyCodec, inputCodec, outputCodec, inputStrategy, params, DEFAULT_STAGE_CONCURRENCY);
```
hmitnflx, line 36: nit: you could call the other constructor in this class instead of duplicate call to super + keyCodec assignment.

mantis-runtime/src/main/java/io/mantisrx/runtime/ScalarToGroup.java
```diff
@@ -116,6 +118,12 @@ public Config<T, K, R> concurrentInput() {
             return this;
         }
 
+        public Config<T, K, R> concurrentInput(final int concurrency) {
```
hmitnflx, line 121: this could cause issues, no?

hmitnflx (approved):

mantis-runtime/src/main/java/io/mantisrx/runtime/ScalarToGroup.java
crioux-stripe: I went looking around and we do [allocate the same key to the same thread every time](https://github.com/Netflix/mantis/blob/a59f5f659ad105e0c5aea948dadd748655c09911/mantis-runtime/src/main/java/io/mantisrx/runtime/executor/StageExecutors.java#L238). We have concurrency either way, this just lets the user specify it precisely, unless I'm misunderstanding the if/else that contains the linked line.

PR #721, opened 2024-10-31
Add a buffering functionality to the SseWorkerConnection

crioux-stripe:
### Context

The `SseWorkerConnection` uses a `DropOperator` with a metric named suffixed with `withBuffer` set in the [SseWorkerConnectionFunction](https://github.com/Netflix/mantis/blob/master/mantis-server/mantis-server-worker-client/src/main/java/io/mantisrx/server/worker/client/SseWorkerConnectionFunction.java#L61) class. The class takes a `bufferSize` parameter but does nothing with it! I've had the illusion that we've been buffering here for darn near a decade!

This change is pretty straightforward implementation. `rebatchRequests` simply sends `n` requests into the `DropOperator` and continues to stay ahead of the requests made by the downstream by up to n. There is a batching implementation inside of the operator that RxJava uses to implement this.

Users could then set the buffer with the: `mantisClient.buffer.size` and `workerClient.buffer.size` properties.

### Checklist

### DIFF
--- a/mantis-server/mantis-server-worker-client/src/main/java/io/mantisrx/server/worker/client/SseWorkerConnection.java
+++ b/mantis-server/mantis-server-worker-client/src/main/java/io/mantisrx/server/worker/client/SseWorkerConnection.java
@@ -305,6 +305,7 @@ public class SseWorkerConnection {
         }
         return response.getContent()
                 .lift(new DropOperator<ServerSentEvent>(metricGroupId))
+                .rebatchRequests(this.bufferSize <= 0 ? 1 : this.bufferSize)
                 .flatMap((ServerSentEvent t1) -> {
                     lastDataReceived.set(System.currentTimeMillis());
                     if (isConnected.get() && isReceivingData.compareAndSet(false, true))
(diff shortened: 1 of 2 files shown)


crioux-stripe:
On the subject of the pushes for test values: It took me a few cycles to realize that multiple unit tests were sharing the same metrics so I needed to use a unique ID.

I believe the current publish failure is unrelated, it looks like an actor call flaked/timed out on an unrelated piece of code.

crioux-stripe:
Confirmed by committing a no-op comment change that the CI failure was a flaky unrelated test.

Andyz26:
> On the subject of the pushes for test values: It took me a few cycles to realize that multiple unit tests were sharing the same metrics so I needed to use a unique ID.
> 
> I believe the current publish failure is unrelated, it looks like an actor call flaked/timed out on an unrelated piece of code.

this is very cool! btw did you see any comparison on latency if the buffer size is set to something higher?

Andyz26 (approved):

crioux-stripe:
We didn't see a ton of latency impact but mostly because the use case is just a little bursty. We have enough capacity in any given `50ms` period but we just may not have the capacity in the `5ms` when the data comes in.

We did see some memory usage increase but the job is also handling a million RPS over three workers so that was to be expected! We also had memory to spare.

PR #724, opened 2024-11-07
Handle overloaded agent's heartbeat timeout

Andyz26:
### Context

We found a case where agents with shutdown workers got leaked:
* A job with multiple workers is being shut down.
* Some workers receive the shutdown signals and terminate the running workers. However, some workers fail to receive the signals (timeout, etc.) and continue to be alive. Now, all upstream traffic is re-distributed to these workers, and they become very busy (peaked CPU/network usage).
* Retried kill signals cannot reach these workers while the worker's host agent's heartbeat message to the control plane timed out, causing them to be leaked.

Improvements:
* increase heartbeat connection client's thread priority
* increase heartbeat connection timeout settings.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/CoreConfiguration.java
+++ b/mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/CoreConfiguration.java
@@ -67,15 +67,15 @@ public interface CoreConfiguration {
     int getAsyncHttpClientMaxConnectionsPerHost();
 
     @Config("mantis.asyncHttpClient.connectionTimeoutMs")
-    @Default("10000")
+    @Default("90000")
     int getAsyncHttpClientConnectionTimeoutMs();
 
     @Config("mantis.asyncHttpClient.requestTimeoutMs")
-    @Default("10000")
+    @Default("90000")
     int getAsyncHttpClientRequestTimeoutMs();
 
     @Config("mantis.asyncHttpClient.readTimeoutMs")
-    @Default("10000")
+    @Default("90000")
     int getAsyncHttpClientReadTimeoutMs();
 
     @Config("mantis.asyncHttpClient.followRedirect")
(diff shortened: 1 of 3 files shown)


mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/CoreConfiguration.java
```diff
@@ -67,15 +67,15 @@ public interface CoreConfiguration {
...
     int getAsyncHttpClientConnectionTimeoutMs();
 
     @Config("mantis.asyncHttpClient.requestTimeoutMs")
-    @Default("10000")
+    @Default("90000")
     int getAsyncHttpClientRequestTimeoutMs();
 
     @Config("mantis.asyncHttpClient.readTimeoutMs")
-    @Default("10000")
+    @Default("90000")
```
sundargates, line 78: should we change the defaults for everyone? Can we just change this in our codebase?

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfiguration.java
```diff
@@ -85,7 +85,7 @@ default String getTaskExecutorHostName() {
     int getTolerableConsecutiveHeartbeatFailures();
 
     @Config("mantis.taskexecutor.heartbeats.timeout.ms")
-    @Default("5000")
+    @Default("90000")
```
sundargates, line 88: same here.

sundargates: It feels strange that we send heartbeats every 10 seconds by default but wait for 90 seconds for those heartbeats to be acknowledged. I think if you want to change this, you should also change the interval.

sundargates (approved):

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfiguration.java
Andyz26: In a happy case where the request is completed right away nothing changes. I think the semantics of hb every 10 seconds is good, but we want to wait longer before aborting the request to avoid the leak here when both client and the control-plane are actually healthy (in this case, the request can be completed if the connection waits long enough).

mantis-control-plane/mantis-control-plane-core/src/main/java/io/mantisrx/server/core/CoreConfiguration.java
Andyz26: see the other comment.

PR #725, opened 2024-11-13
Resubmit Worker Allocations

kmg-stripe:
A recent change removed worker resubmits when workers are stuck in accepted: https://github.com/Netflix/mantis/pull/719

Our underlying scheduler will not retry the allocations, so we need a way to conditionally enable the ability to resubmit.

### Context

I did not see unit tests for the functionality that was removed.  I'd be happy to add them, but would like to get this merged first, since we had to pin master to a different version than the agents to avoid stuck workers.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
@@ -1906,13 +1906,29 @@ public class JobActor extends AbstractActorWithTimers implements IMantisJobManag
                 for (JobWorker worker : stage.getAllWorkers()) {
                     IMantisWorkerMetadata workerMeta = worker.getMetadata();
                     if (!workerMeta.getLastHeartbeatAt().isPresent()) {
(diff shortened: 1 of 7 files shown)


kmg-stripe:
cc: @Andyz26

Andyz26:
@kmg-stripe this lgtm. just curious what was the error/root cause you saw to have the workers stuck in accepted? For us we usually get that if the new job artifact contains bugs or failed to init so retry doesn't help in those cases (and sometimes it pollutes our agent pool e.g. error cause frequent agent crash or filled the disk etc.)

Andyz26 (approved):

kmg-stripe:
> @kmg-stripe this lgtm. just curious what was the error/root cause you saw to have the workers stuck in accepted? For us we usually get that if the new job artifact contains bugs or failed to init so retry doesn't help in those cases (and sometimes it pollutes our agent pool e.g. error cause frequent agent crash or filled the disk etc.)

@Andyz26 thanks!  yup, this was it.  one our end, this was triggered by slowness in the underlying ASG to spin-up new instances.  it is an internal limitation we will hope to fix soon, but need to tolerate it as "expected behavior" for now.

PR #728, opened 2024-12-05
Fix Incorrect Cluster Size Update in ExperimentalControlLoop When Scaling Down is Disabled

fdc-ntflx:
### Background

While monitoring the autoscaling behavior of our clusters, I observed an unexpected scale-down event in the shared MRE source ~3 AM PST, despite having scaling down disabled (new feature introduced for the next live event). This anomaly was not present in the PlayAPI source, which behaved correctly under the same conditions.

### Issue

Upon investigation, I discovered that the `ExperimentalControlLoop` class, used by the Clutch RPS scaling policy, was incorrectly updating the `currentSize` to the `desiredSize` even when the actuator did not perform a scale-down due to scaling being disabled. This led to a discrepancy between the `currentSize` and the actual cluster size, causing future scaling decisions to be based on incorrect data.

### Impact

This behavior resulted in a critical issue where, upon scaling up, the system could mistakenly reduce the cluster size. For example, with scaling down disabled:

* The currentSize was reduced from 100 to 50, while the actual size remained 100.
* When traffic increased, the scaler decided to "scale up" to 60, which effectively reduced the cluster from 100 to 60 nodes—a scale-down operation.

## Solution

The fix ensures that the currentSize is not updated when the actuator is unable to perform a scale-down due to scaling being disabled. This maintains the integrity of the cluster size data, preventing erroneous scaling operations.

## Testing 

* Added unit tests to simulate scenarios where scaling down is disabled, verifying that the currentSize remains unchanged.
* Confirmed that the system behaves correctly when scaling is enabled, with appropriate updates to the currentSize.
* TODO: apply this change to SharedMRE and test again in production

### DIFF
--- a/mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoScaler.java
+++ b/mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoScaler.java
@@ -398,13 +398,13 @@ public class JobAutoScaler {
         return desiredWorkers;
       }
 
-      public void scaleDownStage(final int numCurrentWorkers, final int desiredWorkers, final String reason) {
+      public boolean scaleDownStage(final int numCurrentWorkers, final int desiredWorkers, final String reason) {
         logger.info("scaleDownStage decrementing number of workers from {} to {}", numCurrentWorkers, desiredWorkers);
         cancelOutstandingScalingRequest();
         final StageScalingPolicy scalingPolicy = stageSchedulingInfo.getScalingPolicy();
         if (scalingPolicy != null && scalingPolicy.isAllowAutoScaleManager() && !jobAutoscalerManager.isScaleDownEnabled()) {
             logger.warn("Scaledown is disabled for all autoscaling strategy. For stage {} of job {}", stage, jobId);
-            return;
+            return false;
         }
         final Subscription subscription = masterClientApi.scaleJobStage(jobId, stage, desiredWorkers, reason)
           .retryWhen(retryLogic)
@@ -414,6 +414,7 @@ public class JobAutoScaler {
           })
         .subscribe();
         setOutstandingScalingRequest(subscription);
+        return true;
       }
 
       public int getStage() {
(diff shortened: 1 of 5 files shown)


Andyz26 (approved):
Nice catch!

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoScaler.java
```diff
@@ -398,13 +398,13 @@ public int getDesiredWorkersForScaleDown(final int decrement, final int numCurre
...
+      public boolean scaleDownStage(final int numCurrentWorkers, final int desiredWorkers, final String reason) {
         logger.info("scaleDownStage decrementing number of workers from {} to {}", numCurrentWorkers, desiredWorkers);
         cancelOutstandingScalingRequest();
         final StageScalingPolicy scalingPolicy = stageSchedulingInfo.getScalingPolicy();
         if (scalingPolicy != null && scalingPolicy.isAllowAutoScaleManager() && !jobAutoscalerManager.isScaleDownEnabled()) {
             logger.warn("Scaledown is disabled for all autoscaling strategy. For stage {} of job {}", stage, jobId);
-            return;
+            return false;
         }
         final Subscription subscription = masterClientApi.scaleJobStage(jobId, stage, desiredWorkers, reason)
```
sundargates, line 409: why is this an observable? it feels like there's no need for a stream here?

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/control/actuators/ClutchMantisStageActuator.java
```diff
@@ -38,7 +38,9 @@ protected Double processStep(Tuple3<String, Double, Integer> tup) {
 
         String reason = tup._1;
         if (desiredNumWorkers < tup._3) {
-            scaler.scaleDownStage(tup._3, desiredNumWorkers, reason);
+            if (!scaler.scaleDownStage(tup._3, desiredNumWorkers, reason)) {
+                return tup._3 * 1.0;
```
sundargates, line 42: can you add a comment on what this is doing?

sundargates (approved):

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoScaler.java
fdc-ntflx: Good point. I didn't check the full logic, so I don't really know.

PR #729, opened 2024-12-05
Fix sse connection infinite retry failure

Andyz26:
### Context

We noticed that when the SSE connection triggered "doOnError", the SSE HTTP client was closed internally (via resetConn) but the following "retryWhen" operation started an infinite retry action on the closed HTTP Client (and every retry failed right away since the client is not going to start the actual connection).

Solution:
Refactor to split reset and close operations on the SSE client so that the "doOnErro" within the retry loop will only reset instead of closing the client. Added an outer "doOnError" to make sure the client gets closed once the retry loop is broken.

### Checklist

### DIFF
--- a/mantis-server/mantis-server-worker-client/src/main/java/io/mantisrx/server/worker/client/SseWorkerConnection.java
+++ b/mantis-server/mantis-server-worker-client/src/main/java/io/mantisrx/server/worker/client/SseWorkerConnection.java
@@ -98,11 +98,11 @@ public class SseWorkerConnection {
                                 @Override
                                 public Observable<?> call(Integer integer) {
                                     if (isShutdown) {
-                                        logger.info(getName() + ": Is shutdown, stopping retries");
+                                        logger.info("{}: Is shutdown, stopping retries", getName());
                                         return Observable.empty();
                                     }
                                     long delay = 2 * (integer > 10 ? 10 : integer);
-                                    logger.info(getName() + ": retrying conx after sleeping for " + delay + " secs");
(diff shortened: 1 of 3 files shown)


fdc-ntflx (approved):

mantis-server/mantis-server-worker-client/src/main/java/io/mantisrx/server/worker/client/SseWorkerConnection.java
```diff
@@ -249,18 +249,34 @@ public synchronized Observable<MantisServerSentEvent> call() {
...
                             resetConnected();
-                            logger.warn(getName() +
-                                    "Error on getting response from SSE server: " + throwable.getMessage());
+                            logger.warn("{}: Error on getting response from SSE server: {}",
+                                getName(), throwable.getMessage());
                             connectionResetHandler.call(throwable);
                         })
                         .retryWhen(retryLogic)
-                        .doOnCompleted(this::resetConnected);
+                        .doOnError((Throwable throwable) -> {
```
calvin681, line 260: Having double `doOnError` (here and on line 251) seems wrong to me.

calvin681: I think both will still be called, regardless that there is a `retryWhen` in between.

Andyz26: let me try test this. My understanding is the retryWhen will swallow the error so outer doOnError won't see it if it's being retried

calvin681 (approved):

PR #730, opened 2024-12-10
re-enable disabled node on expiry

Andyz26:
### Context

If a node marked as disabled was never destroyed by the host cluster and maintained heartbeats (thus, its executor state remained valid in memory), it would remain disabled until the next leader's re-election.

Fix:
On disable node expiration, ensure the internal state (if still valid) is reset to be not disabled.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
@@ -518,6 +518,13 @@ class ResourceClusterActor extends AbstractActorWithTimers {
             if (activeDisableTaskExecutorsByAttributesRequests.remove(request.getRequest()) || (request.getRequest().getTaskExecutorID().isPresent() && disabledTaskExecutors.remove(request.getRequest().getTaskExecutorID().get()))) {
                 mantisJobStore.deleteExpiredDisableTaskExecutorsRequest(request.getRequest());
             }
+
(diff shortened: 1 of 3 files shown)


markcho (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -518,6 +518,13 @@ private void onDisableTaskExecutorsRequestExpiry(ExpireDisableTaskExecutorsReque
             if (activeDisableTaskExecutorsByAttributesRequests.remove(request.getRequest()) || (request.getRequest().getTaskExecutorID().isPresent() && disabledTaskExecutors.remove(request.getRequest().getTaskExecutorID().get()))) {
                 mantisJobStore.deleteExpiredDisableTaskExecutorsRequest(request.getRequest());
             }
+
+            // also re-enable the node if the state is still valid.
```
james-lubin, line 522: Should we add logging here that we're re-enabling the node?

james-lubin (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
Andyz26: there is a log above so I figure it's covered.

fdc-ntflx (approved):
LGTM!

PR #731, opened 2024-12-13
Ensure worker failed before first heartbeat gets resubmission

Andyz26:
### Context

Problem:
The missing heartbeat logic relies on the last HB timestamp to determine whether it should be re-submitted by the job actor or the scheduler. This creates a problem for workers failed right after getting scheduled but before it's able to emit the first heartbeat message. This problem gets amplified when we have higher transient worker startup failures due to other issues.

Solution:
Instead of last heartbeat, use the launched event status to decide whether the job actor should start taking over the job resubmission from the scheduler.

### Checklist


fdc-ntflx (approved):
LGTM!

sundargates (approved):
LGTM!

PR #732, opened 2024-12-16
report error creating cron

dtrager02:
### Context

Currently, initializing a job cluster with an incorrect cron schedule that cannot be parsed by Quartz silently errors with a warning log. We should return a response instead

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
@@ -795,6 +795,7 @@ public class JobClusterActor extends AbstractActorWithTimers implements IJobClus
                     initReq.requestId, e instanceof SchedulerException?CLIENT_ERROR:SERVER_ERROR,
                     "Job Cluster " + jobClusterName + " could not be created due to cron initialization error" + e.getMessage(),
                     jobClusterName), getSelf());
+                return;
             }
             initRunningJobs(initReq, sender);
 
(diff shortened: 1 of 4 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -790,6 +792,10 @@ public void onJobClusterInitialize(JobClusterProto.InitializeJobClusterRequest i
                 cronManager = new CronManager(name, getSelf(), jobClusterMetadata.getJobClusterDefinition().getSLA());
             } catch (Exception e) {
                 logger.warn("Exception initializing cron", e);
+                getSender().tell(new JobClusterManagerProto.CreateJobClusterResponse(
```
Andyz26, line 795: maybe separate this catch to known error (SchedulerException) and other error (exception) and use different error code + message.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -790,6 +791,10 @@ public void onJobClusterInitialize(JobClusterProto.InitializeJobClusterRequest i
...
+                    initReq.requestId, e instanceof SchedulerException?CLIENT_ERROR:SERVER_ERROR,
+                    "Job Cluster " + jobClusterName + " could not be created due to cron initialization error" + e.getMessage(),
+                    jobClusterName), getSelf());
```
Andyz26, line 797: in this case don't you want to stop here and not initRunning workers?

Andyz26 (approved):

PR #733, opened 2024-12-17
Fix scheduler double booking task executors

Andyz26:
### Context

Problem:
When a TE is returned from here to be used for scheduling, its state remain active until the scheduler trigger another message to update (lock) the state. However when large number of the requests are active at the same time on same sku, the gap between here and the message to lock the state can be large so another schedule request message can be in between and got the same set of TEs. 

Solution:
To avoid this, a lease is added to each TE state (based on last scheduled usage timestamp) to temporarily lock the TE to be used again. Since this is only lock between actor messages and lease duration can be short.

* Also fixed a NPE from the node state update call.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorState.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorState.java
@@ -288,7 +288,8 @@ class TaskExecutorState {
         return Duration.between(this.lastSchedulerLeased, this.clock.instant());
     }
 
-    void updateLastSchedulerLeased() {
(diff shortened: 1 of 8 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/TaskExecutorState.java
```diff
@@ -278,6 +283,15 @@ Instant getLastActivity() {
         return this.lastActivity;
     }
 
+    Duration getLastSchedulerLeasedDuration()
```
markcho, line 286: Nit: Format
```suggestion
    Duration getLastSchedulerLeasedDuration() {
```

markcho (approved):

PR #734, opened 2024-12-18
Fix worker no heartbeat race condition

Andyz26:
### Context

A race condition was introduced from #731 when a heartbeat check happened between launch event and first heartbeat and can trigger NPE on the actor. 
Refactored the condition check on the case above + improve UTs to cover the related states.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
@@ -1943,18 +1943,18 @@ public class JobActor extends AbstractActorWithTimers implements IMantisJobManag
                                 acceptedAt);
                         }
                     } else {
+                        // no heartbeat or heartbeat too old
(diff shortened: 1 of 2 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
```diff
@@ -1943,15 +1943,25 @@ public void checkHeartBeats(Instant currentTime) {
                                 acceptedAt);
                         }
                     } else {
-                        if (Duration.between(workerMeta.getLastHeartbeatAt().get(), currentTime).getSeconds()
+                        if (!workerMeta.getLastHeartbeatAt().isPresent() || Duration.between(workerMeta.getLastHeartbeatAt().get(), currentTime).getSeconds()
                             > missedHeartBeatToleranceSecs) {
-                            // heartbeat too old
+                            // no heartbeat or heartbeat too old
```
sundargates, line 1948: add this above line 1946

sundargates (approved):
lgtm

PR #737, opened 2024-12-23
Use a singleton for DynamoDbMasterMonitor

timmartin-stripe:
### Context

We encountered an issue where the MasterMonitor's `shutdown` method was called somewhere.  This led to clients relying on updates to the MasterMonitor stopping receiving updates but they had no mechanism of knowing that the monitor they were subscribed to was no longer updating because the observable was not completed and getting the latest version would continue to return the last known value (even though that value was not updated).

We were deciding between either using a Singleton or tracking down `shutdown` calls to `MasterMonitor` and ensuring that usages could resubscribe if necessary.  The latter seemed like it would be prone to breakage in the future (e.g. particularly with the master observable). Additionally, a Singleton pattern seemed fitting given that almost every reasonable use case of Mantis relies on a sustained connection to the Mantis Master.  So there's no reason to expect that shutting down the executor pool for updating the master is necessary.  Additionally, using a Singleton avoids any unbounded resource leakage (e.g. it is still safe to create N MasterMonitors).

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-dynamodb/src/main/java/io/mantisrx/extensions/dynamodb/DynamoDBMasterMonitor.java
+++ b/mantis-control-plane/mantis-control-plane-dynamodb/src/main/java/io/mantisrx/extensions/dynamodb/DynamoDBMasterMonitor.java
@@ -15,180 +15,33 @@
  */
 package io.mantisrx.extensions.dynamodb;
 
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDBLockClient;
-import com.amazonaws.services.dynamodbv2.LockItem;
-import io.mantisrx.common.metrics.Counter;
-import io.mantisrx.common.metrics.Metrics;
-import io.mantisrx.common.metrics.MetricsRegistry;
 import io.mantisrx.server.core.BaseService;
-import io.mantisrx.server.core.json.DefaultObjectMapper;
 import io.mantisrx.server.core.master.MasterDescription;
 import io.mantisrx.server.core.master.MasterMonitor;
-import io.mantisrx.shaded.com.fasterxml.jackson.databind.ObjectMapper;
-import java.io.IOException;
-import java.nio.ByteBuffer;
-import java.time.Duration;
-import java.util.Optional;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.TimeUnit;
-import javax.annotation.Nullable;
 import lombok.extern.slf4j.Slf4j;
-import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


timmartin-stripe:
cc @crioux-stripe

mantis-control-plane/mantis-control-plane-dynamodb/src/main/java/io/mantisrx/extensions/dynamodb/DynamoDBMasterMonitor.java
```diff
@@ -15,180 +15,33 @@
...
-    private final Counter noLockPresentCounter;
-    private final Counter lockDecodeFailedCounter;
-    private final Counter nullNextLeaderCounter;
+    private final DynamoDBMasterMonitorSingleton singleton;
 
     /**
      * Creates a MasterMonitor backed by DynamoDB. This should be used if you are using a {@link DynamoDBLeaderElector}
      */
     public DynamoDBMasterMonitor() {
-        this(DynamoDBClientSingleton.getLockClient(),
```
timmartin-stripe, line 84: Most of this logic was moved into `DynamoDBMasterMonitorSingleton`

mantis-control-plane/mantis-control-plane-dynamodb/src/test/java/io/mantisrx/extensions/dynamodb/DynamoDBMasterMonitorSingletonTest.java
```diff
@@ -133,13 +138,13 @@ public void monitorDoesNotReturnNull() throws IOException, InterruptedException
...
-            .untilAsserted(() -> assertEquals(MasterDescription.MASTER_NULL, m.getLatestMaster()));
+            .untilAsserted(() -> assertEquals(MasterDescription.MASTER_NULL, m.getMasterSubject().getValue()));
         lockSupport.releaseLock(lockKey);
 
         m.shutdown();
 
-        testSubscriber.assertNoTerminalEvent();
-        testSubscriber.assertNotCompleted();
+        testSubscriber.awaitTerminalEvent(10, TimeUnit.SECONDS);
+        testSubscriber.assertCompleted();
```
timmartin-stripe, line 147: We do want the observable to complete so that subscribers can do any necessary clean up.

crioux-stripe: That tracks.

crioux-stripe (approved):
Thanks for looking into this @timmartin-stripe !

I'm inclined to get this merged quickly and cut an RC so we can deploy tomorrow. I believe we're the only ones using this so the risk is our own.

PR #740, opened 2025-01-07
Add metrics to SingleThreadedChunker

timmartin-stripe:
### Context

Using a similar set of metrics to what the `TimedChunker` uses.  It's helpful for diagnosing queue depth issues.  When we switched to the SPSC queue, we were surprised the metrics were missing so I figured we could add them in.

### Checklist

### DIFF
--- a/mantis-network/src/main/java/io/reactivex/mantis/network/push/SingleThreadedChunker.java
+++ b/mantis-network/src/main/java/io/reactivex/mantis/network/push/SingleThreadedChunker.java
@@ -16,6 +16,10 @@
 
 package io.reactivex.mantis.network.push;
 
+import io.mantisrx.common.metrics.Counter;
+import io.mantisrx.common.metrics.Metrics;
+import io.mantisrx.common.metrics.MetricsRegistry;
+import io.mantisrx.common.metrics.spectator.MetricGroupId;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


mantis-network/src/main/java/io/reactivex/mantis/network/push/SingleThreadedChunker.java
```diff
@@ -41,6 +50,17 @@ public SingleThreadedChunker(ChunkProcessor<T> processor, MonitoredQueue<T> iQ,
         this.connectionManager = connMgr;
         chunk = new Object[this.chunkSize];
 
+        MetricGroupId metricsGroup = new MetricGroupId("SingleThreadedChunker");
```
Andyz26, line 53: maybe add an index from ctor to help identify between threads.

mantis-network/src/main/java/io/reactivex/mantis/network/push/SingleThreadedChunker.java
```diff
@@ -93,6 +115,7 @@ private void drain() {
             }
 
             processor.process(connectionManager, copy);
```
Andyz26, line 117: i wonder if the "copy" list can be converted to be an immutable list.

Andyz26 (approved):

PR #741, opened 2025-01-09
Handle Case Where Recently Launched Worker Does Not Immediately Heartbeat

kmg-stripe:
It seems like there is a race condition where a recently launched worker has not sent a heartbeat, the duration is still within the missed heartbeat threshold and the JobActor treats the lack of heartbeat with a resubmit.

### Context

### Checklist

### DIFF
--- a/.github/workflows/nebula-ci.yml
+++ b/.github/workflows/nebula-ci.yml
@@ -48,7 +48,7 @@ jobs:
           CI_BRANCH: ${{ github.ref }}
           COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Upload Test Results
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         if: always()
         with:
           name: Unit Test Results
@@ -59,7 +59,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Upload
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: Event File
           path: ${{ github.event_path }}
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
@@ -1943,9 +1943,12 @@ public class JobActor extends AbstractActorWithTimers implements IMantisJobManag
                                 acceptedAt);
                         }
                     } else {
-                        // no heartbeat or heartbeat too old
-                        if (!workerMeta.getLastHeartbeatAt().isPresent() || Duration.between(workerMeta.getLastHeartbeatAt().get(), currentTime).getSeconds()
-                            > missedHeartBeatToleranceSecs) {
+                        // no heartbeat in a timely manner since launched or heartbeat too old
+                        boolean noTimelyHeartbeatSinceLaunched = !workerMeta.getLastHeartbeatAt().isPresent()
+                            && Duration.between(Instant.ofEpochSecond(workerMeta.getLaunchedAt()), currentTime).getSeconds() > missedHeartBeatToleranceSecs;
+                        boolean heartbeatTooOld = workerMeta.getLastHeartbeatAt().isPresent()
+                            && Duration.between(workerMeta.getLastHeartbeatAt().get(), currentTime).getSeconds() > missedHeartBeatToleranceSecs;
+                        if (noTimelyHeartbeatSinceLaunched || heartbeatTooOld) {
                             this.numWorkerMissingHeartbeat.increment();
 
                             if (!workerMeta.getLastHeartbeatAt().isPresent()) {
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
@@ -1944,6 +1944,7 @@ public class JobActor extends AbstractActorWithTimers implements IMantisJobManag
                         }
                     } else {
                         // no heartbeat in a timely manner since launched or heartbeat too old
+                        // note: the worker has been launched
                         boolean noTimelyHeartbeatSinceLaunched = !workerMeta.getLastHeartbeatAt().isPresent()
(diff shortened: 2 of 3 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
```diff
@@ -1943,9 +1943,12 @@ public void checkHeartBeats(Instant currentTime) {
                                 acceptedAt);
                         }
                     } else {
-                        // no heartbeat or heartbeat too old
-                        if (!workerMeta.getLastHeartbeatAt().isPresent() || Duration.between(workerMeta.getLastHeartbeatAt().get(), currentTime).getSeconds()
-                            > missedHeartBeatToleranceSecs) {
+                        // no heartbeat in a timely manner since launched or heartbeat too old
+                        boolean noTimelyHeartbeatSinceLaunched = !workerMeta.getLastHeartbeatAt().isPresent()
+                            && Duration.between(Instant.ofEpochSecond(workerMeta.getLaunchedAt()), currentTime).getSeconds() > missedHeartBeatToleranceSecs;
```
kmg-stripe, line 1948: ah wait, using launched at might not work here...  i'll have to figure something else out.

kmg-stripe:
Hey @Andyz26 !  When releasing the latest version of Mantis in our QA environment, we started noticing excessive resubmits.  The problem goes away when we rollback.  I "think" the issue could be related to the fix in #734 .

It looks like a recently "launched" worker that has not reported a heartbeat could be resubmitted by the master.  I am wondering if we want to check the duration `now() - launchedAt` exceeds the heartbeat threshold before resubmitting.

If this seems reasonable, I can create another test (or add to the existing one).  If not, can you think of a reason we would see excessive resubmits.

cc: @crioux-stripe

.github/workflows/nebula-ci.yml
```diff
@@ -48,7 +48,7 @@ jobs:
           CI_BRANCH: ${{ github.ref }}
           COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Upload Test Results
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
```
kmg-stripe, line 51: this is unrelated, but was breaking CI because GH deprecated upload-artifact@v3

mantis-control-plane/mantis-control-plane-server/src/test/java/io/mantisrx/master/jobcluster/job/JobTestLifecycle.java
```diff
@@ -823,9 +823,9 @@ public void testNoHeartBeatAfterLaunchResubmit() {
             assertEquals(JobState.Accepted, resp4.getJobMetadata().get().getState());
 
             // 1 original submissions and 0 resubmits because of worker not in launched state with HB timeouts
-            verify(schedulerMock, times(2)).scheduleWorkers(any());
+            verify(schedulerMock, times(1)).scheduleWorkers(any());
```
kmg-stripe, line 826: just temporary to get the tests passing.  i'll update the tests if we think the above logic is reasonable.

kmg-stripe:
@Andyz26 - if you get a chance, could you approve the snapshot build?  i was hoping to see if this fixes the issue we are seeing on our end.  thanks!

Andyz26:
@kmg-stripe i think i understand the problem now and i think your solution in pr makes sense to me. 
Also I would like to use this chance to understand better how you are dealing with the lifecycle events. In our setup the launch event is emitted when the worker is basically "ready to work" which is probably why we don't see this issue.

kmg-stripe:
> @kmg-stripe i think i understand the problem now and i think your solution in pr makes sense to me. Also I would like to use this chance to understand better how you are dealing with the lifecycle events. In our setup the launch event is emitted when the worker is basically "ready to work" which is probably why we don't see this issue.
> 
> It would be great if we can consolidate on the lifecycle events expectations for future changes.

@Andyz26 I believe we are not doing anything special regarding worker lifecycle.  That is, I think we are relying on whatever the "default" behavior is.  My impression is that "launched" means a worker has landed on an instance, but hasn't started yet.  I'd say the time between "launched" and "started" can be from 10s of seconds up to a few minutes.  @crioux-stripe can keep me honest here, if we are doing anything out of the ordinary.

Also - looks like gradle failed when [publishing a snapshot](https://github.com/Netflix/mantis/actions/runs/12713553187/job/35441706645).  I cannot tell why based on the error:

```
Caused by: org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 1; Content is not allowed in prolog.
```
Not sure if this is a transient issue...  Could you approve the latest snapshot job when you get a chance?  Thanks!

crioux-stripe:
Approving to see if I can get a snapshot triggered.

crioux-stripe (approved):

crioux-stripe:
Tracing backwards our workers are getting set to launched in the [JobWorker#processEvent(WorkerEvent)](https://github.com/Netflix/mantis/blob/master/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/worker/JobWorker.java#L175-L177) method as expected. I'll dig into what actually triggers that event...

The `WorkerLaunched` event is created in our [ResourceClusterAwareSheculderActor#onSubmittedScheduleRequstEvent(SubmittedScheduleRequestEvent)](https://github.com/Netflix/mantis/blob/master/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java#L292-L299). I'm guessing the preceding line `final TaskExecutorRegistration info = resourceCluster.getTaskExecutorInfo(taskExecutorID).join();` is much more sophisticated in the Netflix Titus implementation. as compared to our thin ASG wrapper. Created [here](https://github.com/Netflix/mantis/blob/master/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/server/master/scheduler/ResourceClusterAwareSchedulerActor.java#L224).

Andyz26:
Maybe some operations during the worker startup was taking longer? I think we have a shorter window from submitted/launched to first HB sent thus this is not causing issues for us. I think the logic here is sound and we can merge and get a RC started.

crioux-stripe:
Hrmm. Not really sure what is going on with this snapshot. I'm of the opinion we can probably merge this as is, but want to wait on @Andyz26 's approval.

On the subject of the lifecycle, @Andyz26 we're just relying on the default behavior I linked to above. Does the resource provider used internally at Netflix do something smarter with the agent states?

crioux-stripe:
Ooops! Missed @Andyz26 's intervening comment. Will merge!

PR #746, opened 2025-01-29
DynamoDBMasterMonitor: Add option to disable shutdown hook.

crioux-stripe:
### Context

The `DynamoDBMasterMonitor` presently ensures a clean shutdown when the JVM shutdown hook is called. This works well for Master instances as it will release the lock on the leader election. However in when running on the agent this has the downside of sometimes shutting down before other threads that require master communication. The result is workers that stay stuck until a `180s` heartbeat timeout. Furthermore closing the lock client has no benefit for the agents -- they don't lock it.

This would enable us to configure agents to let this thread keep running. It is configured as a daemon thread so it will not prevent the JVM from shutting down.

### Checklist


kmg-stripe (approved):
lgtm!

crioux-stripe:
I believe we're the only known users of the DynamoDB components so I'm going to merge this in when CI finishes.

PR #747, opened 2025-02-04
Add availabilityZone to AsyncConnection

tcai-stripe:
### Context

Add an `availabilityZone` field to `AsyncConnection` which can then later be accessed in the routers for in-az request routing 

### Checklist

### DIFF
--- a/mantis-network/src/main/java/io/reactivex/mantis/network/push/AsyncConnection.java
+++ b/mantis-network/src/main/java/io/reactivex/mantis/network/push/AsyncConnection.java
@@ -42,6 +42,13 @@ public class AsyncConnection<T> {
     public AsyncConnection(String host, int port, String id,
                            String slotId,
                            String groupId, Observer<List<byte[]>> subject,
+                           Func1<T, Boolean> predicate) {
+        this(host, port, id, slotId, groupId, subject, predicate, null);
+    }
+
+    public AsyncConnection(String host, int port, String id,
+                           String slotId,
+                           String groupId, Observer<List<byte[]>> subject,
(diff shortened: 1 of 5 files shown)


crioux-stripe:
Hey @Andyz26 

Context here is that we're trying to insert a custom router into our endpoint communications to allow us to bias connections to the same AZ. @tcai-stripe has come up with a pretty clean mechanism for the Sink that just involves constructing it with the AZ aware router. This pull request is a first approach to potentially getting this information in the stage to stage communication. **All of this is in the service of reducing network costs for Mantis.**

What are your thoughts on the best way to customize the construction / injection of custom routers? I think this sort of brings us back to the discussion we've been kicking down the road of how to allow for the injection of custom classes in the open source codebase.

Thoughts?

mantis-network/src/main/java/io/reactivex/mantis/network/push/AsyncConnection.java
```diff
@@ -34,20 +34,23 @@ public class AsyncConnection<T> {
     private String slotId;
     private String id;
 
+    private String availabilityZone;
+
     private Observer<List<byte[]>> subject;
     private Func1<T, Boolean> predicate;
 
     public AsyncConnection(String host, int port, String id,
```
crioux-stripe, line 42: @tcai-stripe One change I'd like to see here is having the original constructor call this one with a null for `availabilityZone`. That'll ensure that we don't break any code in Netflix's internal wrapper. Otherwise this is just adding a getter/setter which makes this pretty easy for us to sneak in.

crioux-stripe (changes requested):

crioux-stripe (approved):

PR #749, opened 2025-02-12
Add withRouter to SSESink

tcai-stripe:
### Context

Add a `withRouter` to `ServerSentEventsSink` builder to allow users to plug in a customized router instead of using the default round robin one
example use case: route requests to the same availability zone

cc @crioux-stripe 

### Checklist


crioux-stripe (approved):
Also approving this one as it is strictly additive. This allows us to cut an RC for our internal project for minimizing cross-az traffic.

PR #751, opened 2025-02-18
Make executor state manager pluggable

andresgalindo-stripe:
### Context

Making the ExecutorStateManager pluggable so we can implement some custom scheduling logic 

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManager.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManager.java
@@ -34,7 +34,7 @@ import javax.annotation.Nullable;
 /**
  * A component to manage the states of {@link TaskExecutorState} for a given {@link ResourceClusterActor}.
  */
-interface ExecutorStateManager {
+public interface ExecutorStateManager {
     /**
      * Store and track the given task executor's state inside this {@link ExecutorStateManager} if there is no existing
      * state already. Ignore the given state instance if there is already a state associated with the given ID.
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
@@ -154,7 +154,8 @@ class ResourceClusterActor extends AbstractActorWithTimers {
         String jobClustersWithArtifactCachingEnabled,
         boolean isJobArtifactCachingEnabled,
         Map<String, String> schedulingAttributes,
-        FitnessCalculator fitnessCalculator
+        FitnessCalculator fitnessCalculator,
+        Optional<ExecutorStateManager> executorStateManager
     ) {
         return Props.create(
             ResourceClusterActor.class,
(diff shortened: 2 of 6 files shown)


Andyz26:
@andresgalindo-stripe could you share what's the logic need? Since we plan to work on some improvements around this part too later this year I would like to understand better how we might be able to consolidate or decouple the logic here.

crioux-stripe:
I think we definitely want to wait for @Andyz26 's take on this one.

On the subject of using [Optional](https://stackoverflow.com/questions/23454952/uses-for-optional) as a parameter there are a few schools of thought. I'm not sure I have a strong or well developed opinion here but I'm curious if anyone from the Netflix side has a strong opinion about wether we should pass `Optional` or `null`?

Andyz26:
> I think we definitely want to wait for @Andyz26 's take on this one.
> 
> On the subject of using [Optional](https://stackoverflow.com/questions/23454952/uses-for-optional) as a parameter there are a few schools of thought. I'm not sure I have a strong or well developed opinion here but I'm curious if anyone from the Netflix side has a strong opinion about wether we should pass `Optional` or `null`?

In general i would avoid using optional as argument too.
As for the change plz see my other comment.

andresgalindo-stripe:
> @andresgalindo-stripe could you share what's the logic need? Since we plan to work on some improvements around this part too later this year I would like to understand better how we might be able to consolidate or decouple the logic here.

Hey @Andyz26, the prime target at the moment is to balance worker assignments across availability zones so we can reduce cross-az traffic from things talking to jobs. Looking through what we could do current scheduling constraints / fitness, I wasn't able to think of a good solution so I was planning to override the current ExecutorStateManager and fit in the AZ balancing myself.

crioux-stripe:
@Andyz26 We're doing a fair amount of work to reduce crozz-AZ traffic, for cost reasons. A few of the latest Source/Sink changes in OSS are to enable us to inject some AZ-affinity components. If that is of interest to Netflix we can probably have @tcai-stripe open source the `Router` implementation for Source/Sink.

Andyz26:
Thanks @andresgalindo-stripe @crioux-stripe for the context! A few thoughts here:
* Yeah i think we can certainly utilize the cross-az traffic reduction and it would be great to make that part of the default OSS support if possible.
* @andresgalindo-stripe for the az-aware scheduling, would it be possible to add the metadata on agent registration metadata then update the existing fitnessCalculator or plugin a custom fitnessCalculator (this is already supported and we do the same for some internal scheduling override)? any gap here?

andresgalindo-stripe:
When I reviewed the fitness calculator the gaps were:
* Machine Definitions don't have availability zone - but this seems fixable without a huge lift (maybe add some metadata map to MachineDefinition?) 
* We need to know what the other task executors are running so we can try and balance for any given job accordingly

Andyz26:
@andresgalindo-stripe thanks and I think I understand the case now. 
My main concern here is that the executorStateManager is quite big/complex and a full fork could cause more trouble for future changes and unexpected regressions (unfortunately i think the state-leaking and implicit-coupling is kind of bad on this interface). 
Do you think it would be possible to refactor the executor manager interface and impl to extract/abstract the interaction you need and the existing state instead of a full fork?

andresgalindo-stripe:
> @andresgalindo-stripe thanks and I think I understand the case now. My main concern here is that the executorStateManager is quite big/complex and a full fork could cause more trouble for future changes and unexpected regressions (unfortunately i think the state-leaking and implicit-coupling is kind of bad on this interface). Do you think it would be possible to refactor the executor manager interface and impl to extract/abstract the interaction you need and the existing state instead of a full fork? This is something could benefit future work and collab too i think.

Yep! I'll close this PR and give it another approach

PR #753, opened 2025-02-19
Scheduling hook into ExecuteStateManagerImpl

andresgalindo-stripe:
### Context

Allow us to plug a hook that mutates the list of available task executors 

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/test/java/io/mantisrx/master/resourcecluster/ResourceClusterActorTest.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/test/java/io/mantisrx/master/resourcecluster/ResourceClusterActorTest.java
@@ -211,7 +211,8 @@ public class ResourceClusterActorTest {
                 "",
                 false,
                 ImmutableMap.of(),
-                new CpuWeightedFitnessCalculator());
+                new CpuWeightedFitnessCalculator(),
+                null);
 
         resourceClusterActor = actorSystem.actorOf(props);
         resourceCluster =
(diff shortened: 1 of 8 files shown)


mantis-control-plane/mantis-control-plane-server/src/test/java/io/mantisrx/master/resourcecluster/ResourceClusterActorTest.java
```diff
@@ -211,7 +211,8 @@ public void setupActor() {
                 "",
                 false,
                 ImmutableMap.of(),
-                new CpuWeightedFitnessCalculator());
+                new CpuWeightedFitnessCalculator(),
```
crioux-stripe, line 214: nit: I generally prefer the overloaded constructor so that the tests / clients who do not use the change do not need to change in any way to ignore new functionality. Especially true since the Netflix injector will likely break without having an appropriate `Bean`.

crioux-stripe (changes requested):
Generally looks good, besides the (blocking) nit I left one more item to think about... What is the intention if the user wants multiple mutator hooks? Right now we could do a composite hook and that is probably fine.

crioux-stripe (approved):

andresgalindo-stripe:
@crioux-stripe I've gone ahead and overloaded `.props()` to keep one that matches the signature of what was there before

w.r.t multiple hooks, yep, my thought that it was more flexible for the user if they implement a ChainedHook themselves and let them define how they want to cycle through hooks themselves (e.g. if one hook returns no instances do we just return all instances from the previous hook in stack in an attempt to schedule no matter what, or, do we really return nothing / throw an exception to stop this from being scheduled, etc)

Andyz26 (approved):

PR #755, opened 2025-02-25
Add new configs to support inter AZ stage-to-stage routing

tcai-stripe:
### Context

We want to reduce cross AZ traffic during scalar stage to stage communication so adding the following two keys to ServiceLocator so we can plug in a custom InterAZRouter: 

- `RouterFactory.class`: which returns a router factory class that implements `scalarStageToStageRouter` to support custom stage to stage router. It will fall back to the default round robin router if not provided
- `AvailabilityZoneUtils.class`: which returns a utils that implements `getAvailabilityZone` to allow users to provide availability zone for each worker. Default to empty string if not provided

### Checklist


crioux-stripe (approved):
This looks pretty straightforward and as though we could use it to reduce costs stage to stage. I wanted @Andyz26 to take a look as well considering the impact it could have on Netflix.

Though I like that we're defaulting to the old behavior so the base effect is no change for Netflix.

tcai-stripe:
@Andyz26 could you please help review this when you get a chance? As @codyrioux mentioned earlier that this shouldn't have any impact on Netflix and we want to plugin our own router to reduce cross AZ traffic

kmg-stripe (approved):
lgtm!  we definitely need input from Netflix to make sure this doesn't have weird interactions on their end.

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfigurationUtils.java
```diff
@@ -29,6 +30,7 @@ public static <T extends WorkerConfiguration> T frmProperties(Properties propert
             properties);
         configurationObjectFactory.addCoercible(new MetricsCoercer(properties));
         configurationObjectFactory.addCoercible(new PluginCoercible<>(MetricsCollector.class, properties));
+        configurationObjectFactory.addCoercible(new PluginCoercible<>(AvailabilityZoneUtils.class, properties));
```
Andyz26, line 33: can we avoid using the class instance injection here? We've been trying to remove these class instance injection from the the config and keep worker config as a simple pojo and eventually move to use a new config framework.

mantis-runtime/build.gradle
```diff
@@ -23,6 +23,7 @@ dependencies {
     api project(':mantis-remote-observable')
     api project(':mantis-network')
     api project(':mantis-common')
+    api project(":mantis-runtime-loader")
```
Andyz26, line 26: The loader is setup as a clean proxy to isolate runtime and agent executor so it's best not to introduce dep between loader and runtime. Can you remove this if you don't use the worker config injection?

mantis-runtime-loader/src/main/java/io/mantisrx/runtime/loader/config/WorkerConfiguration.java
```diff
@@ -192,4 +193,16 @@ default Map<String, String> getTaskExecutorAttributes() {
...
+    //  Routing related configurations
+    // ------------------------------------------------------------------------
+    @Config("mantis.taskexecutor.router.scalar-stage-to-stage")
+    @Default("io.reactivex.mantis.network.push.RoundRobinRouter")
+    String getScalarStageToStageRouterClassName();
+
+
+    @Config("mantis.availabilityZoneUtils.class")
+    @Default("io.mantisrx.common.util.DefaultAvailabilityZoneUtils")
+    AvailabilityZoneUtils getAvailabilityZoneUtils();
```
Andyz26, line 207: please avoid non-primitive types in the worker config as they've been causing many trouble for us in integration and making migration to other config framework harder.
I would recommend using either ctor injection, serviceLocator or reconstruct using primitive types here.

Andyz26 (changes requested):
overall lgtm. Only change requested here is regarding the config injection and avoiding dependency between loader and runtime.

mantis-runtime/build.gradle
tcai-stripe: @Andyz26 I used service locator for injection instead and removed the dependency on loader, let me know if this works for you

Andyz26 (approved):
Overall lgtm!
Can you also provide a doc or enhance the PR description about how to leverage this feature as well? e.g. what would be required to impl in the internal wrappers etc? Do you plan to contribute back the router side of work too?

tcai-stripe:
> Overall lgtm! Can you also provide a doc or enhance the PR description about how to leverage this feature as well? e.g. what would be required to impl in the internal wrappers etc? Do you plan to contribute back the router side of work too?

I've updated PR description. As for the router work, we plan to contribute it back if it aligns with Netflix's interests.

Andyz26:
Feel free to merge too!

PR #757, opened 2025-03-10
Fix use after release concurrency bug in SseWorkerConnection

jpittis-stripe:
### Context

We've noticed our JobSource job experiences "refCnt: 0" exceptions when an SSE connection is under high throughput (e.g., when saturated at 50MB/s, though we've seen the issue manifest at much lower throughput). This refCnt exception causes the SSE connection to terminate, and can even cause our job to enter a cascading failure where the load from the failed connection transfers to another one causing that one to fall over, etc.

With some debugging, we've tracked down the exception to being emitted by [this contentAsString call in SseWorkerConnection.java](https://github.com/Netflix/mantis/blob/master/mantis-server/mantis-server-worker-client/src/main/java/io/mantisrx/server/worker/client/SseWorkerConnection.java#L336). Our theory was that the call was racing with [the release in AutoReleaseByteBufOperator](https://github.com/ReactiveX/RxNetty/blob/0.4.x/rxnetty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java#L273), which we confirmed was true by catching the exception and confirm the connection terminations stopped.

```
try {
    return Observable.just(t1.contentAsString());
} catch (Throwable ex) {
    logger.info("Caught refCnt exception", ex);
    return Observable.just("");
}
```

At which point, we had a smoking gun that somehow the AutoReleaseByteBufOperator was racing with SseWorkerConnection when under high load.

This PR fixes the race condition by moving the contentAsString call into a synchronous call to map (moving the drop operator that forces the use of a flatMap to afterwards). I don't have strong understanding as to why this fixes the issue besides the hunch that flatMap might introduce some interesting asynchronous behavior.

Will defer to @crioux-stripe for feedback on why this works, and how to make it more idiomatic if required.

### Checklist

### DIFF
--- a/mantis-server/mantis-server-worker-client/src/main/java/io/mantisrx/server/worker/client/SseWorkerConnection.java
+++ b/mantis-server/mantis-server-worker-client/src/main/java/io/mantisrx/server/worker/client/SseWorkerConnection.java
@@ -320,9 +320,7 @@ public class SseWorkerConnection {
                     .subscribe();
         }
         return response.getContent()
-                .lift(new DropOperator<ServerSentEvent>(metricGroupId))
-                .rebatchRequests(this.bufferSize <= 0 ? 1 : this.bufferSize)
-                .flatMap((ServerSentEvent t1) -> {
+                .map((ServerSentEvent t1) -> {
                     lastDataReceived.set(System.currentTimeMillis());
                     if (isConnected.get() && isReceivingData.compareAndSet(false, true))
                         if (updateDataRecvngStatus != null)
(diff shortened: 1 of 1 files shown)


crioux-stripe (approved):
I had originally had some feedback about the last iteration with a `return ""`. I was hoping we could wrap the error and filter out the empty strings / error results. I was also looking to get a dropped metric incremented. However this change neatly eliminates those two issues.

As far as the failure mode goes I believe your theory is correct. The call to `map` essentially causes `.onNext` to be called in sequence all the way down the chain. The `flatMap` asynchronously batches the sub elements (even though there is only one) and thus introduces the potential for the race condition.

If I had to guess as to why this was a `flatMap` it was specifically to enable the ability to return `Observable.error()` and potentially `Observable.empty()` in cases where the contents couldn't be retrieved. This is a typical Rx pattern. Throwing (as you have done here) has the desired effect.

jpittis-stripe:
Following up here to confirm that this fix 100% resolved the stability issues we were experiencing. We did a little more digging to understand the repro, and it seems like Netty expects a connection's buffers to be pinned to a single scheduler, and that this flatMap call was causing those buffers to sometimes get used on a compute scheduler. In-fact, we were able to reproduce a pretty worrying bug at the RxJava layer where an observeOn call chained after the flatMap operator was able to affect the flatMap operator which was upstream to the observeOn call (which is unexpected given the docs claim observeOn only impacts downstream operators). Unclear to me if this is correct RxJava behavior, a bug, or a fault with RxJava 1. I'd be curious if someone on the Netflix side had context. If it did end up being a bug with RxJava 1, I presume that upgrading RxJava versions would be quite an undertaking. Regardless PR seems to fully resolve the issue on our side.

@Andyz26: In-case Netflix has any context on this shape of issue with RxJava observeOn behavior. I'm also curious why we ran into the bug but you guys seem to not have.

PR #760, opened 2025-03-17
Add jobAutoScaler level metrics

Andyz26:
### Context

Add a few metrics at job auto scaler level.
* Track auto scaler version using scaler id (job id + rule id)
* Metrics in scaler and metrics' request.

### Checklist

### DIFF
--- a/mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoScaler.java
+++ b/mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoScaler.java
@@ -16,8 +16,13 @@
 
 package io.mantisrx.server.worker.jobmaster;
 
+import com.netflix.spectator.api.Tag;
 import io.mantisrx.common.MantisProperties;
 import io.mantisrx.common.SystemParameters;
+import io.mantisrx.common.metrics.Counter;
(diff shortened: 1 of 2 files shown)


mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoScaler.java
```diff
@@ -378,9 +408,18 @@ public class StageScaler {
           return Observable.timer(delay, TimeUnit.SECONDS);
         });
 
-      public StageScaler(int stage, StageScalingPolicy scalingPolicy) {
+      public StageScaler(int stage, StageScalingPolicy scalingPolicy, String scalerId, Context context) {
```
dtrager02, line 411: nit: maybe instead of passing context, just pass a MetricsRegistry here from outside to avoid duplicate logic of getting the registry?

dtrager02 (approved):

mantis-runtime-executor/src/main/java/io/mantisrx/server/worker/jobmaster/JobAutoScaler.java
Andyz26: i was thinking that we might need more things from the context in upcoming changes.

PR #763, opened 2025-04-02
Fix Worker Heartbeat Timeout Logic for Launched Workers

Andyz26:
# Fix Worker Heartbeat Timeout Logic for Launched Workers

## Summary

This PR fixes a critical bug in the worker heartbeat timeout detection logic that was causing workers stuck in the `Launched` state to not be properly resubmitted. The issue was due to a time unit mismatch in the `JobActor` class where `Instant.ofEpochSecond()` was incorrectly used instead of `Instant.ofEpochMilli()` when calculating the time since a worker was launched.

### Bug Fix

The core issue was in `JobActor.java` where the system was incorrectly converting the worker's launch timestamp:

```java
// Before:
boolean noTimelyHeartbeatSinceLaunched = !workerMeta.getLastHeartbeatAt().isPresent()
    && Duration.between(Instant.ofEpochSecond(workerMeta.getLaunchedAt()), currentTime).getSeconds() > missedHeartBeatToleranceSecs;

// After:
boolean noTimelyHeartbeatSinceLaunched = !workerMeta.getLastHeartbeatAt().isPresent()
    && Duration.between(Instant.ofEpochMilli(workerMeta.getLaunchedAt()), currentTime)
        .getSeconds() > missedHeartBeatToleranceSecs;
```

This change ensures that the system correctly calculates the time elapsed since a worker was launched. The previous implementation incorrectly interpreted millisecond timestamps as seconds, resulting in vastly overestimated durations that prevented proper timeout detection.

### New Tests

Added a new test suite (`JobHeartbeatTimeoutTest.java`) with three test cases:

1. `testWorkerStuckInLaunchedStateWithoutHeartbeat`: Verifies that workers stuck in the `Launched` state without any heartbeats are properly resubmitted after the timeout period.

2. `testWorkerWithStaleHeartbeat`: Ensures that workers that initially sent heartbeats but then stopped sending them are properly resubmitted after the timeout period.

3. `testTimeUnitMismatchInHeartbeatCheck`: Specifically targets the time unit mismatch bug by verifying that the system correctly calculates the time since a worker was launched using milliseconds instead of seconds.

### Updated Existing Tests

Modified `JobTestLifecycle.java` to reflect the fixed behavior:
- Updated assertions to expect worker resubmission for launched workers with heartbeat timeouts
- Added state verification to ensure workers transition through the expected states
- Updated verification counts for scheduler interactions

### Other Changes

- In `ResourceClusterActorTest.java`:
  - Increased a timeout duration from 1 second to 5 seconds to improve test stability
  - Replaced `when()` with `doReturn()` in a mock setup to remove the flaky behavior introduced by the stubbing steup.

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/JobActor.java
@@ -2035,7 +2035,8 @@ public class JobActor extends AbstractActorWithTimers implements IMantisJobManag
                         // no heartbeat in a timely manner since launched or heartbeat too old
                         // note: the worker has been launched
                         boolean noTimelyHeartbeatSinceLaunched = !workerMeta.getLastHeartbeatAt().isPresent()
-                            && Duration.between(Instant.ofEpochSecond(workerMeta.getLaunchedAt()), currentTime).getSeconds() > missedHeartBeatToleranceSecs;
+                            && Duration.between(Instant.ofEpochMilli(workerMeta.getLaunchedAt()), currentTime)
+                                .getSeconds() > missedHeartBeatToleranceSecs;
                         boolean heartbeatTooOld = workerMeta.getLastHeartbeatAt().isPresent()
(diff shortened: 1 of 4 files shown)


dtrager02 (approved):
Great catch!

PR #766, opened 2025-04-16
[WIP] DynamoDB client session monitor

andresgalindo-stripe:
### Context

Setting session monitor so we can give up leadership if master loses lock for reasons other than a restart

### Checklist


mantis-control-plane/mantis-control-plane-dynamodb/src/main/java/io/mantisrx/extensions/dynamodb/DynamoDBClientSingleton.java
```diff
@@ -49,6 +49,7 @@ public class DynamoDBClientSingleton {
     private static DynamoDbClient instanceClient;
     private static String partitionKey;
     private static DynamoDBConfig conf;
+    private static long safeTimeWithoutHeartbeat;
```
kmg-stripe, line 52: nit: `safeTimeWithoutHeartbeatMs`?

kmg-stripe (approved):
just a nit

mantis-control-plane/mantis-control-plane-dynamodb/src/main/java/io/mantisrx/extensions/dynamodb/DynamoDBClientSingleton.java
andresgalindo-stripe: Ah, yeah, good call out - let me fix that and the related config up to make things more clear.

PR #770, opened 2025-04-22
Added log to indicate when we became leader

andresgalindo-stripe:
### Context

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-dynamodb/src/main/java/io/mantisrx/extensions/dynamodb/DynamoDBLeaderElector.java
+++ b/mantis-control-plane/mantis-control-plane-dynamodb/src/main/java/io/mantisrx/extensions/dynamodb/DynamoDBLeaderElector.java
@@ -139,6 +139,7 @@ public class DynamoDBLeaderElector extends BaseService {
                                     .withSessionMonitor(safeTimeWithoutHeartbeatMs, Optional.of(this::giveUpLeadership))
                                     .build());
             if (optionalLock.isPresent()) {
+                log.info("became leader");
                 leaderLock = optionalLock.get();
                 shouldLeaderElectorBeRunning.set(false);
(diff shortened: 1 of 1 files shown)


kmg-stripe (approved):

mantis-control-plane/mantis-control-plane-dynamodb/src/main/java/io/mantisrx/extensions/dynamodb/DynamoDBLeaderElector.java
```diff
@@ -139,6 +139,7 @@ protected boolean tryToBecomeLeader() {
                                     .withSessionMonitor(safeTimeWithoutHeartbeatMs, Optional.of(this::giveUpLeadership))
                                     .build());
             if (optionalLock.isPresent()) {
+                log.info("became leader");
```
hmitnflx, line 142: nit: consider adding leader identity like ip-address, etc.
I'm guessing those are already logged as part of the metadata - MDC/NDC tags

hmitnflx (approved):

andresgalindo-stripe:
@hmitnflx wanting to double check that merging this + tagging it will result in a publish to maven? It seems that wasn't the case with this tag I made recently: https://github.com/Netflix/mantis/releases/tag/v3.3.1-rc.1

hmitnflx:
> @hmitnflx wanting to double check that merging this + tagging it will result in a publish to maven? It seems that wasn't the case with this tag I made recently: https://github.com/Netflix/mantis/releases/tag/v3.3.1-rc.1

Yes, that's correct. I see the action run here - https://github.com/Netflix/mantis/actions/runs/14600909803
and the artifacts are available in the netflixoss artifactory here - https://netflixoss.jfrog.io/ui/native/maven-oss-candidates/io/mantisrx/mantis-control-plane-server/3.3.1-rc.1/

PR #784, opened 2025-08-05
Handle out of sync restarted TE

Andyz26:
### Context
From https://github.com/Netflix/mantis/discussions/781
This fix aim to handle the split state between resource cluster actor and job actor via an explicit message based on the previous state's workerId to help locate the target job actor when the reconnection happened. A new message type for such reconnection will be sent to the job actor and trigger the scheduling info update.

### DIFF
--- a/context/context-te-state-sync.md
+++ b/context/context-te-state-sync.md
@@ -1,0 +1,123 @@
+# Problem
+We recently hit a weird case that seems like a potential bug, or maybe a limitation.
+
+We have a 2 stage job, where for a short time all stage 2 workers were pulling events from what they thought was a stage 1 worker, but ended up being a worker for an unrelated job.
+
+As far as I can tell from the logs and code, it appears that:
+
+A stage 1 worker, say A-worker-1 on task executor TE-1 fails hard (looks like a JVM segfault)
+A minute or so later, TE-1 re-registers itself and is marked available
+A minute later B-worker-2 is mapped to TE-1
+At this point Job A still has A-worker-1 mapped to TE-1 in its scheduling info, so workers in stage 2 of A are pulling events from B-worker-2
+This persists for 2 more minutes, when A-worker-1 hits a heartbeat timeout and is failed by the JobActor.
+The main issue: Even though TE-1 is bound to A-worker-1 according to the JobActor, when it fails hard and comes back up, it is re-registered as available. I would expect there to be state knowing that TE-1 should not be marked as available.
+
+
+# Extra info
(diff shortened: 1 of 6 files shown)


context/context-te-state-sync.md
```diff
@@ -0,0 +1,123 @@
...
+
+#### 2. TaskExecutorState Tracking
+```java
+// Added to TaskExecutorState.java
+@Nullable
+private WorkerId previousWorkerId;
+
+// Store previous WorkerId on disconnection
+boolean onDisconnection() {
+    previousWorkerId = getWorkerId();
```
kmg-stripe, line 60: In the case we saw, I do not see an explicit disconnection event in the logs.  I wonder if the fact that the JVM seg faulted prevented this event, or maybe we were unlucky enough to have this specific event dropped?  If this is true, would we do the surgical approach if we have the worker id, else broadcast?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ResourceClusterActor.java
```diff
@@ -644,9 +645,22 @@ private void onTaskExecutorRegistration(TaskExecutorRegistration registration) {
...
             mantisJobStore.storeNewTaskExecutor(registration);
             if (stateChange) {
+                // If this TE was previously running a worker, notify the affected job
+                if (previousWorkerId != null) {
+                    log.info("Task executor {} reconnected, was previously running worker {}. Notifying job for scheduling refresh.",
+                             taskExecutorID, previousWorkerId);
+                    jobMessageRouter.routeWorkerEvent(new TaskExecutorReconnectedEvent(previousWorkerId, taskExecutorID));
+                    // Clear the previousWorkerId since we've sent the notification
+                    state.clearPreviousWorkerId();
+                }
```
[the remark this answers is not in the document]
Andyz26: i found a few caveats on the current approach. will send a new iteration later.

context/context-te-state-sync.md
kmg-stripe: nice!  it looks like this cleanly covers both cases of clean disconnection and fatal failures.

kmg-stripe (approved):
lgtm!  thanks for taking care of this!

we are currently in the process of upgrading to java 17, but it'll be at least another week before we have everything upgraded on our end.  would it be easiest for me to back port this fix to 3.x once it merges?

PR #789, opened 2025-09-10
Move common code to utils and cleanup

hellolittlej:
### Context

1. add validation for numOfPorts, as it's not supported right now 
2. move shared job request validation logic to utils 
3. remove wrapped request in taskexecutor, as there's no subscriber registered on this wrapped object in the current code base

### Checklist

@Andyz26


mantis-common/src/main/java/io/mantisrx/runtime/MantisJobDefinition.java
```diff
@@ -144,6 +144,8 @@ private void validateSchedulingInfo(boolean schedulingInfoOptional) throws Inval
                 throw new InvalidJobException("disk must be >=0, not " + machineDefinition.getDiskMB());
             if (machineDefinition.getNumPorts() < 0)
                 throw new InvalidJobException("numPorts must be >=0, not " + machineDefinition.getNumPorts());
+            if (machineDefinition.getNumPorts() > 1)
+                throw new InvalidJobException("custom number of ports has not supported yet.");
```
Andyz26, line 148: "custom port configuration is not currently supported."

Andyz26 (approved):

PR #793, opened 2025-09-24
Add log to check #TE archived was not in disabled state

hellolittlej:
### Context

- TE will be set to disabled state when scale down, add logs to check how many TE stayed in the archived state are not in disabled state.
- the archived state contains all the ghost TEs and other explicitly removed TE, adding this log to check if we could filter the ghost ones from this archived state

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
@@ -119,8 +119,11 @@ public class ExecutorStateManagerImpl implements ExecutorStateManager {
     private final Cache<TaskExecutorID, TaskExecutorState> archivedState = CacheBuilder.newBuilder()
         .maximumSize(10000)
         .expireAfterWrite(24, TimeUnit.HOURS)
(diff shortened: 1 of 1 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ExecutorStateManagerImpl.java
```diff
@@ -119,8 +119,10 @@ private TaskExecutorGroupKey findBestFitGroupOrDefault(SchedulingConstraints con
     private final Cache<TaskExecutorID, TaskExecutorState> archivedState = CacheBuilder.newBuilder()
         .maximumSize(10000)
         .expireAfterWrite(24, TimeUnit.HOURS)
-        .removalListener(notification ->
-            log.info("Archived TaskExecutor: {} removed due to: {}", notification.getKey(), notification.getCause()))
+        .removalListener(notification -> {
+            TaskExecutorState state = (TaskExecutorState) notification.getValue();
+            log.info("Archived TaskExecutor: {} with disabled state: {} removed due to: {}", notification.getKey(), state.onNodeDisabled(), notification.getCause());
```
Andyz26, line 124: null check on state?

Andyz26 (approved):

PR #843, opened 2026-04-10
Healthcheck Endpoint

james-lubin:
### Context

Adds a health check endpoint that verifies that all workers for every active job in a given job cluster have the status of `started`. Originally, the plan was to also include an interface that would allow plugging in custom health checks ([experimented with this here](https://github.com/Netflix/mantis/pull/842)). However, I've pivoted away from this because it added a lot of complexity, and constrained callers. For customizing health checking, user should proxy this Akka route, and then add additional health checks afterwards.

### Checklist


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v1/JobClustersRoute.java
```diff
@@ -785,4 +797,23 @@ private Route deleteScalerRuleRoute(String clusterName, String ruleId) {
             HttpRequestMetrics.HttpVerb.DELETE
         );
     }
+
+    private Route healthCheckRoute(String clusterName) {
+        logger.trace("GET /api/v1/jobClusters/{}/healthcheck called", clusterName);
+
+        return parameterMap(params -> {
+            String jobIdsParam = params.get("job-ids");
```
hellolittlej, line 805: what about we have users to provide the region, and we will determine the health for all jobs in that region?

here it seems we are expecting users to provide list of job ids, do we have use case where we only need inspect subset of jobs within a given region?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -2633,6 +2641,58 @@ public void onJobScalerRuleStream(JobClusterScalerRuleProto.GetJobScalerRuleStre
...
+        ListJobCriteria criteria = new ListJobCriteria();
+        getFilteredNonTerminalJobList(criteria, prefilteredJobIds)
+                .collect(Lists::<FailedWorker>newArrayList, (failedWorkers, view) -> {
+                    if (view.getWorkerMetadataList() != null) {
+                        for (FilterableMantisWorkerMetadataWritable worker : view.getWorkerMetadataList()) {
+                            if (worker.getState() != MantisJobState.Started) {
+                                failedWorkers.add(new FailedWorker(
+                                        worker.getWorkerIndex(),
+                                        worker.getWorkerNumber(),
+                                        worker.getState().name()));
```
Andyz26: this is categorizing pending workers as failed?

Andyz26: imo the health logic is something like a composite of (all workers for each index have started, failed/resubmit workers count in past x duration is below a threshold; alert pack etc).

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/api/akka/route/v1/JobClustersRoute.java
james-lubin: By default, it checks any active jobs. The job-ids param is optional, it just does the default check if it's null. But if you do specify it, it only checks the job-ids that were inserted. 

This probably won't be used right away, it's more of a forward looking change. In the future, where we have a fully managed Mantis CD system, this will allow us to check just jobs that were recently deployed. It's also an additional knob that gives some of the power users more flexibility.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
james-lubin: Yeah, maybe I can improve the naming here, but the behavior is intentional, the idea is that after a deploy you can call this endpoint with n amount of retries until the workers have started. So if there are some stuck in accepted/pending the healthcheck will fail.

Alternatively, since we're using SSE, we could just keep the connection open and send an event when the workers start or a timeout is reached. But I think the way its implemented is easier to build on top of.

Edit: Didn't see your second comment until refreshing. Agreed on the composite, I'm just planning to add the rest of that within nfmantis.

Andyz26: maybe change the name to something like "notReady"?

james-lubin: good call, went with unready

james-lubin:
After making the previous naming change, the build was repeatedly failing due to flaky tests. The tests were all in `ResourceClusterActorTest.java`, but the failure wasn't caused by the same test on every build. After doing some Claude assisted sleuthing, it seems that the issue is due to multiple threads accessing Mockito mock objects at the same time, which is a problem because the mocks aren't thread safe. This snippet from one of the failed builds supports this theory.

```
2026-04-14 00:33:28,414 ERROR [default-akka.actor.internal-dispatcher-2] MantisActorSupervisorStrategy:48 - Stopping the actor because of exception 
2026-04-14T00:33:31.5836098Z     akka.actor.ActorInitializationException: akka://default/user/$n: exception during creation
.
.
.
2026-04-14T00:33:31.5838934Z     Caused by: org.mockito.exceptions.misusing.WrongTypeOfReturnValue: 
2026-04-14T00:33:31.5839150Z     TaskExecutorRegistration cannot be returned by getJobArtifactsToCache()
2026-04-14T00:33:31.5839410Z     getJobArtifactsToCache() should return List
2026-04-14T00:33:31.5839482Z     ***
2026-04-14T00:33:31.5839628Z     If you're unsure why you're getting above error read on.
2026-04-14T00:33:31.5839797Z     Due to the nature of the syntax above problem might occur because:
2026-04-14T00:33:31.5839990Z     1. This exception *might* occur in wrongly written multi-threaded tests.
2026-04-14T00:33:31.5840164Z        Please refer to Mockito FAQ on limitations of concurrency testing.
2026-04-14T00:33:31.5840384Z     2. A spy is stubbed using when(spy.foo()).then() syntax. It is safer to stub spies - 
2026-04-14T00:33:31.5840717Z        - with doReturn|Throw() family of methods. More in javadocs for Mockito.spy() method.
```

Sometimes I saw this one instead, it would trigger on a mock that was stubbed correctly.
```
org.mockito.exceptions.misusing.UnfinishedStubbingException:                                                                                         
  Unfinished stubbing detected here:                                                                                                                   
  -> at io.mantisrx.master.resourcecluster.ResourceClusterActorTest.testTaskExecutorIsDisabledEvenAfterRestart(ResourceClusterActorTest.java:960)      
                                                                                                                                                       
  E.g. thenReturn() may be missing.                                                                                                                    
  Examples of correct stubbing:                                                                                                                        
      when(mock.isOk()).thenReturn(true);                                                                                                              
      when(mock.isOk()).thenThrow(exception);                                                                                                          
      doThrow(exception).when(mock).someVoidMethod();                                                                                                  
  Hints:                                                                                                                                               
   1. missing thenReturn()                                                                                                                             
   2. you are trying to stub a final method, which is not supported                                                                                    
   3. you are stubbing the behaviour of another mock inside before 'thenReturn' instruction is completed                                               
                                                                                                                                                       
      at                                                                                                                                               
  app//io.mantisrx.master.resourcecluster.ResourceClusterActorTest.testTaskExecutorIsDisabledEvenAfterRestart(ResourceClusterActorTest.java:960)
```

Here's Claude's explanation of the issue.
```
  1. @Before setupTest() calls setupActor(), which does actorSystem.actorOf(props) — this returns immediately without waiting for preStart() to finish.
  2. The actor's preStart() runs on an Akka dispatcher thread and calls mantisJobStore.getJobArtifactsToCache(clusterID) (line 622 of
  ResourceClusterActor.java).
  3. The test body then hits line 960: doReturn(...).when(mantisJobStore).getJobArtifactsToCache(CLUSTER_ID) — this calls the mock method to record the
   stub.
  4. Both the actor thread and the test thread are simultaneously calling methods on the same Mockito mock object. Mockito's InvocationContainerImpl is
   not thread-safe, so this corrupts its internal state and triggers UnfinishedStubbingException.
```

The fix was to add this line
```
resourceCluster.getRegisteredTaskExecutors().get();
```
to the end of the `setupActor()` function. Internally, this uses an `ask()` call which blocks the test thread until it gets a reply from the actor. The actor can only reply once `preStart()` has finished executing. Which means the race condition between the test thread and actor thread accessing that mock object is eliminated.

This is good enough for a work-around, but we should update this test class in the future to make it properly thread safe.

Andyz26 (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -2633,6 +2641,58 @@ public void onJobScalerRuleStream(JobClusterScalerRuleProto.GetJobScalerRuleStre
...
+                    .filter(Optional::isPresent)
+                    .map(Optional::get)
+                    .forEach(prefilteredJobIds::add);
+        }
+
+        ListJobCriteria criteria = new ListJobCriteria();
+        getFilteredNonTerminalJobList(criteria, prefilteredJobIds)
+                .collect(Lists::<UnreadyWorker>newArrayList, (unreadyWorkers, view) -> {
+                    if (view.getWorkerMetadataList() != null) {
+                        for (FilterableMantisWorkerMetadataWritable worker : view.getWorkerMetadataList()) {
```
Andyz26: another case here: worker index 1 started then failed, another agent got index 1 worker and started correctly. can this case be handled here?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
```diff
@@ -2633,6 +2641,58 @@ public void onJobScalerRuleStream(JobClusterScalerRuleProto.GetJobScalerRuleStre
...
+                            if (worker.getState() != MantisJobState.Started) {
```
hellolittlej: can this be per job actor level message, so we can get each job details to determine the status for each job?

Andyz26: +1

james-lubin: yeah, I'll update this

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/JobClusterActor.java
james-lubin: In that case, I'd expect the first call to report the job cluster as unhealthy, and the second call to succeed.

PR #844, opened 2026-04-22
Fix noisy retry when noResourceError with rev based scheduling

hellolittlej:
### Context

- The inflightReservationRequest retried every 5s for no resource Error, previous it used to be 1 minute with master scheduling schedule.
- add the delay for scaling operation

### Checklist


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/resourcecluster/ReservationRegistryActor.java
```diff
@@ -465,17 +468,20 @@ private void onStatusFailure(Status.Failure failure) {
...
-                    // This tracks when the reservation last received a NoResourceAvailableException
-                    // The reservation will be retried if this timestamp is older than the timeout
-                    inFlightReservationRequestTimestamps.put(exceptionConstraintKey, clock.instant());
+                    // Push the in-flight timestamp into the future so shouldSkipConstraintGroup's
+                    // elapsed-time check (now - timestamp < inFlightReservationTimeout) keeps this
+                    // group skipped for an additional DEFAULT_NO_RESOURCE_EXTRA_DELAY before the
+                    // next retry after 60s.
+                    inFlightReservationRequestTimestamps.put(
+                        exceptionConstraintKey,
+                        clock.instant().plus(DEFAULT_NO_RESOURCE_EXTRA_DELAY));
```
james-lubin, line 484: is `inFlightReservationRequestTimestamps` only used so that `shouldSkipConstraintGroup` can make its determination? If we add metrics/logs based on the values in `inFlightReservationRequestTimestamps`, this could get confusing. I wonder if we should add this handling within `shouldSkipConstraintGroup` instead?

PR #853, opened 2026-05-05
Make scale-up worker additions transactional

Andyz26:
# Make scale-up worker additions transactional

  ## Summary

  `JobActor.scaleStage` previously allowed in-memory job metadata to drift from
  durable storage on `MantisJobStore.storeNewWorker` failures: the new `JobWorker`
   was added to the stage's index/number maps but no record reached the store,
  leaving sparse worker indexes that broke subsequent scale-down, heartbeat
  resubmits, and index-based lookups.

  This PR makes each scale-up addition transactional via a two-phase
  prepare/commit/rollback flow. Failed adds are reverted in-memory, the stage
  target is shrunk by one, and the loop continues with the failed slot reused —
  keeping indexes dense. Partial scales are visible to the caller through a richer
   response and a new metric.

  ## Changes

  **Core (`JobActor`, `MantisJobMetadataImpl`, `MantisStageMetadataImpl`)**
  - New `PendingWorkerAddition` wraps the in-memory mutation done by
  `prepareWorker`. Caller commits after a successful `storeNewWorker` or rolls
  back on failure.
  - `IWorkerManager.scaleStage` now returns `ScaleStageResult` (actual / requested
   / failedAdditions) instead of a bare int. Response messages distinguish full
  success from partial: `"Partially scaled stage 1 to 3 workers (requested 4; 1
  add failures)"`.
  - New rollback primitives `MantisJobMetadataImpl.unsafeRemoveWorkerMetadata` and
   `MantisStageMetadataImpl.removeWorkerIndex` use a check-first-then-CAS pattern
  (`remove(K, V)`) and fully revert if the second remove fails.
  - `MantisStageMetadataImpl.unsafeSetNumWorkers` reverts `this.numWorkers` if
  `updateStage` throws, keeping in-memory and durable counts coherent.
  - `JobActor.onJobInitialize` stops the actor on submit-time init failure
  (`isSubmit=true` only, to preserve init-retry semantics for store reloads).

  **Failure-mode escalation**
  - *Shrink-failure after partial success:* throw
  `PartialScaleStageFailureException` carrying the realized partial count; queued
  workers are flushed first; caller gets `SERVER_ERROR` with `actualNumWorkers` =
  realized count. Original add-failure preserved as cause; shrink failure attached
   as suppressed.
  - *Rollback corruption (stage maps inconsistent):* fatal — sends
  `JobClusterProto.KillJobRequest` to the parent so the job is failed and
  archived. Workers stored before the corruption are cleaned up via the job
  archive lifecycle.

  **Metrics**
  - New `numWorkerStoreFailures` counter increments per rollback-and-shrink.
  - `numScaleStage` now also increments on partial-success scales (still skipped
  on plain failures that durably stored nothing).

  **Misc**
  - `mantis.worker.resubmission.interval.secs` default lengthened from `5:10:20` →
   `5:10:30:60:120:600`. (Tangential — flag for split if reviewers prefer.)

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/MantisStageMetadataImpl.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/MantisStageMetadataImpl.java
@@ -332,9 +332,14 @@ public class MantisStageMetadataImpl implements IMantisStageMetadata {
      * @throws Exception
      */
     public void unsafeSetNumWorkers(int numWorkers, MantisJobStore store) throws Exception {
+        int previousNumWorkers = this.numWorkers;
         this.numWorkers = numWorkers;
-
-        store.updateStage(this);
+        try {
+            store.updateStage(this);
+        } catch (Exception e) {
+            this.numWorkers = previousNumWorkers;
+            throw e;
+        }
     }
 
     /**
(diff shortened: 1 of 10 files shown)


mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/MantisStageMetadataImpl.java
```diff
@@ -332,9 +332,14 @@ public IMantisStageMetadata build() {
...
     public void unsafeSetNumWorkers(int numWorkers, MantisJobStore store) throws Exception {
+        int previousNumWorkers = this.numWorkers;
         this.numWorkers = numWorkers;
-
-        store.updateStage(this);
+        try {
+            store.updateStage(this);
+        } catch (Exception e) {
+            this.numWorkers = previousNumWorkers;
+            throw e;
```
hellolittlej, line 341: do we log the exception and track the error here?

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/MantisStageMetadataImpl.java
```diff
@@ -570,6 +575,70 @@ public boolean addWorkerIndex(JobWorker newWorker) {
...
+                            + "(workerByIndex=%s, workerByNumber=%s)",
+                    jobId.getId(),
+                    stageNum,
+                    workerIndex,
+                    workerNumber,
+                    describeRollbackWorker(workerByIndex),
+                    describeRollbackWorker(workerByNumber)));
+        }
+
+        if (!workerByIndexMetadataSet.remove(workerIndex, workerByIndex)) {
```
hellolittlej, line 604: in which case we will fail this check and below one would fail when the first check passed?

james-lubin (approved):

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/jobcluster/job/MantisStageMetadataImpl.java
Andyz26: i think the log is at upper layer/caller.

PR #854, opened 2026-05-07
Make oss mantis java8 compatible

hellolittlej:
### Context

Using Java 17 toolchain to compile, but with --release 8 to produce Java 8 compatible bytecode.
mantis publish library depends on the publish-core, and we do not wanna publish library to be stick with java17 for compatibility concerns

### Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -132,9 +132,6 @@ subprojects {
 
     group = 'io.mantisrx'
 
-    sourceCompatibility = JavaVersion.VERSION_17
-    targetCompatibility = JavaVersion.VERSION_17
-
     if (project.hasProperty('useMavenLocal')) {
         repositories {
             mavenLocal()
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -132,8 +132,9 @@ subprojects {
 
     group = 'io.mantisrx'
 
-    sourceCompatibility = JavaVersion.VERSION_17
-    targetCompatibility = JavaVersion.VERSION_17
+    tasks.withType(JavaCompile).configureEach {
```
Andyz26: maybe limit this to publish module only?

hellolittlej: yea good point. I was updating it as well coz control plane core using seal/record not compatible with 8

Andyz26 (approved):

james-lubin (approved):

PR #858, opened 2026-05-22
Fix incorrect number of source job types in docs

saurabhsingh72487-hub:
The documentation stated there are four types of Source Jobs, but only two are currently documented on the page.

Updated the text to reflect the documented behavior.


hellolittlej (approved):

hellolittlej:
Close this PR as we not accepting this change for now

PR #866, opened 2026-07-30
Rate limit the job submission request

hellolittlej:
### What

Adds an opt-in, per route based bucket rate limit.For jobsRoute, add the rate limit in front of the three v1 job-creation endpoints:

- `POST /api/v1/jobs`
- `POST /api/v1/jobClusters/{cluster}/jobs`
- `POST /api/v1/jobs/actions/quickSubmit`

The rate

### Configuration

| Property | Default | Notes |
| --- | --- | --- |
| `mantis.master.api.v1.submitJob.throttle.enabled` | `false` | Off by default; no limiter is constructed when off |
| `mantis.master.api.v1.submitJob.permitsPerSecond` | `1000` | Re-readable at runtime, no deploy needed |

- Guava RateLimiter, non-blocking tryAcquire(); token-bucket with the usual ~1s of burst accumulation, so a 1000/s limit admits up to a ~1000-request burst after an idle period.
- Shed → HTTP 429, checked in BaseRoute.withThrottle:141 before the entity is unmarshalled — a storm costs a permit check, not a parse plus an actor ask.

Off by default because the right ceiling depends on cluster size and traffic —
operators opt in once they've measured their own submit rate. The rate is backed by a
`LongDynamicProperty`, so it re-reads on the dynamic-property refresh interval
(`mantis.config.dynamic.refreshSecs`, default 30s).

### Metrics

`MasterApiMetrics/throttledRequestCount` gains a `source` tag (`submitJob`,
`resourceCluster`), so an operator can tell which limit is shedding. Previously the
route-level throttle and the resource-cluster gateway's `RequestThrottledException`
both incremented the same untagged counter.

### Checklist

### DIFF
--- a/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/utils/ApiRequestRateLimiter.java
+++ b/mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/utils/ApiRequestRateLimiter.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2024 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.mantisrx.master.utils;
+
+import akka.http.javadsl.model.HttpRequest;
+
+/**
+ * Admission control for a set of API endpoints: decides whether an inbound request may proceed or
+ * should be shed (typically as a 429).
+ *
+ * <p>Implementations are called on the request thread before the entity is unmarshalled, so they
+ * must be cheap and must never block.
+ *
+ * <p>An instance owns one bucket, and which endpoints draw from it is decided entirely by where the
+ * instance is wired in — the granularity is a wiring decision, not a property of this interface.
+ * Endpoints wired to one instance share its permits, so group them by the resource they contend for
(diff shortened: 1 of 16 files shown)


mantis-control-plane/mantis-control-plane-server/src/test/java/io/mantisrx/master/api/akka/route/v0/JobRouteTest.java
```diff
@@ -251,7 +252,8 @@ public static void setup() throws Exception {
                 final JobsRoute v1JobsRoute = new JobsRoute(
                         jobClusterRouteHandler,
                         jobRouteHandler,
-                        system);
+                        system,
+                        ApiRequestRateLimiter.UNLIMITED);
```
Andyz26, line 256: can this be default so root signature remain the same?

hellolittlej: This is v0 route test but it's using the v1 job route with the legacy reason. ...

I prefer we not adding the overload just to accommodate a legacy test setup.

mantis-control-plane/mantis-control-plane-server/src/main/java/io/mantisrx/master/utils/ApiRequestRateLimiter.java
```diff
@@ -50,4 +51,20 @@ public interface ApiRequestRateLimiter {
...
+     *
+     * <p>The default of one second is the smallest value {@code Retry-After} can express in seconds, and
+     * suits any bucket refilling at a permit per second or faster. An implementation whose bucket refills
+     * more slowly than that should override this — telling a caller to come back in a second when the next
+     * permit is a minute away just converts one shed request into sixty.
+     *
+     * @return a positive duration; whole seconds, since {@code Retry-After} has no finer resolution
+     */
+    default Duration getRetryAfter() {
+        return Duration.ofSeconds(1);
```
Andyz26, line 79: always retry after 1s?

Andyz26: maybe add a todo item here to extend this logic

Andyz26 (approved):

